PR #463, opened 2014-12-30
OGM-640 : Add support for Character type

ajaybhat:
Please review.


core/src/main/java/org/hibernate/ogm/type/impl/TypeTranslatorImpl.java
```diff
@@ -35,11 +35,14 @@
...
 import org.hibernate.type.descriptor.java.LongTypeDescriptor;
 import org.hibernate.type.descriptor.java.PrimitiveByteArrayTypeDescriptor;
+import org.hibernate.type.descriptor.java.ShortTypeDescriptor;
 import org.hibernate.type.descriptor.java.StringTypeDescriptor;
 import org.hibernate.type.descriptor.java.UUIDTypeDescriptor;
 import org.hibernate.type.descriptor.java.UrlTypeDescriptor;
 import org.hibernate.usertype.UserType;
 
+import static org.hibernate.ogm.util.impl.CollectionHelper.newHashMap;
+
```
DavideD: We try to avoid this kind of changes in a pull request since they are unrelated to the issue.
We also tend to keep the staic import before the regular imports

core/src/main/java/org/hibernate/ogm/type/impl/CharacterType.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.type.impl;
+
+import org.hibernate.MappingException;
+import org.hibernate.engine.spi.Mapping;
+import org.hibernate.ogm.type.descriptor.impl.WrappedGridTypeDescriptor;
+import org.hibernate.type.descriptor.java.CharacterTypeDescriptor;
+
+public class CharacterType extends AbstractGenericBasicType<Character> {
```
DavideD: Could you add the javadoc with the author tag and a small description?

core/src/test/java/org/hibernate/ogm/backendtck/type/Bookmark.java
```diff
@@ -1,240 +1 @@
...
	}

	public Character getDelimiter() {
		return delimiter;
	}

	public void setDelimiter(Character delimiter) {
		this.delimiter = delimiter;
	}
}
```
DavideD: What happend to this class?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -652,6 +654,12 @@ public GridType overrideType(Type type) {
 		else if ( type == StandardBasicTypes.BYTE ) {
 			return ByteStringType.INSTANCE;
 		}
+		else if ( type == StandardBasicTypes.SHORT ) {
+			return ShortType.INSTANCE;
```
DavideD: you haven't created a specific type for MongoDB, so you don't t override the type in this case

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -652,6 +654,12 @@ public GridType overrideType(Type type) {
...
+		}
+		else if ( type == StandardBasicTypes.FLOAT ) {
+			return FloatType.INSTANCE;
```
DavideD: you haven't created a specific type for MongoDB, so there is no need to override the type in this case (same for short type)

DavideD:
The patch looks good, I would only change a few minor things, if you could address them and remove all the unrelated changes, it would be great.

By the way, are you using our template and code styles? I think we need to improve the "How to contribute page" on our web sitem to include them but let me know if you haven't found them.

ajaybhat:
I'll incorporate your suggestions. 
For author tags, do I have to add an email ID?

As for formatting, I use the code template described here and I haven't encountered any checkstyle errors after using this template. 
https://developer.jboss.org/wiki/ContributingToHibernateUsingIntelliJ

I use IntelliJ's autoformat a lot, which is why some of the classes have been reformatted entirely. I don't know if that's okay.

DavideD:
> For author tags, do I have to add an email ID?

No, it's better without the email

> I use IntelliJ's autoformat a lot, which is why some of the classes have been reformatted entirely. I don't know if that's okay

The templates you are using are the right ones.
The autoformat option on existing code should be used carefully, Sometimes it changes the code in a way that it is less readable. It might also add a lot of small changes that makes the patch harder to review.

In general, I try not to change the format of existing code and If I really desire to change it, I keep it in a seperate commit so that it's easier to understand the real fix. Note that this is  a rule of thumb and it might change depending who is reviewing the pull request :)

core/src/test/java/org/hibernate/ogm/backendtck/type/Bookmark.java
emmanuelbernard: Maybe @ajaybhat has mistakenly used spaces instead of tabs for indentation?

gunnarmorling:
> In general, I try not to change the format of existing code and If I really desire to change it, I keep it in a seperate commit so that it's easier to understand the real fix

+1 Please never mix actual code changes with formatting changes in one commit, as otherwise reviewing is a huge pain. Re auto-formatting, we should stay away from re-formatting entire files for no reason, as Davide is saying.

gunnarmorling:
@ajaybhat This looks very good now! It's ready for integration, could you only "sign" the [contributor license agreement](https://cla.jboss.org/index.seam) if you haven't done so yet?

We also should add the new types to the [reference documentation](https://docs.jboss.org/hibernate/ogm/4.1/reference/en-US/html_single/); Do you think you could add a commit for that as well? The docs are written in AsciiDoc and live [here](https://github.com/hibernate/hibernate-ogm/tree/master/documentation/manual). There is a list of supported types in [5.2](https://docs.jboss.org/hibernate/ogm/4.1/reference/en-US/html_single/#ogm-mapping-supported-types) and for each datastore the mapping of the types is described (e.g. just search for "BigInteger").

Thanks again!

ajaybhat:
> could you only "sign" the contributor license agreement if you haven't done so yet?

I've signed it now; I think everything is in order there.
I've also added to the manual, is there anything else that I need to do?

> Thanks again!

No problem :) Can you suggest another issue for me to work on?

gunnarmorling:
Rebased and applied.

> I've signed it now; I think everything is in order there.

Cool.

> I've also added to the manual, is there anything else that I need to do?

Nope that's it. I'll send one more PR for some more doc updates.

gunnarmorling:
> Can you suggest another issue for me to work on?

Yes, sure! Great that you have you support for the project :)

You may go to JIRA, and look for issues marked as "suitable for new contributors", amongst these you should find something. You can find these issues [here](https://hibernate.atlassian.net/issues/?jql=project%20%3D%20OGM%20AND%20status%20in%20%28Open%2C%20Reopened%29%20AND%20%22Suitable%20for%20new%20contributors%22%20%3D%20%22Yes%2C%20likely%22%20ORDER%20BY%20key%20DESC%2C%20priority%20DESC).

Of course it depends on what you'd like to work on (a specific dialect, or the core engine etc.). E.g. you could take a look at [OGM-616](https://hibernate.atlassian.net/browse/OGM-616), [OGM-265](https://hibernate.atlassian.net/browse/OGM-265) or [OGM-674](https://hibernate.atlassian.net/browse/OGM-674).

Just let me know in case you have any questions or would like to work on something else, there is much to do :)

ajaybhat:
> Just let me know in case you have any questions or would like to work on something else, 

I'd like to work on this issue, but I need some help on it, like has any work been done on it, and whether the BackendCustomLoader needs to follow exactly the design of CustomLoader
[Improvements around BackendCustomLoader] (https://hibernate.atlassian.net/browse/OGM-564)

gunnarmorling:
@ajaybhat, just replied on OGM-564. It doesn't have to exactly follow `CustomLoader`, but it should basically do everything the latter does :) Ideally we'd even be able to share parts of the implementation, not sure though how realistic it is. One could consider to extract a generic implementation, with an ORM (result set based) and an OGM (tuple based) implementation. I guess it requires some tinkering :)

PR #465, opened 2015-01-05
OGM-701 Use enum for authentication options

DavideD:
@gunnarmorling Is this what you meant?

### DIFF
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/configuration/impl/MongoDBConfiguration.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/configuration/impl/MongoDBConfiguration.java
@@ -15,6 +15,7 @@ import org.hibernate.ogm.datastore.mongodb.MongoDBProperties;
 import org.hibernate.ogm.datastore.mongodb.impl.MongoDBDatastoreProvider;
 import org.hibernate.ogm.datastore.mongodb.logging.impl.Log;
 import org.hibernate.ogm.datastore.mongodb.logging.impl.LoggerFactory;
+import org.hibernate.ogm.datastore.mongodb.options.AuthenticationMechanismType;
 import org.hibernate.ogm.datastore.mongodb.options.impl.ReadPreferenceOption;
 import org.hibernate.ogm.datastore.mongodb.options.impl.WriteConcernOption;
 import org.hibernate.ogm.options.spi.OptionsContext;
@@ -53,7 +54,7 @@ public class MongoDBConfiguration extends DocumentStoreConfiguration {
 	private final int timeout;
 	private final WriteConcern writeConcern;
 	private final ReadPreference readPreference;
-	private final String authenticationMechanism;
+	private final AuthenticationMechanismType authenticationMechanism;
 
 	/**
 	 * Creates a new {@link MongoDBConfiguration}.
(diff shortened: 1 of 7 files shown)


mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/AuthenticationMechanismType.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Available authentication methods for MongoDB.
+ *
+ * @see com.mongodb.MongoCredential
+ * @author Davide D'Alto
+ */
+public enum AuthenticationMechanismType {
+
+	GSSAPI("GSSAPI"),
+	MONGODB_CR("MONGODB-CR"),
```
gunnarmorling: So far we used the 1:1 string representation of the enum constants when specifying their value as properties. I understand it's not as nice here due to the underscore, but I'd stick to that pattern for the sake of consistency. WDYT?

DavideD: That's the value used by mongodb so it has to be `MONGO-CR` note that MONGODB_CR will work anyway.

Now that I think about it it should be `MongoCredential.MONGODB_CR_MECHANISM`

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/configuration/impl/MongoDBConfiguration.java
```diff
@@ -92,19 +110,18 @@ public MongoClientOptions buildOptions() {
...
-			}
-			if ( MongoCredential.GSSAPI_MECHANISM.equals( authenticationMechanism ) ) {
-				return Arrays.asList( MongoCredential.createGSSAPICredential( getUsername() ) );
-			}
-			if ( MongoCredential.MONGODB_X509_MECHANISM.equals( authenticationMechanism ) ) {
-				return Arrays.asList( MongoCredential.createMongoX509Credential( getUsername() ) );
-			}
-			if ( MongoCredential.PLAIN_MECHANISM.equals( authenticationMechanism ) ) {
-				return Arrays.asList( MongoCredential.createPlainCredential( getUsername(), getDatabaseName(), password() ) );
+			switch ( authenticationMechanism ) {
```
gunnarmorling: Not too important, but you might consider to establish a method on the enum itself instead:

    return authenticationMechanism.getCredentials( getUsername(), getDatabaseName(), password() );

It saves you the switch and you don't even need the error message below.

DavideD: ok

DavideD: I'll change it

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/AuthenticationMechanismType.java
gunnarmorling: > so it has to be MONGO-CR

Why that, we don't pass that value to MongoDB as is, right?

> note that MONGODB_CR will work anyway.

Ah, I just see you do some manual parsing above. That can be left to the configuration parser to deal with (provided the toString values of the enum members are used):

    AuthenticationMechanismType authenticationMechanism = propertyReader
        .property( MongoDBProperties.AUTHENTICATION_MECHANISM, AuthenticationMechanismType.class )
        .withDefault( AuthenticationMechanismType.MONGODB_CR )
        .getValue();

That way the value can also be given as an actual enum member when programmatically bootstrapping OGM. That was actually the main motivation for bringing up the enum topic, otherwise we could stick to the current implementation.

DavideD:
I've applied the remarks, 
I've squashed the remarks in the latest commit since it is a small fix

gunnarmorling:
Rebased and applied. Thx, @DavideD.

PR #468, opened 2015-01-08
OGM-708 Logging name of actual dialect class, also logging name of affec...

gunnarmorling:
...ted entity type


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/468
JIRA: [OGM-708](https://hibernate.atlassian.net/browse/OGM-708)

Thanks

PR #469, opened 2015-01-08
OGM-710 Remove unneeded dependencies and dependency exports from WildFly...

Sanne:
... modules

https://hibernate.atlassian.net/browse/OGM-710

I could not remove the export of
    `<module name="org.jboss.jts" export="true" />`

as that gets much trickier. I'll create a specific JIRA for it.

### DIFF
--- a/modules/wildfly/src/main/modules/ogm/couchdb/module.xml
+++ b/modules/wildfly/src/main/modules/ogm/couchdb/module.xml
@@ -10,11 +10,12 @@
         <resource-root path="hibernate-ogm-couchdb-${project.version}.jar" />
     </resources>
     <dependencies>
-        <module name="org.hibernate" export="true" services="import" slot="${hibernate.ogm.module.slot}" />
+        <module name="org.hibernate" slot="${hibernate.ogm.module.slot}" />
         <module name="org.hibernate.hql" slot="${hibernate.hql.module.slot}" />
 
         <module name="javax.ws.rs.api" />
         <module name="javax.persistence.api" />
+        <module name="javax.transaction.api" />
         <module name="org.jboss.resteasy.resteasy-jackson2-provider" />
         <module name="org.jboss.resteasy.resteasy-jaxrs" />
         <module name="org.jboss.logging" />
--- a/modules/wildfly/src/main/modules/ogm/infinispan/module.xml
+++ b/modules/wildfly/src/main/modules/ogm/infinispan/module.xml
@@ -10,11 +10,10 @@
         <resource-root path="hibernate-ogm-infinispan-${project.version}.jar" />
     </resources>
     <dependencies>
-        <module name="org.hibernate" export="true" services="import" slot="${hibernate.ogm.module.slot}" />
+        <module name="org.hibernate" slot="${hibernate.ogm.module.slot}" />
         <module name="org.hibernate.hql" slot="${hibernate.hql.module.slot}" />
 
-        <module name="org.infinispan" optional="true" slot="${infinispan.module.slot}" />
-        <module name="org.infinispan.commons" optional="true" slot="${infinispan.module.slot}" />
+        <module name="org.infinispan" slot="${infinispan.module.slot}" />
         <module name="javax.api" />
         <module name="javax.persistence.api" />
         <module name="javax.transaction.api" />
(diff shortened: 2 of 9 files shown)


modules/wildfly/src/main/modules/ogm/couchdb/module.xml
```diff
@@ -10,11 +10,12 @@
         <resource-root path="hibernate-ogm-couchdb-${project.version}.jar" />
     </resources>
     <dependencies>
-        <module name="org.hibernate" export="true" services="import" slot="${hibernate.ogm.module.slot}" />
+        <module name="org.hibernate" slot="${hibernate.ogm.module.slot}" />
         <module name="org.hibernate.hql" slot="${hibernate.hql.module.slot}" />
 
         <module name="javax.ws.rs.api" />
         <module name="javax.persistence.api" />
+        <module name="javax.transaction.api" />
```
gunnarmorling: What's the reason for not leaving `<module name="javax.transaction.api" export="true" />` in core an instead removing it from all the other modules? Core exposes a JTA type on one of its APIs used by these other modules, so IMO it's in charge to make it available rather than putting this task to the modules using it. Both works, just curious about your reasoning.

modules/wildfly/src/main/modules/ogm/infinispan/module.xml
```diff
@@ -10,11 +10,10 @@
         <resource-root path="hibernate-ogm-infinispan-${project.version}.jar" />
     </resources>
     <dependencies>
-        <module name="org.hibernate" export="true" services="import" slot="${hibernate.ogm.module.slot}" />
+        <module name="org.hibernate" slot="${hibernate.ogm.module.slot}" />
         <module name="org.hibernate.hql" slot="${hibernate.hql.module.slot}" />
 
-        <module name="org.infinispan" optional="true" slot="${infinispan.module.slot}" />
-        <module name="org.infinispan.commons" optional="true" slot="${infinispan.module.slot}" />
+        <module name="org.infinispan" slot="${infinispan.module.slot}" />
```
gunnarmorling: Strange that it was optional before?!

gunnarmorling:
Merged; Added one commit to make the CouchDB IT pass again by adding `services="import"` to the OGM core dependency in jboss-deployment-structure.xml.

modules/wildfly/src/main/modules/ogm/couchdb/module.xml
Sanne: Because it would export it to all users (end users too), which is not needed. It's cleaner to have each module which actually needs it to state so explicitly: even if it results in a slightly larger total number or lines across all xml files, it results in smaller dependency graphs.

Generally speaking I've learned that every time you need to declare "export" there is something very wrong, I'm using it only as temporary workarounds for bugs, or circumvent limitations of modules which are not under our control.
For example I'm thinking we should not export OGM at all, but the ORM should import it as "optional".

modules/wildfly/src/main/modules/ogm/infinispan/module.xml
Sanne: You just nailed the very reason for which I started to review them all ;-)
BTW my initial patch was quite more aggressive, but was getting complex quickly. Some more polishing will be possible with some more in-depth refactoring with how extensions and classloaders are handled, for example I don't think any of our modules should export resources as the end user couldn't care less of those.

Sanne:
> Added one commit to make the CouchDB IT pass again by adding services="import" to the OGM core dependency in jboss-deployment-structure.xml.

Thanks but strange. I'm pretty sure the PR I sent was working fine, I'll double-check

modules/wildfly/src/main/modules/ogm/couchdb/module.xml
Sanne: P.S. a legit usage of export is to have the OGM main module to export the ORM (vanilla) one.
We're pretty sure nobody would want to use OGM without ORM on his classpath :)
Likewise, for Hibernate Search to export Apache Lucene since its types are exposed on its public API.

Sanne:
> Maybe you didn't run the CouchDB IT locally? You need to export the COUCHDB_HOSTNAME env variable, otherwise the module is skipped.

right, forgot about that. the machine I was using for tests doesn't yet have a couch :)

modules/wildfly/src/main/modules/ogm/infinispan/module.xml
Sanne: > What exactly DYM by "resources" here?

Sorry meant services. The serviceloader definitions are resources, so the option actually exposes all resources from META-INF

PR #470, opened 2015-01-09
OGM-621 Extracting intermediary parent POM to avoid exposure of Guava an...

gunnarmorling:
...d other test dependencies on the public BOM

It's the best coming to my mind to avoid the leakage of Guava and the other stuff we use for testing through the public BOM. I thought about having an additional "test-only BOM" but that seemed more prone to being accidentally imported by users.

Let me know in case you have better ideas. Thx!

### DIFF
--- a/integrationtest/pom.xml
+++ b/integrationtest/pom.xml
@@ -1,9 +1,11 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
+
     <parent>
         <groupId>org.hibernate.ogm</groupId>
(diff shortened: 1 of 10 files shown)


integrationtest/pom.xml
```diff
@@ -1,9 +1,11 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
+
     <parent>
         <groupId>org.hibernate.ogm</groupId>
-        <artifactId>hibernate-ogm-parent</artifactId>
+        <artifactId>hibernate-ogm-module-parent</artifactId>
```
DavideD: Is this correct? It seems we don't have a hibernate-ogm-module-parent, did you forget a commit?

gunnarmorling:
Closing for now as I'm pursuing an alternative approach.

PR #472, opened 2015-01-09
OGM-401, OGM-640 Adding new supported types to type mapping lists per st...

gunnarmorling:
...ore


DavideD:
Should we add in the documentation the float can have rounding issues?
At the moment I can see that 10.4 is stored as 10.399999618530273 in MongoDB

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/472

DavideD:
THanks

gunnarmorling:
> Should we add in the documentation the float can have rounding issues?

I don't think that's needed here as it's a common issue with these types.

PR #474, opened 2015-01-13
OGM-715 Making "hibernate-hql-lucene" optional in core, adding it to tho...

gunnarmorling:
...se backend modules which don't have their own parser backend


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/474

PR #475, opened 2015-01-13
OGM-716 FIx distirbution package

DavideD:
Hello can you guys have alook a the resulting distribution package and let me know?

It should be similar to the previous one, except for:
- There is on optional folder containing parboiled and hibernate search dependencies
- core contains hibernate-ogm-core and all the required dependencies

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -103,6 +103,12 @@
                 <groupId>org.hibernate</groupId>
                 <artifactId>hibernate-envers</artifactId>
                 <version>${hibernateVersion}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>org.jboss.logging</groupId>
+                        <artifactId>jboss-logging-annotations</artifactId>
+                    </exclusion>
+                </exclusions>
             </dependency>
             <dependency>
                 <groupId>org.hibernate</groupId>
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -31,6 +31,8 @@
              than Neo4j core
         -->
         <neo4jCypherCompilerOlderVersion>2.0.3</neo4jCypherCompilerOlderVersion>
+        <!-- Neo4j does not support Lucene 4 currently-->
+        <neo4jLuceneVersion>3.6.2</neo4jLuceneVersion>
         <hibernateVersion>4.3.8.Final</hibernateVersion>
         <hibernateSearchVersion>5.0.0.Final</hibernateSearchVersion>
         <hibernateParserVersion>1.1.0.Final</hibernateParserVersion>
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -106,10 +106,6 @@
             <groupId>org.hibernate.ogm</groupId>
             <artifactId>hibernate-ogm-couchdb</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.hibernate</groupId>
-            <artifactId>hibernate-infinispan</artifactId>
-        </dependency>
 
         <!--
             The following dependencies are listed because their are optional or shared among different backends.
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -37,10 +37,124 @@
             <plugin>
                 <artifactId>maven-assembly-plugin</artifactId>
             </plugin>
+            <!-- 
+              Neo4j currently does not support Lucene 4 so we need to include Lucene 3 in the distribution package.
+
+              Maven only consider one version of a dependency so we copy the dependency in a separate folder and
+              we include it in the distribution package using the fileSet section in the assembly.
+             -->
+            <plugin>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>copy-model</id>
+                        <phase>prepare-package</phase>
+                        <goals>
+                            <goal>copy</goal>
+                        </goals>
+                        <configuration>
+                            <artifactItems>
+                                <artifactItem>
+                                    <groupId>org.apache.lucene</groupId>
+                                    <artifactId>lucene-core</artifactId>
+                                    <version>${neo4jLuceneVersion}</version>
+                                    <type>jar</type>
+                                </artifactItem>
+                           </artifactItems>
+                            <outputDirectory>${project.build.directory}/lib/neo4j</outputDirectory>
+                        </configuration>
+                    </execution>
+                </executions>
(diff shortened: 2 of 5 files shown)


bom/pom.xml
```diff
@@ -103,11 +105,12 @@
                 <groupId>org.hibernate</groupId>
                 <artifactId>hibernate-envers</artifactId>
                 <version>${hibernateVersion}</version>
-            </dependency>
```
gunnarmorling: Why did you remove this module from the BOM?

distribution/pom.xml
```diff
@@ -37,10 +37,97 @@
...
+            <groupId>org.hibernate.ogm</groupId>
+            <artifactId>hibernate-ogm-javadoc</artifactId>
+            <version>${project.version}</version>
+            <classifier>javadoc</classifier>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate.ogm</groupId>
+            <artifactId>hibernate-ogm-couchdb</artifactId>
+        </dependency>
+        <dependency>
```
gunnarmorling: Why is Envers needed?

distribution/pom.xml
```diff
@@ -37,10 +37,97 @@
...
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-envers</artifactId>
+        </dependency>
+
+        <dependency>
```
gunnarmorling: Out of interest, why are some of OGM's dependency listed here explicitly, but not all?

distribution/src/main/assembly/dist.xml
```diff
@@ -20,254 +20,139 @@
...
-            </binaries>
-        </moduleSet>
+        <!-- Provided libraries -->
+        <dependencySet>
+            <useProjectArtifact>false</useProjectArtifact>
+            <outputDirectory>dist/lib/provided</outputDirectory>
+            <useTransitiveFiltering>false</useTransitiveFiltering>
+            <unpack>false</unpack>
+            <includes>
+                 <include>org.jboss.logging:jboss-logging-annotations</include>
```
gunnarmorling: We don't need the annotations. They are only needed at OGM's build time (we somewhat ab-use "provided" to emulate a "compile-only" scope for this).

distribution/src/main/assembly/dist.xml
```diff
@@ -20,254 +20,139 @@
...
+                 <include>org.jboss.spec.javax.transaction:jboss-transaction-api_1.2_spec</include>
```
gunnarmorling: As we expect it provided by the environment, should we bundle provided stuff at all? Was it part of the dist before?

modules/wildfly/pom.xml
```diff
@@ -61,11 +61,6 @@
             <optional>true</optional>
         </dependency>
         <dependency>
-            <groupId>org.hibernate</groupId>
```
gunnarmorling: We don't need that anymore? Isn't it needed as part of putting the latest ORM version in place?

gunnarmorling:
@DavideD, Looks very good overall, added some detail remarks. Seems like it's even cleaner than before as all the transitive deps don't need to be listed anymore.

gunnarmorling:
The JPA API is under dist/lib/optional. Shouldn't it be under core?

bom/pom.xml
DavideD: I removed it because we don't use it anywhere? Is there any particular reason to keep it?

Anyway, I kept the change in a separate commit because I wasn't sure about it

distribution/src/main/assembly/dist.xml
DavideD: It was part of the dist before so I kept it, I'm fine removing it

distribution/src/main/assembly/dist.xml
DavideD: Yes, it was part of the dist before, that's why I kept it

modules/wildfly/pom.xml
DavideD: I dont know, none of our projects requires hibernate-infinispan and we never include it in modules or dist.

distribution/pom.xml
DavideD: I don't think it's needed I'll remove it

bom/pom.xml
gunnarmorling: The idea of the BOM is to give a user matching versions of all relevant
libs. Say you have an app using ORM@RDBMS + OGM and want to use Envers with
the former. Without that entry you'd have to match the Envers version with
the other ORM components defined by the BOM.

distribution/src/main/assembly/dist.xml
gunnarmorling: Ok, cool.

modules/wildfly/pom.xml
gunnarmorling: +1 for not adding it to the dist.

But in case of the app server ZIPs, omitting it would leave behind the user
with a "half" ORM version, potentially falling onto their feet in case
they'd like to use ISPN as L2 cache. Not sure whether anyone is doing it,
but I'd leave this as is.

bom/pom.xml
DavideD: Ok, I'll keep it in the BOM

distribution/pom.xml
DavideD: It's related to the way the assembly plugin works when `useTransitiveFiltering` is set to true.

Those are the dependencies that are optional or shared among different backends,
I need them as direct dependencies of the project to include them in separates folder using the dependencySet in the assembly.

modules/wildfly/pom.xml
DavideD: This is only the declaration of the dependency, we are not using it in the assembly that creates the module for WIldFly.
WildFly already contains hibernate-infinispan in the ORM module.

The only issue I can see is that WildFly 8.2 contains ORM 4.3.7 instead of 4.3.8 but these does not seems to cause problems at the moment.

This does not seem to cause issues, I would wait until they upgrade the dependencies in WIldFly

distribution/pom.xml
gunnarmorling: Ah, I see. Probably worth a short comment.

DavideD:
@gunnarmorling 
Hi, I've applied your remarks and squashed the commit:
- I've added back hibernate-infinispan in the pom
- I've moved jboss-logging and hibernate-jpa-2.1-api in core

PR #478, opened 2015-01-27
Resolves issue OGM-725

lordofthejars:
Resolves issue OGM-725 by implementing FongoDB DataStore Provider.

### DIFF
--- a/mongodb/pom.xml
+++ b/mongodb/pom.xml
@@ -12,6 +12,11 @@
     <name>Hibernate OGM for MongoDB</name>
     <description>Persist objects in MongoDB</description>
 
+    <properties>
+        <mongodbProvider>mongodb</mongodbProvider>
+        <slf4jVersion>1.7.10</slf4jVersion>
+    </properties>
+
     <build>
         <plugins>
             <plugin>
@@ -22,6 +27,9 @@
                         <MONGODB_HOSTNAME>${env.MONGODB_HOSTNAME}</MONGODB_HOSTNAME>
                         <MONGODB_PORT>${env.MONGODB_PORT}</MONGODB_PORT>
                     </environmentVariables>
+                    <systemPropertyVariables>
+                        <mongodbProvider>${mongodbProvider}</mongodbProvider>
+                    </systemPropertyVariables>
                     <forkMode>once</forkMode>
                 </configuration>
             </plugin>
@@ -37,6 +45,12 @@
                 <artifactId>maven-checkstyle-plugin</artifactId>
             </plugin>
         </plugins>
+        <testResources>
+            <testResource>
+                <directory>src/test/resources</directory>
+                <filtering>true</filtering>
+            </testResource>
+        </testResources>
     </build>
 
     <dependencies>
@@ -64,6 +78,17 @@
         <dependency>
             <groupId>org.parboiled</groupId>
             <artifactId>parboiled-java</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.github.fakemongo</groupId>
+            <artifactId>fongo</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+            <version>${slf4jVersion}</version>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.jboss.logging</groupId>
@@ -123,6 +148,26 @@
     </dependencies>
 
     <profiles>
+        <profile>
+            <id>fongo</id>
+            <properties>
+                <mongodbProvider>fongodb</mongodbProvider>
+            </properties>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <systemPropertyVariables>
+                                <mongodbProvider>${mongodbProvider}</mongodbProvider>
+                            </systemPropertyVariables>
+                            <forkMode>once</forkMode>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
         <profile>
             <id>embedded-mongodb</id>
             <activation>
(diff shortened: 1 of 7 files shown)


mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
```diff
@@ -168,4 +168,20 @@ private DB extractDatabase(MongoClient mongo, MongoDBConfiguration config) {
...
+
+	protected MongoDBConfiguration getConfig() {
+		return config;
+	}
+
+	protected MongoClient getMongoClient() {
+		return mongo;
+	}
+
+	protected void setMongoClient(MongoClient mongoClient) {
```
lordofthejars: I prefer to make setter and getter protected rather than attributes to have control on how something is set or get from children.

DavideD:
Jenkins, add to whitelist

emmanuelbernard:
@lordofthejars  I haven't looked at the code but you mentioned subclassing the `MongoDBDatastoreProvider`. Have you considered the idea of a composition approach instead.
A common core in a `CoreMongoDBDatabaseAccessor` (bad name), and have `MongoDBDatastoreProvider` and `FongoDatastoreProvider` instantiate and use `CoreMongoDBDatabaseAccessor` for the common operations

emmanuelbernard:
It might not be feasible but generally these composition approaches are less of a pain in the butt to maintain.

lordofthejars:
Yes I have thought this kind of design but because I am not a regular
contributors of hibernate ogm I didn't want to touch everything, I am going
to redesign the solution.
El dc., 28 de gen., 2015 a les 12.33 Emmanuel Bernard <
notifications@github.com> va escriure:

gunnarmorling:
@lordofthejars Thanks for this stuff!

I've ran the test suite and noticed it still unpacks and starts am MongoDB instance via the Flapdoodle plug-in (check out the "embedded-mongodb" profile). It'd be nice to find a way around that. If I actively suppress that profile (by specifying "-DuseExternalMongoDb" which is a bit hack-ish but does the trick), the `DatastoreInitializationTest` fails with Fongo. That should be excluded using `@SkipByGridDialect`.

mongodb/pom.xml
```diff
@@ -12,6 +12,11 @@
     <name>Hibernate OGM for MongoDB</name>
     <description>Persist objects in MongoDB</description>
 
+    <properties>
+        <mongodbProvider>mongodb</mongodbProvider>
+        <slf4jVersion>1.7.10</slf4jVersion>
```
gunnarmorling: Out of interest, we do you override the version? The one from the BOM seems to work fine.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
gunnarmorling: Maybe you can get rid of these when pursuing the delegation approach and pass in stuff via the constructor?

mongodb/pom.xml
```diff
@@ -66,6 +80,17 @@
...
         <dependency>
+            <groupId>com.github.fakemongo</groupId>
+            <artifactId>fongo</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+            <version>${slf4jVersion}</version>
+            <scope>provided</scope>
```
gunnarmorling: I feel this should rather be marked as optional. "Provided" suggests its needed at runtime (just provided by the environment), but that's not the case for the normal MongoDB case. I'd mark both deps as optional and add a short comment here. WDYT?

gunnarmorling:
@lordofthejars This looks very nice overall! I've just added some minor remarks here and there.

mongodb/pom.xml
lordofthejars: I haven't noticed that slf4j was already defined, sorry I fix that.

mongodb/pom.xml
lordofthejars: Ok.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
lordofthejars: I was thinking about using Template pattern (or decorator), but @emmanuelbernard thinks it would be better to not relay on extends and use aggregation. If you prefer I use Decorator pattern, no problem.

lordofthejars:
About Maven I will take a look because I have only executed the tests from mongodb project and they worked. I run by activating fongo profile which is the responsible of configuring fongo. 
To avoid downloading MongoDB if it is on a profile it seems that if it is not activated it should not be downloaded right?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
gunnarmorling: I don't have any preference pattern-wise. Only the class should be "effectively immutable" after its initialization, i.e. no changes to its state should be done after `configure()` has been called. The setters now suggests that the state also can be altered later on. Thus I'd prefer to not have them.

gunnarmorling: To give some background: `DatastoreProvider`s (and other services) are accessed concurrently from several threads at the same time. So their state (fields) must either be immutable or we need to synchronize access to it and make sure changes are visible to other threads (through volatile, synchronized etc.). Of course we strive for avoiding the latter for the sake of performance.

When it comes to immutability, the best is to make state real immutable (by means of setting it in the constructor and use final). But some contracts have a configure() or initialize() method which initializes state after the constructor invocation. Thus we cannot mark those fields as final. Nevertheless no synchronization is required and this is fine as long as these fields are not changed later on again (they are "effectively immutable") as the memory barrier established during building the session factory (which caused the initialization of datastore providers) ensures that these changes from the initialization phase are safely published to other threads.

The same wouldn't apply for a setter call in the course of the program after the initialization phase. Now it may be fine if the setter only actually is called during bootstrap (which I think it is), but that's not implied API-wise. So if we cannot get rid of it, we should at least document the conditions for its usage.

gunnarmorling:
> To avoid downloading MongoDB if it is on a profile it seems that if it is not activated it should not be downloaded right?

Yes, but the problem is that the "mongodb-embedded" profile is enabled by default. We'd need a way to disable it when using Fongo. Maybe something with several properties (should be possible as of Maven 3.2.2)?

gunnarmorling:
Hey @alexsnaps, I took these commits and sent a new PR (#483) which does some polishing. Specifically, we don't start the actual MongoDB via Maven any longer, when using the Fongo profile. I'll thus close this PR, your commits will get in with the other one.

Thanks a lot for your work!

PR #480, opened 2015-02-23
OGM-729 Pass correct number of arguments in debug call (Neo4J SchemaDefi...

emmanuelbernard:
...ner)

Also improve subsequent log to avoid information repetition


Sanne:
merged!

PR #482, opened 2015-02-23
OGM-733 Be defensive with our use of listDatabases in MongoDB

emmanuelbernard:
Restricted environment do not allow users to list all databases

This code is not "tested" because MongoDB role based security is a mystery to me.
If carefully reviewed, I'd say we should go for it anyways.

### Some links I used

[The user report](https://forum.hibernate.org/viewtopic.php?f=31&t=1038226)  
[Setting up a user admin](http://docs.mongodb.org/manual/tutorial/add-user-administrator/)  
[List of privileged actions](http://docs.mongodb.org/manual/reference/privilege-actions/)  
[Revoking roles fomr users](http://docs.mongodb.org/manual/reference/command/revokeRolesFromUser/#dbcmd.revokeRolesFromUser)  
[Built-in roles](http://docs.mongodb.org/manual/reference/built-in-roles/#built-in-roles)  
[Create a role](http://docs.mongodb.org/manual/tutorial/define-roles/)  
[Localhost bypass](http://docs.mongodb.org/manual/reference/parameters/)  

My user setting

```
> show users;
{
        "_id" : "admin.siteUserAdmin",
        "user" : "siteUserAdmin",
        "db" : "admin",
        "roles" : [
                {
                        "role" : "userAdminAnyDatabase",
                        "db" : "admin"
                }
        ]
}
> use ogm_test_database
switched to db ogm_test_database
> show users;
{
        "_id" : "ogm_test_database.user_read_write",
        "user" : "user_read_write",
        "db" : "ogm_test_database",
        "roles" : [
                {
                        "role" : "readWrite",
                        "db" : "ogm_test_database"
                }
        ]
}
```

### DIFF
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
@@ -149,13 +149,28 @@ public class MongoDBDatastoreProvider extends BaseDatastoreProvider implements S
 			String databaseName = config.getDatabaseName();
 			log.connectingToMongoDatabase( databaseName );
 
-			if ( !mongo.getDatabaseNames().contains( databaseName ) ) {
+			Boolean containsDatabase;
+			try {
+				containsDatabase = mongo.getDatabaseNames().contains( databaseName );
+			}
+			catch (MongoException me) {
+				// we don't have enough privileges, ignore the database creation
+				containsDatabase = null;
+			}
+
+			if ( containsDatabase != null && containsDatabase == Boolean.FALSE ) {
 				if ( config.isCreateDatabase() ) {
 					log.creatingDatabase( databaseName );
 				}
 				else {
 					throw log.databaseDoesNotExistException( config.getDatabaseName() );
 				}
+			}
+			DB db = mongo.getDB( databaseName );
+			if ( containsDatabase == null ) {
+				// force a connection to make sure we do have read access
+				// otherwise the connection failure happens during the first flush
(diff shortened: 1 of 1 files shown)


emmanuelbernard:
Oops, I pushed it as OGM-729 but it's OGM-733.
I have fixed the title and the commit. I left the branch as OGM-729 to avoid breaking the PR.

emmanuelbernard:
Looks like that patch works for the user

> I applied your update to my datastore provider and it worked as expected. 

Let's go!

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
```diff
@@ -157,6 +166,12 @@ private DB extractDatabase(MongoClient mongo, MongoDBConfiguration config) {
 					throw log.databaseDoesNotExistException( config.getDatabaseName() );
 				}
 			}
+			DB db = mongo.getDB( databaseName );
+			if ( containsDatabase == null ) {
+				// force a connection to make sure we do have read access
+				// otherwise the connection failure happens during the first flush
+				db.collectionExists( "WeDoNotCareWhatItIsWeNeedToConnect" );
```
gunnarmorling: Hm, there is no nicer way to do that, I guess.

gunnarmorling:
Rebased and applied. Thx.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
emmanuelbernard: There might be but I haven't found one. There is not explicit API nor a `select 1 = 1` like query either that I could find.

PR #483, opened 2015-02-24
OGM-725 Adding support for Fongo

gunnarmorling:
Based on @lordofthejars 's work, this PR adds support for working with [Fongo](https://github.com/fakemongo/fongo) as provider for MongoDB.

The additional commits make sure that MongoDB is not started via Maven when using the Fongo profile and disables one test which requires an actual MongoDB instance (`DatastoreInitializationTest`). For that purpose it's now possible to skip tests by datastore provider.


emmanuelbernard:
@gunnarmorling it's a bit unclear to me whether or not fongo is tested when we run `mvn clean install`.
If not should we add a special run in the CI (assuming we fix the mongodb instability first)?

gunnarmorling:
> it's a bit unclear to me whether or not fongo is tested when we run mvn clean install.

By default MongoDB is started/stopped via Maven as it used to be before. As you say we should add a job on CI which runs it against Fongo.

Sanne:
What is the goal for this? Is it just to make our testing more convenient, or do we want to make sure people using OGM with Fongo won't run into troubles?

gunnarmorling:
The goal is to enable OGM users to work with Fongo e.g. when testing their app.

emmanuelbernard:
Applied to upstream/4.2

emmanuelbernard:
Also created https://hibernate.atlassian.net/browse/WEBSITE-257

PR #484, opened 2015-02-24
OGM-739 Do not warn of incorrect shortname when a FQCN if given as datas...

emmanuelbernard:
...tore provider


gunnarmorling:
Rebased and applied. Thx!

PR #485, opened 2015-02-24
OGM-736 Move Neo4J + Transaction requirement as CAUTION

emmanuelbernard:
Also added to the site's FAQ in parallel

### DIFF
--- a/documentation/manual/src/main/asciidoc/en-US/modules/neo4j.asciidoc
+++ b/documentation/manual/src/main/asciidoc/en-US/modules/neo4j.asciidoc
@@ -752,8 +752,13 @@ The following summary will help you to keep track of the labels assigned to a ne
 [[ogm-neo4j-transactions]]
 === Transactions
 
-Neo4j operations can be executed only inside a transaction.
-Unless a different +org.hibernate.engine.transaction.jta.platform.spi.JtaPlatform+ is specified, OGM will use a specific implementation to integrate with the Neo4j transaction mechanism.
+[CAUTION]
+====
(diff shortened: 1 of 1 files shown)


documentation/manual/src/main/asciidoc/en-US/modules/neo4j.asciidoc
```diff
@@ -752,8 +752,13 @@ The following summary will help you to keep track of the labels assigned to a ne
 [[ogm-neo4j-transactions]]
 === Transactions
 
-Neo4j operations can be executed only inside a transaction.
-Unless a different +org.hibernate.engine.transaction.jta.platform.spi.JtaPlatform+ is specified, OGM will use a specific implementation to integrate with the Neo4j transaction mechanism.
+[CAUTION]
+====
+Neo4J operations must be executed inside a transaction.
+Make sure your interactions with Hibernate OGM are within a transaction when you target Neo4J.
```
gunnarmorling: What happens if no TX is running? Do we silently fail or raise a meaningful exception?

DavideD: If I remember correctly, Neo4j will throw an exception saying that you cannot execute an operation outside a transaction.

emmanuelbernard: Yes it depends. Some JPA operations must run within a transaction. For these ORM's code raises an exception.
But for read-only ops, Neo4J raises an exception. Check http://hibernate.org/ogm/faq/

PR #487, opened 2015-02-24
OGM-735 CouchDB byte[]

hferentschik:


couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/utils/BackendTckHelper.java
```diff
@@ -0,0 +1,25 @@
...
+ * Helper class allowing you to run all or any specified subset of test available on the classpath.
+ *
+ * This method is for example useful to run all or parts of the <i>backendtck</i>.
+ *
+ * @author Hardy Ferentschik
+ */
+@RunWith(ClasspathSuite.class)
+@ClasspathSuite.ClassnameFilters({ "org.hibernate.ogm.backendtck.*" })
+//@ClasspathSuite.ClassnameFilters({ ".*CRUDTest" })
+public class BackendTckHelper {
```
DavideD: I don't think I understand what this class does and why you need it only in CouchDB.

What can you do with this class that you could not do before?

couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/test/crud/EntityWithByteArrayTest.java
```diff
@@ -0,0 +1,100 @@
...
+import org.hibernate.ogm.utils.OgmTestCase;
+import org.hibernate.ogm.utils.TestForIssue;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * @author Hardy Ferentschik
+ */
+public class EntityWithByteArrayTest extends OgmTestCase {
```
DavideD: We actually have a test that checks all the built-in types: BuiltInTypeTest

Any particular reason to add a new one?

couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/utils/BackendTckHelper.java
emmanuelbernard: same question here, I'm lost :)

hferentschik: Run/Debug the backendtck test within the scope of the couchdb module from within the IDE. It is just a helper. This is the alternative to copying a test from the tck into the module in order to run it.

DavideD: As far as I understand, I still need to update this file everytime I want to run a specific test.
I prefer to create a new class that extends the test class I want to run.

I would prefer not to add a new library for this but I would like to know what everyone else think about it

couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/test/crud/EntityWithByteArrayTest.java
hferentschik: > Any particular reason to add a new one?

Several actually.

* I was not aware of the test 
* It did not fail which was another reason I did not see it. Seems in the test the byte[] is mapped as a `@Lob` as well, so I guess that makes the difference. I think the mapping without `@Lob` should be valid as well
* Looking at `BuiltInTypeTest` I have to say I am not a big fan of this test all in one go tests. I find them hard to understand and maintain.  I am much more for small dedicated tests
* The test I wrote just addresses the actual JIRA issue in a minimal test case. Since this was a CouchDB test I added the test case there. I guess one could argue that a backendtck test for byte[] would make sense as well, but  I see that as an add on to this dedicated test

couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/test/crud/EntityWithByteArrayTest.java
```diff
@@ -0,0 +1,100 @@
...
+	private Session testSession;
+	private byte[] testData;
+
+	@Before
+	public void setUp() {
+		super.setUp();
+		testSession = openSession();
+
+		testData = new byte[20];
+		new Random().nextBytes( testData );
```
DavideD: I usually avoid the use of  random values as they make the test not really repeatable

couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/utils/BackendTckHelper.java
hferentschik: > As far as I understand, I still need to update this file everytime I want to run a specific test.

If you want to run a specific test, yes. To run the whole suite, you can just run it as is. It is also a helper class, nothing happens during the default build.

> I prefer to create a new class that extends the test class I want to run.

So you need to make modifications as well. Mine seems a bit more transparent.

> I would prefer not to add a new library for this but I would like to know what everyone else think about it

It is a test library. Just another JUnit custom runner. I don't see an issue here. @gunnarmorling introduced the TestNG equivalent into Validator last week and it makes sense imo. It makes working in the IDE easier.

couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/test/crud/EntityWithByteArrayTest.java
hferentschik: I compare the bye array before persisting and then after persisting/retrieval one by one. They need to match. That's all I care about.

To a certain degree you might have a point, but I don't think it matters in this case. If it makes you happy I create a fixed byte array.

couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/test/crud/EntityWithByteArrayTest.java
DavideD: > It did not fail which was another reason I did not see it. Seems in the test the byte[] is mapped as a @Lob as well, so I guess that makes the difference. I think the mapping without @Lob should be valid as well

As you said, the test didn't fail because the use case is different from the one we are testing

> Looking at BuiltInTypeTest I have to say I am not a big fan of this test all in one go tests. I find them hard to understand and maintain. I am much more for small dedicated tests

Yes, BuiltInTypeTest.testTypesSupport is pretty big and I think it would be nice to refactor it.
We could probably separte each assert in it's own unit test.
That said, I think for the issue you are working at the moment it would be better to be consistent with what we have now.

> The test I wrote just addresses the actual JIRA issue in a minimal test case. Since this was a CouchDB test I added the test case there. I guess one could argue that a backendtck test for byte[] would make sense as well, but I see that as an add on to this dedicated test

The user discovered the exception using CouchDB but we should check and support built-in types on all the other datastores as well. It's very likely that it will work out of the box. If it doesn't it should be easy to fix. We have several dedicated test in in the project but I don't think this is a good use case for one.

emmanuelbernard: Ah yes, we should create a dedicated issue for supporting `byte[]` without a `@Lob` for all datastores. I did not think that was supported and indeed JPA's spec is not clear. But Hibernate ORM itself supports `byte[]` without `@Lob`.

emmanuelbernard: Hum ok. Open an issue for improving the `BuiltInTypeTest`. We have a bunch of those improvements in the pipeline. I don't think it's all that important to address it now.

couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/utils/BackendTckHelper.java
DavideD: > As far as I understand, I still need to update this file everytime I want to run a specific test.
> 
> If you want to run a specific test, yes. To run the whole suite, you can just run it as is. It is also a helper class, nothing happens during the default build.

I can run all the test using maven (from the ide as well) so nothing change for me in that case.




> It is a test library

For me it's still an additional library that we need and that might cause problem. 

Anyway, I guess if noone else as problem with it we can add it. I just wanted to understand the huge benefits of using it.

emmanuelbernard: Interesting, I did not really realised you could run the tests for maven from your IDE.
I tried to put debug points and the test suite does not seem to stop on them.
So in case you want to run the full test suite and debug that might be useful.

I use the vanilla support from IntelliJ IDEA to set the classpath to the right module, so I won't sue that facility personally.

DavideD: I don't know why the debug does not work (I remember trying it in the past)
but how often do you run all the tests in the project to check to debug a
failure?

couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/test/crud/EntityWithByteArrayTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+	@Override
+	protected Class<?>[] getAnnotatedClasses() {
+		return new Class<?>[] { Foo.class };
+	}
+
+	@Entity
+	public static class Foo {
+		@Id
+		@GeneratedValue(generator = "uuid")
```
gunnarmorling: We tend to mostly use self-assigned ids ("foo-1") in recent times. Doesn't make a big difference here, but it helps largely when debugging tests which involve several entities of the same type and/or associations between entities.

couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/test/crud/EntityWithByteArrayTest.java
gunnarmorling: > Ah yes, we should create a dedicated issue for supporting byte[] without a @Lob for all datastores.

Isn't the issue at hand the right one to do it? If it's an issue at other stores as well, I'd generalize the issue and fix it for all. 

That also allows to add the test to `BuiltInTypeTest` (agreed it's not perfect but it's what we have for now for testing data types across all backends).

emmanuelbernard: > Isn't the issue at hand the right one to do it? If it's an issue at other stores as well, I'd generalize the issue and fix it for all.

Yes that's what I should have said and thought about :)

hferentschik: > Yes, BuiltInTypeTest.testTypesSupport is pretty big and I think it would be nice to refactor it.
We could probably separate each assert in it's own unit test.

Separating the assertions is probably a good start. One can still use a single test entity, but at least separate the different assertions.

> That said, I think for the issue you are working at the moment it would be better to be consistent with what we have now.

I don't see how consistency is broken. IMO are there two different concerns. One is the issue I was working on. I think it deserves it's own dedicated test case referenced by `@TestForIssue`. It also should live in the CouchDB module. 

I also agree that we need a "tck" test in case we want to make sure this case works for all datastores, but most likely this should be a new issue. And even if we address it within the same issue, I'd argue that there is no harm in keeping two tests verifying the same thing. The tck and the actual module test fulfill different purposes imo. We have the same in Validator. We have the BV TCK and the actual tests in the engine as well. There is a huge overlap.

hferentschik: > Ah yes, we should create a dedicated issue for supporting byte[] without a @Lob for all datastores. I did not think that was supported and indeed JPA's spec is not clear. But Hibernate ORM itself supports byte[] without @Lob.

+1 for creating a more general issue for supporting this on all data stores, provided we agree that it is valid. I goes if we decided to make it not valid, it would be problematic anyways, right? One would in this case expect some sort of mapping exception, but the configuration (annotation parsing etc) is done in ORM code, right? There is no simple way for us to change/modify that?

hferentschik: > Hum ok. Open an issue for improving the BuiltInTypeTest. We have a bunch of those improvements in the pipeline. I don't think it's all that important to address it now.

Not sure whether it needs a dedicated issue. If we create an issue for supporting byte[] in general, it would be a given for me to refactor the test first.

hferentschik: > Yes that's what I should have said and thought about :)

Seem @emmanuelbernard is weak. I stand to my take on this. Just generalizing a user created specific issue seems wrong to me.

gunnarmorling: > I'd argue that there is no harm in keeping two tests verifying the same thing.

It creates increased efforts for maintaining two tests and increases runtime of the test suite. Generally we prefer to have tests in the TCK (they make sure that things work as expected across all backends), and only for backend-specific stuff have tests in these modules (e.g. for specific options).

couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/utils/BackendTckHelper.java
hferentschik: > I use the vanilla support from IntelliJ IDEA to set the classpath to the right module, so I won't sue that facility personally.

That works as well. I do that as well sometimes, but I find this easier.

hferentschik: > I don't know why the debug does not work (I remember trying it in the past) but how often do you run all the tests in the project to check to debug a failure?

Well I won't run the full suite in debug mode. On the other hand, it is nice to run the whole suite from within the IDE once in a while in "normal" test mode.

hferentschik: Anyways, I don't have hard feeling about this one. I find it nice to have and imo there is no harm in adding it, but if you guys think it is not worth it, drop the commit.

couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/test/crud/EntityWithByteArrayTest.java
hferentschik: > It creates increased efforts for maintaining two tests and increases runtime of the test suite. 

Which are both minimal.

If I move this test into the TCK I  have to potentially fix a whole bunch of other things, which are not related to the user issue.

gunnarmorling: > Just generalizing a user created specific issue seems wrong to me.

It's what we often do. Naturally people notice certain shortcomings only on the store(s) they use. IMO it makes sense to handle and fix it generically if we find out it applies to all stores actually.

hferentschik: > IMO it makes sense to handle and fix it generically if we find out it applies to all stores actually.

Sure,  I am not saying that it should not be fixed for the other stores, but I don't like modifying or changing the scope of a specific user issue.

This could for example be merged and fix the specific user issue. If I increase the scope of the issue by trying to fix it for all datastores, it might not get ready for the next release. Or maybe we say that the support for byte[] has a lower priority than other issue and we should do something else first. In this case there is again no immediate fix for this particular issue even so it is ready.

Supporting byte[] for all data stores is a different beast to supporting it for CouchDB only and addressing this particular issue. 

Also, the whole idea of not duplicating tests is hard to keep up anyways, since it would require at all times to know what and where things are tested.

DavideD: > Just generalizing a user created specific issue seems wrong to me.
> 
> It's what we often do. Naturally people notice certain shortcomings only on the store(s) they use. IMO it makes sense to handle and fix it generically if we find out it applies to all stores actually.

I don't think we can ask the user to install all the supported datastores and check if the problem happens somewhere else before opening a JIRA, It's up to us to recognize the features that must be consistent across the datastores.

> Which are both minimal.

But it's unecessary, and it can be easly avoided with something that we already have.

couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/utils/BackendTckHelper.java
DavideD: > On the other hand, it is nice to run the whole suite from within the IDE once in a while in "normal" test mode.

My point is that you can already do that using maven in the IDE or in the terminal.

couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/test/crud/EntityWithByteArrayTest.java
DavideD: > Sure, I am not saying that it should not be fixed for the other stores, but I don't like modifying or changing the scope of a specific user issue.

I can do that if it makes you feel better :)

hferentschik: > I don't think we can ask the user to install all the supported datastores and check if the problem happens somewhere else before opening a JIRA, It's up to us to recognize the features that must be consistent across the datastores.

Exactly. So if I were a general contributor and contributed this fix together with the issue you would probably accept it, right?  You then would probably create a new issue to make sure it works on all data stores (provided you think it is required).

couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/utils/BackendTckHelper.java
hferentschik: Triggering a maven build execution from within the IDE is different to running the tests directly in the IDE (maybe not if you are just interested in the results, but for debugging for sure).

And yes, I also driving the build regularly from the command line. This is just a convenience. All it is.

PR #488, opened 2015-02-25
OGM-735 Handling of byte[] array property type w/o explicit @Lob annotation

hferentschik:


neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/utils/Neo4jBackendTckHelper.java
```diff
@@ -0,0 +1,25 @@
...
+ * Helper class allowing you to run all or any specified subset of test available on the classpath.
+ *
+ * This method is for example useful to run all or parts of the <i>backendtck</i>.
+ *
+ * @author Hardy Ferentschik
+ */
+@RunWith(ClasspathSuite.class)
+//@ClasspathSuite.ClassnameFilters({ "org.hibernate.ogm.backendtck.*" })
+@ClasspathSuite.ClassnameFilters({ ".*BuiltInTypeTest" })
+public class Neo4JBackendTckHelper {
```
gunnarmorling: This seems to cause a compile error (note the upper-case "J" vs. lower case in the file name).

hferentschik: > This seems to cause a compile error (note the upper-case "J" vs. lower case in the file name).

I know. I had it initially lower cased, but then changed it to upper case. However, Mac's genius file system comes into my way here.

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/488

hferentschik:
Thanks :-)

DavideD:
Thank you

PR #489, opened 2015-02-25
OGM-674 Unset null embeddable objects in mongodb

emmanuelbernard:


core/src/test/java/org/hibernate/ogm/backendtck/embeddable/EmbeddableTest.java
```diff
@@ -292,6 +292,56 @@ public void testPersistWithListEmbeddedInNestedComponent() throws Exception {
 		session.close();
 	}
 
+	@Test
+	public void testNestedEmbeddedWithNullProperties() {
+		final Session session = openSession();
+
+		// persist entity without the embeddables
```
gunnarmorling: s/without/with/

core/src/test/java/org/hibernate/ogm/backendtck/embeddable/EmbeddableTest.java
```diff
@@ -292,6 +292,56 @@ public void testPersistWithListEmbeddedInNestedComponent() throws Exception {
...
+		Account loadedAccount = (Account) session.get( Account.class, account.getLogin() );
+		loadedAccount.getHomeAddress().setType( null );
+		transaction.commit();
+		session.clear();
+
+		// read back nested embedded and set regular embedded to null
+		transaction = session.beginTransaction();
+		loadedAccount = (Account) session.get( Account.class, account.getLogin() );
+		assertThat( loadedAccount ).as( "Cannot load persisted object with nested embeddedables" ).isNotNull();
+		assertThat( loadedAccount.getHomeAddress() ).isNotNull();
```
gunnarmorling: Should we have a test in _mongodb_ which asserts the actual persistent structure at this point? I think that'd make sense. There are some *Mapping tests in _mongodb_.

gunnarmorling: Ah, I should have looked till the end. There is one :)

core/src/test/java/org/hibernate/ogm/backendtck/embeddable/EmbeddableTest.java
emmanuelbernard: fixed

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/dialect/impl/EmbeddableStateFinder.java
```diff
@@ -0,0 +1,132 @@
...
+	 *
+	 * Returns the most outer embeddable containing {@code column} that is entirely null.
+	 * Return null otherwise i.e. not embeddable.
+	 *
+	 * The implementation lazily compute the embeddable state and caches it.
+	 * The idea behind the lazy computation is that only some columns will be set to null
+	 * and only in some situations.
+	 * The idea behind caching is that an embeddable contains several columns, no need to recompute its state.
+	 */
+	public String getMostOuterNullEmbeddable(String column) {
```
gunnarmorling: Ok, took me a bit to wrap my head around it :)

I think it is sound functionally, but that impl is tough to grok. The many inline comments make me think that this single method is too long and too complex.

See https://github.com/gunnarmorling/hibernate-ogm/commit/259e4e9c05bd7128c194ec1cc851a147af0fd2fd how it could be decomposed into several methods which are much friendlier IMO. The method names make most of the inline comments superfluous and let you digest one piece a time.

core/src/test/java/org/hibernate/ogm/backendtck/embeddable/EmbeddableTest.java
```diff
@@ -292,6 +292,56 @@ public void testPersistWithListEmbeddedInNestedComponent() throws Exception {
...
+
+
+		// set nested embedded to null
+		transaction = session.beginTransaction();
+		Account loadedAccount = (Account) session.get( Account.class, account.getLogin() );
+		loadedAccount.getHomeAddress().setType( null );
+		transaction.commit();
+		session.clear();
+
+		// read back nested embedded and set regular embedded to null
```
gunnarmorling: I would split this into several methods. This will make it easier to understand what's wrong if one of the cases fails.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/dialect/impl/EmbeddableStateFinder.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.hibernate.ogm.dialect.spi.TupleContext;
+import org.hibernate.ogm.model.spi.Tuple;
+
+/**
+ * Captures useful data around the state of embeddable objects in tuples
+ * Note that the current implementation is *not* specific to MongoDB on purpose.
```
gunnarmorling: Should we move it to `org.hibernate.ogm.datastore.document` in core? That way could e.g. re-use it in CouchDB. Wouldn't have to be an "official" SPI, we refer to some "impl" stuff from core in our dialects, I think that's fine for now.

gunnarmorling:
@emmanuelbernard Looks good to me overall. I would only arrange some pieces a bit differently, see my comments above.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/dialect/impl/EmbeddableStateFinder.java
emmanuelbernard: ah yes I did not think of this approach, we can do that. Keeping it in impl but shared in core.

emmanuelbernard:
@gunnarmorling thanks, I'll process your proposals later this pm.

core/src/test/java/org/hibernate/ogm/backendtck/embeddable/EmbeddableTest.java
emmanuelbernard: It would be a bit slower and the benefit is not that significant. I'll skip this one.

emmanuelbernard:
@gunnarmorling I have applied your minor comments and changed the algorithm to look like the one you proposed instead. Do a quick second look and this can ship.

PR #490, opened 2015-02-26
Jdk9

Sanne:
I'm sending the PR against master as I'm confused about the branches. My guess is you'll want to apply this to the 4.2 branch, or maybe both?

I guess at least the patches related to requiring Java 7 should not be merged in whatever we're using for branch 4.1

 - https://hibernate.atlassian.net/browse/OGM-747
 - https://hibernate.atlassian.net/browse/OGM-752
 - https://hibernate.atlassian.net/browse/OGM-753

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -95,12 +95,6 @@
         <dependency>
             <groupId>org.hibernate.ogm</groupId>
             <artifactId>hibernate-ogm-neo4j</artifactId>
-       </dependency>
-        <dependency>
-            <groupId>org.hibernate.ogm</groupId>
-            <artifactId>hibernate-ogm-javadoc</artifactId>
-            <version>${project.version}</version>
-            <classifier>javadoc</classifier>
         </dependency>
         <dependency>
             <groupId>org.hibernate.ogm</groupId>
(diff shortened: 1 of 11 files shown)


gunnarmorling:
I will only apply it to 4.2 as we shouldn't alter the JDK requirement with a bugfix release.

distribution/pom.xml
```diff
@@ -172,4 +166,26 @@
              </exclusions>
         </dependency>
     </dependencies>
+
+    <profiles>
+        <profile>
+            <id>doc</id>
```
gunnarmorling: Shouldn't this be "dist" here?

documentation/manual/src/main/asciidoc/en-US/modules/getting-started.asciidoc
```diff
@@ -8,7 +8,7 @@ and retrieving an entity using Hibernate OGM.
 
 Before we can start, make sure you have the following tools configured:
 
-* Java JDK 6 or above
+* Java JDK 7 or above
 * Maven 3.x
```
gunnarmorling: I will change this into the specific requirement given in the POM.

distribution/pom.xml
Sanne: No. _dist_ is needed to include the module which contains this pom, but this profile needs to be in synch with the _doc_ profile to make sure the javadocs module it adds it the correct one.

Otherwise it might fail as there is no such valid javadoc module (say you just changed version and didn't build it), or you risk including an stale version of the documentation.

By making the dependency valid only when it's build, that also prevents Eclipse to freak out.

documentation/manual/src/main/asciidoc/en-US/modules/getting-started.asciidoc
Sanne: ok if you want but we never went so far. Normally JDK's don't change too often, and you'll need to double-check various docs and websites either way.

distribution/pom.xml
gunnarmorling: Ok, cool.

documentation/examples/pom.xml
```diff
@@ -7,7 +7,6 @@
     <relativePath>../pom.xml</relativePath>
   </parent>
   <artifactId>hibernate-ogm-documentation-examples</artifactId>
-  <version>4.1.2-SNAPSHOT</version>
```
gunnarmorling: Good catch.

pom.xml
```diff
@@ -443,6 +443,16 @@
                     <version>1.0.2</version>
                 </plugin>
                 <plugin>
+                    <groupId>com.github.joelittlejohn.embedmongo</groupId>
```
gunnarmorling: This one was specified already. I'll remove it.

gunnarmorling:
Rebased and merged to 4.2 Thx!

PR #492, opened 2015-02-27
OGM-692 Query on embedded properties for Neo4j (and MongoDB)

DavideD:
https://hibernate.atlassian.net/browse/OGM-692
Rebased for 4,2 branch

To create the queries in Neo4j I need to create and keep track of aliases for the various embedded nodes using a tree structure. Alias are created starting form the alias of the entity and adding the position of the
node as a suffix.

For example, an embedded represented in the query as *n.embedded.anotherEmbedded* will generate the following aliases *n_0* for (n.embedded) and *n_0_0* for (n.embedded.anotherEmbedded).

The following query
```
FROM WithEmbedded e
WHERE e.anEmbeddable.anotherEmbeddable.embeddedString = 'string 2'
``` 
becomes:
```
MATCH (e:WithEmbedded)
OPTIONAL MATCH (e) -[:anEmbeddable]-> (`e_0`:EMBEDDED) -[:anotherEmbeddable]->(`e_0_0`:EMBEDDED)
WHERE `e_0_0`.embeddedString="string 2"
RETURN e
```

I noticed that we didn't have enough tests for queries on embedded and I decided to extract the existing one, move them in a separate class and add missing use cases. This caused some errors in MongoDB that should be now solved.

In the process I also discovered [OGM-751](https://hibernate.atlassian.net/browse/OGM-751)

### DIFF
--- a/core/src/test/java/org/hibernate/ogm/backendtck/queries/QueriesWithEmbeddedTest.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/queries/QueriesWithEmbeddedTest.java
@@ -1,0 +1,173 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.backendtck.queries;
+
+import static org.hibernate.ogm.utils.OgmAssertions.assertThat;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import org.hibernate.Session;
+import org.hibernate.SessionFactory;
+import org.hibernate.Transaction;
+import org.hibernate.ogm.utils.OgmTestCase;
+import org.hibernate.ogm.utils.TestSessionFactory;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+/**
+ * @author Sanne Grinovero &lt;sanne@hibernate.org&gt; (C) 2012 Red Hat Inc.
+ * @author Gunnar Morling
+ * @author Davide D'Alto
+ */
+public class QueriesWithEmbeddedTest extends OgmTestCase {
+
+	@TestSessionFactory
+	public static SessionFactory sessions;
+
+	@Rule
+	public ExpectedException thrown = ExpectedException.none();
+
+	private Session session;
+
+	private Transaction tx;
+
+	@Before
+	public void createSession() {
+		closeSession();
+		session = sessions.openSession();
+		tx = session.beginTransaction();
+	}
+
+	@After
+	public void closeSession() {
+		if ( tx != null && tx.isActive() ) {
+			tx.commit();
+			tx = null;
+		}
+		if ( session != null ) {
+			session.close();
+			session = null;
+		}
+	}
+
+	@Test
+	public void testQueryWithEmbeddableInWhereClause() throws Exception {
+		List<?> result = session.createQuery( "from WithEmbedded e where e.anEmbeddable.embeddedString = 'string 1'" ).list();
+		assertThat( result ).onProperty( "id" ).containsOnly( 1L );
+	}
+
+	@Test
+	public void testQueryWithNestedEmbeddableInWhereClause() throws Exception {
+		List<?> result = session.createQuery( "from WithEmbedded e where e.anEmbeddable.anotherEmbeddable.embeddedString = 'string 2'" ).list();
+		assertThat( result ).onProperty( "id" ).containsOnly( 1L );
+	}
+
+	@Test
+	public void testQueryWithEmbeddablePropertyInSelectClause() throws Exception {
+		List<ProjectionResult> result = asProjectionResults( "select e.id, e.anEmbeddable.embeddedString from WithEmbedded e" );
+		assertThat( result ).containsOnly( new ProjectionResult( 1L, "string 1" ) );
+	}
+
+	@Test
+	public void testQueryReturningEmbeddedObject() {
+		List<?> list = session.createQuery( "from WithEmbedded we" ).list();
+
+		assertThat( list )
+			.onProperty( "anEmbeddable" )
+			.onProperty( "embeddedString" )
+			.containsExactly( "string 1" );
+
+		assertThat( list )
+			.onProperty( "anEmbeddable" )
+			.onProperty( "anotherEmbeddable" )
+			.onProperty( "embeddedString" )
+			.containsExactly( "string 2" );
+	}
+
+	@BeforeClass
+	public static void insertTestEntities() throws Exception {
+		final Session session = sessions.openSession();
+		Transaction transaction = session.beginTransaction();
+
+		WithEmbedded with = new WithEmbedded( 1L, new AnEmbeddable( "string 1", new AnotherEmbeddable( "string 2" ) ) );
+		session.persist( with );
+
+		transaction.commit();
+		session.close();
+	}
+
+	private List<ProjectionResult> asProjectionResults(String projectionQuery) {
+		List<?> results = session.createQuery( projectionQuery ).list();
+		List<ProjectionResult> projectionResults = new ArrayList<ProjectionResult>();
+
+		for ( Object result : results ) {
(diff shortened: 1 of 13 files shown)


core/src/test/java/org/hibernate/ogm/backendtck/queries/QueriesWithEmbeddedTest.java
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	@Test
+	public void testQueryWithInOperator() throws Exception {
+		List<?> result = session.createQuery( "from WithEmbedded e where e.anEmbeddable.embeddedString IN ( 'string 1' )" ).list();
+		assertThat( result ).onProperty( "id" ).containsOnly( 1L );
+	}
+
+	@Test
+	@Test
```
gunnarmorling: It's not repeatable :)

DavideD:
Jenkins, retest this please

core/src/test/java/org/hibernate/ogm/backendtck/queries/QueriesWithEmbeddedTest.java
DavideD: Removed a commit at the last minute. Shoudl be fixed now

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/query/parsing/impl/Neo4jQueryRendererDelegate.java
```diff
@@ -71,6 +76,35 @@ private void match(StringBuilder queryBuilder, String targetAlias, String label)
...
+			}
+		}
+	}
+
+	private void appendEmbedded(StringBuilder queryBuilder, String optionalMatch, EmbeddedAliasTree node) {
+		if ( node.getChildren().isEmpty() ) {
+			queryBuilder.append( optionalMatch );
+		}
+		else {
+			int index = 0;
```
gunnarmorling: This variable and the next seem unused.

gunnarmorling:
MATCH (e:WithEmbedded)
    OPTIONAL MATCH (e) -[:anEmbeddable]-> (`e_0`:EMBEDDED) -[:anotherEmbeddable]->(`e_0_0`:EMBEDDED)
    WHERE `e_0_0`.embeddedString="string 2"
    RETURN e

It may be a stupid question, but why is the complex aliasing needed? Couldn't one just assign subsequent numbers to the nodes along the path?

DavideD:
No, it's not sa stupid question.

I think this part is more complicated than it should be. I'll change it.

core/src/test/java/org/hibernate/ogm/backendtck/queries/QueriesWithEmbeddedTest.java
```diff
@@ -0,0 +1,209 @@
...
+			session.persist( entity );
+		}
+
+		transaction.commit();
+		session.close();
+	}
+
+	private List<ProjectionResult> asProjectionResults(String projectionQuery) {
+		List<?> results = session.createQuery( projectionQuery ).list();
+		List<ProjectionResult> projectionResults = new ArrayList<ProjectionResult>();
```
gunnarmorling: Allocating it with results.size() will avoid subsequent re-sizings.

gunnarmorling: Ah, test-only. Not that important then :-)

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/query/parsing/impl/Neo4jQueryRendererDelegate.java
```diff
@@ -71,6 +76,35 @@ private void match(StringBuilder queryBuilder, String targetAlias, String label)
 		node( queryBuilder, targetAlias, label );
 	}
 
+	private void optionalMatch(StringBuilder queryBuilder, String targetAlias) {
```
gunnarmorling: What part of the query does this method exactly add?

I'm confused a bit by the general organization of the logic. I don't think that the renderer delegate should "know" the resolver delegate and invoke methods from it. Rather the alias handling should be part of the predicates IMO. The predicate factory could assign the right alias names depending on the "depth" of the predicate to be created. 

"Rendering" here should be more or less the invocation of the predicate tree root node to render itself, which in turn (recursively) asks all children to render themselves (see the call to `builder.build()` in `MongoDBQueryRendererDelegate#getResult`(. So e.g. "rendering" a `Neo4jComparisonPredicate` would yield the complete query part making up that comparison, including the node path in case its an embeddable. 

At least that was the original idea of the mechanism.

DavideD: >What part of the query does this method exactly add?

It the OPTIONAL MATCH in the cypher query, the part related to
```  
MATCH (e:WithEmbedded)
OPTIONAL MATCH (e) -[:anEmbeddable]-> (`e_0`:EMBEDDED) -[:anotherEmbeddable]->(`e_0_0`:EMBEDDED)
...
```

> I'm confused a bit by the general organization of the logic. I don't think that the renderer delegate should "know" the resolver delegate and invoke methods from it. Rather the alias handling should be part of the predicates IMO. The predicate factory could assign the right alias names depending on the "depth" of the predicate to be created

There are two places where you can refer to embeddedable at the moment, in the SELECT and in the predicates. I think the part related to the predicates work as you would expect: embedded are added when found via method `Neo4jPredicateFactory#alias(...)`.

The embedded aliases for the select clause get added when they get found via `Neo4jQueryRenderDelegate#setPropertyPath(...)`. I need to check both locations because you could refer to an embedded only in the SELECT clause.

It seems reasonable to me to leave the resolution of the aliases to the  
>"Rendering" here should be more or less the invocation of the predicate tree root node to render itself, which in turn (recursively) asks all children to render themselves (see the call to builder.build() in MongoDBQueryRendererDelegate#getResult(. So e.g. "rendering" a Neo4jComparisonPredicate would yield the complete query part making up that comparison, including the node path in case its an embeddable.

DavideD:
I've applied the remarks and rebased. Now the alias for an embedded looks like *_n1*

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/query/parsing/impl/Neo4jQueryRendererDelegate.java
gunnarmorling: > I need to check both locations because you could refer to an embedded only in the SELECT clause.

Understood. I think in the long term the different parts of the query should be handled in the same way. I.e. there should be one object tree representing the entire query. It has been handled for MongoDB differently because we don't have a real query language there. That's why the builder there only applies to the WHERE clause while the SELECT is handled differently. I think for Neo4j this could be unified. But I think we can address this independently from the change at hand.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/query/parsing/impl/EmbeddedAliasTree.java
```diff
@@ -0,0 +1,84 @@
...
+ * Tree structure that can be used to store alias of embedded components in Neo4j.
+ * <p>
+ * For example, a path to two properties like:
+ * <ol>
+ * <li>n.first.anotherEmbedded</li>
+ * <li>n.second.anotherEmbedded</li>
+ * </ol>
+ * Might be represented with the following structure:
+ * <pre>
+ * n (alias = n)|- first (alias = n_0)  -- anotherEmbedded (alias = n_0_0)
```
gunnarmorling: This comment is not correct any longer I think?

gunnarmorling:
@DavideD, merged to 4.2 and rebased everything to the latest master.

Can you only send one more commit for that last JavaDoc comment issue. Thanks!

DavideD:
> @DavideD, merged to 4.2 and rebased everything to the latest master.
> 
> Can you only send one more commit for that last JavaDoc comment issue. Thanks!

Sure, I will send a new commit

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/query/parsing/impl/Neo4jQueryRendererDelegate.java
DavideD: > But I think we can address this independently from the change at hand.

Agree, I think we should handle this in a separate issue.

PR #494, opened 2015-03-02
OGM-465 - PREVIEW - SPI for obtaining applied operations of current flush cycle

gunnarmorling:
Hi @emmanuelbernard, this addresses UC 1.1 from [OGM-465](https://hibernate.atlassian.net/browse/OGM-465): "I want to log all operations applied so far and abort the processing of the flush cycle".

It's not complete yet, but I'd like to get some feedback on the overall approach. It should work well e.g. for CouchDB, where we apply ops one by one. But with MongoDB we have a problem due to the batching we apply. We can report all batches fully applied so far, but we cannot name the previously applied _batch elements_ if e.g. the 7th element of a batch fails.

This will be somewhat addressed if we provide `onFailedOperation()`. Then we'd present the current (failing) batch, and ideally the exception from the datastore will specify the element causing trouble. This should work for MongoDB, but it depends on the exact behavior of the dialects doing batching.


emmanuelbernard:
I guess I could dive more but that's enough feedback for a first pass @gunnarmorling . Plus we are probably better off discussing it irc 2 irc :)

gunnarmorling:
Closing this for now, I'll send a new one soon.

PR #495, opened 2015-03-03
OGM-775 Create an adaptor from JBoss Logging to Neo4J StringLogger

emmanuelbernard:
Used for query execution today

### DIFF
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/StringLoggerToJBossLoggingAdaptor.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/StringLoggerToJBossLoggingAdaptor.java
@@ -1,0 +1,83 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+
+package org.hibernate.ogm.datastore.neo4j.impl;
+
+import org.neo4j.helpers.collection.Visitor;
+import org.neo4j.kernel.impl.util.StringLogger;
+import org.neo4j.kernel.logging.LogMarker;
+
+import org.hibernate.ogm.datastore.neo4j.logging.impl.Log;
+import org.hibernate.ogm.datastore.neo4j.logging.impl.LoggerFactory;
+
+/**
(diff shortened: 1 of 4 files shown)


Sanne:
I was curious to understand more about this, but I see no tests. Not worth it or are they coming next?

emmanuelbernard:
No I wasn't planning on doing any test log.
To be fair, I've never seen the logs that @DavideD mentioned in my consoles.

gunnarmorling:
Instead of feeding the execution engine's log statements through our logger, shouldn't we rather make sure that the execution engine uses the same configuration as the `GraphDatabaseService`?

The latter uses a default configuration which causes messages to go to _<DATABASE_DIR>/messages.log. Having parts of Neo4j messages not there but in "our" log seems confusing to me.

Unfortunately it seems not possible for the execution engine to "inherit" the logging config from the database service. So we may

* resemble that default config by doing the same what `InternalAbstractGraphDatabase#createLogging()` does (requires some tracing where that config originally is created during the bootstrap)
* OR expose a setting for the `Logging` implementation (and `isDebugEnabled()`) 

pass in a user-customizable EmbeddedGraphDatabaseFactory.create()

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/StringLoggerToJBossLoggingAdaptor.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Implements Neo4J's StringLogger and write the content to JBoss Logging.
+ * Regular messages are sent as debug, user messages as info, exception messages as error
+ * and long messages as separated calls to logMessage.
+ * This follows Neo4J's {@code LogbackService} implementation.
+ *
+ * @author Emmanuel Bernard &lt;emmanuel@hibernate.org&gt;
+ */
+public class StringLoggerToJBossLoggingAdaptor extends StringLogger {
```
gunnarmorling: If you want to go down that path, also `isDebugEnabled()` should be overridden (delegating to `log`) so one can actually make the messages visible.

gunnarmorling:
@emmanuelbernard Any thoughts on my comment above?

emmanuelbernard:
@gunnarmorling yes, I am only coming back to it now. I have other things to do today due to the trip but I'll try and address it.

emmanuelbernard:
@gunnarmorling I think I found an acceptable solution. See second commit.
I also added the isDebugEnabled overriding you mentioned.

gunnarmorling:
Merged, thanks!

PR #496, opened 2015-03-04
OGM-265 Introduce failing test which stores an inner class using a table...

Sanne:
...-per-class strategy

This is a preview only, as there's something fishy with Neo4J still.

BTW, the new test `org.hibernate.ogm.backendtck.innertypes.InnerClassFindTest` - which I adapted from the draft on JIRA for this issue - is throwing this exception:

javax.transaction.SystemException: TxManager not recovered
	at org.neo4j.kernel.impl.transaction.TxManager.assertTmOk(TxManager.java:337)
	at org.neo4j.kernel.impl.transaction.TxManager.begin(TxManager.java:313)
	at org.neo4j.kernel.impl.transaction.TxManager.begin(TxManager.java:307)
	at org.hibernate.ogm.backendtck.innertypes.InnerClassFindTest.testInnerClassFind(InnerClassFindTest.java:48)

I haven't debugged it yet, but does it ring a bell?


gunnarmorling:
> I haven't debugged it yet, but does it ring a bell?

Not directly, but I'd try and extend `OgmTestCase`. This should do all the required preparational work, use the right config etc.

PR #497, opened 2015-03-04
OGM-265 Validate / Amend invalid collection names in MongoDB

Sanne:
- https://hibernate.atlassian.net/browse/OGM-265

Now ready for review, the Neo4J tests work again. As suggested by Gunnar, I've moved the validation function into a new contract, so we could add validation of more elements in the future.

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/dialect/spi/BaseDialectNamingAdaptor.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/spi/BaseDialectNamingAdaptor.java
@@ -1,0 +1,21 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.dialect.spi;
+
+public class BaseDialectNamingAdaptor implements DialectNamingAdaptor {
+
+	static final DialectNamingAdaptor DEFAULT_INSTANCE = new BaseDialectNamingAdaptor();
+
+	public BaseDialectNamingAdaptor() {
+	}
+
+	@Override
+	public String makeValidTableName(String requestedName) {
+		return requestedName;
+	}
+
+}
--- a/core/src/main/java/org/hibernate/ogm/dialect/spi/BaseDialectNamingAdaptor.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/spi/BaseDialectNamingAdaptor.java
@@ -6,15 +6,31 @@
  */
 package org.hibernate.ogm.dialect.spi;
 
+import org.hibernate.ogm.util.impl.Contracts;
+
+/**
+ * Base implementation of {@link DialectNamingAdaptor}.
+ *
+ * It's generally a good idea to extend this class to create a new {@link DialectNamingAdaptor} implementation,
+ * however that's optional. The benefit of extending this class is that you won't have compilation errors
+ * in case we add new methods to the contract of the {@link DialectNamingAdaptor} interface in future versions.
+ *
+ * @author Sanne Grinovero
+ */
 public class BaseDialectNamingAdaptor implements DialectNamingAdaptor {
 
 	static final DialectNamingAdaptor DEFAULT_INSTANCE = new BaseDialectNamingAdaptor();
 
-	public BaseDialectNamingAdaptor() {
+	protected BaseDialectNamingAdaptor() {
+		// Shouldn't be needed to be invoked, except by extensions
 	}
 
+	/**
+	 * {@inheritDoc}
+	 */
 	@Override
 	public String makeValidTableName(String requestedName) {
+		Contracts.assertStringParameterNotEmpty( requestedName, "requestedName" );
 		return requestedName;
 	}
 
--- a/core/src/main/java/org/hibernate/ogm/dialect/spi/DialectNamingAdaptor.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/spi/DialectNamingAdaptor.java
@@ -1,0 +1,30 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.dialect.spi;
+
+/**
+ * The Dialect can validate or adjust the names which will be used to actually persist into the database "tables".
+ *
+ * @author Sanne Grinovero
+ */
+public interface DialectNamingAdaptor {
+
+	/**
+	 * If the proposed name is valid, the implementor should return it as-is.
+	 * If the dialect can apply a sensible and non-ambiguous transformation - such as escaping - it should apply it and
+	 * return the transformed string; optionally warning about it.
+	 * If the proposed table name is not acceptable and there is no safe escaping the dialect should throw an
+	 * org.hibernate.MappingException and provide a user friendly explanation, possibly suggesting to use the JPA Table
+	 * annotation to provide a safe alternative explicitly.
+	 *
+	 * @param requestedName the generated default, or as requested by the user mapping
+	 * @return a valid "table" name, potentially the same as requestedName
+	 * @throws org.hibernate.MappingException
+	 */
+	String makeValidTableName(String requestedName);
+
+}
--- a/core/src/main/java/org/hibernate/ogm/dialect/spi/DialectNamingAdaptor.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/spi/DialectNamingAdaptor.java
@@ -7,7 +7,12 @@
 package org.hibernate.ogm.dialect.spi;
 
 /**
- * The Dialect can validate or adjust the names which will be used to actually persist into the database "tables".
+ * The Dialect can validate or adjust the names of structures in the datastores, such as for
+ * example the name of the tables being used in the database.
+ *
+ * Dialect developers can optionally choose to provide a custom implementation; if they choose
+ * to, we recommend to extend {@link BaseDialectNamingAdaptor} to minimize maintenance needs
+ * as this contract will likely evolve.
  *
  * @author Sanne Grinovero
  */
--- a/core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
+++ b/core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
@@ -190,10 +190,12 @@ public abstract class OgmEntityPersister extends AbstractEntityPersister impleme
 			throw log.getIdentityGenerationStrategyNotSupportedException( getEntityName() );
 		}
 
-		tableName = persistentClass.getTable().getQualifiedName(
-				factory.getDialect(),
-				factory.getSettings().getDefaultCatalogName(),
-				factory.getSettings().getDefaultSchemaName()
+		tableName = gridDialect.getNamingAdaptor().makeValidTableName(
+				persistentClass.getTable().getQualifiedName(
+						factory.getDialect(),
+						factory.getSettings().getDefaultCatalogName(),
+						factory.getSettings().getDefaultSchemaName()
+				)
 		);
 
 		this.discriminator = discriminator;
--- a/core/src/test/java/org/hibernate/ogm/backendtck/innertypes/InnerClassFindTest.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/innertypes/InnerClassFindTest.java
@@ -1,0 +1,64 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.backendtck.innertypes;
+
+import static org.fest.assertions.Assertions.assertThat;
+import javax.persistence.EntityManager;
+import org.hibernate.ogm.backendtck.innertypes.CommunityMember.Employee;
+import org.hibernate.ogm.utils.TestForIssue;
+import org.hibernate.ogm.utils.jpa.JpaTestCase;
+import org.junit.Test;
+
+/**
+ * Verifies the dialect is able to store an entity which is represented as an inner type in Java.
+ *
+ * For example on MongoDB this implies the name of the class needs to avoid the dollar symbol,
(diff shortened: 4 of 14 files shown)


Sanne:
re-pushed to fix the commit description

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/logging/impl/Log.java
```diff
@@ -75,4 +76,17 @@
 
 	@Message(id = 1219, value = "Database %s does not exist. Either create it yourself or set property '" + OgmProperties.CREATE_DATABASE + "' to true.")
 	HibernateException databaseDoesNotExistException(String databaseName);
+
+	@Message(id = 1220, value = "When using MongoDB it is not valid to use a name for a table (a collection) which starts with the 'system.' prefix."
+			+ " Please change name for '%s', for example using @Table ")
```
hferentschik: for example *by* using 

same for the other messages

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/logging/impl/Log.java
```diff
@@ -75,4 +76,17 @@
...
+	MappingException collectionNameHasInvalidSystemPrefix(String qualifiedName);
+
+	@Message(id = 1221, value = "When using MongoDB it is not valid to use a name for a table (a collection) which contains the NUL character '\\0'."
+			+ " Please change name for '%s', for example using @Table ")
+	MappingException collectionNameContainsNULCharacter(String qualifiedName);
+
+	@Message(id = 1222, value = "When using MongoDB it is not valid to use a name for a table (a collection) which contains the dollar character '$'."
```
hferentschik: Maybe add a reference to inner classes, so that people understand where the '$' potentially comes from!?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/dialect/impl/MongoDBNamingAdaptor.java
```diff
@@ -0,0 +1,41 @@
...
+		}
+		else if ( requestedName.startsWith( "system." ) ) {
+			throw log.collectionNameHasInvalidSystemPrefix( requestedName );
+		}
+		else if ( requestedName.contains( "\u0000" ) ) {
+			throw log.collectionNameContainsNULCharacter( requestedName );
+		}
+		else if ( requestedName.contains( "$" ) ) {
+			throw log.collectionNameContainsDollarCharacter( requestedName );
+		}
```
hferentschik: not much adapting going on here ;-)

Sanne: Indeed, MongoDB doesn't have any reliable escaping capabilities. See the discussion on the JIRA as well ;-)

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/logging/impl/Log.java
Sanne: +1 good idea

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/logging/impl/Log.java
Sanne: thanks! changing

Sanne:
thanks @hferentschik , I've pushed an updated

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/dialect/impl/MongoDBNamingAdaptor.java
hferentschik: Sure. I am btw a bit surprised that the name is only needed/relevant in `OgmEntityPersister.java`.

core/src/main/java/org/hibernate/ogm/dialect/spi/BaseDialectNamingAdaptor.java
```diff
@@ -0,0 +1,21 @@
...
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.dialect.spi;
+
+public class BaseDialectNamingAdaptor implements DialectNamingAdaptor {
+
+	static final DialectNamingAdaptor DEFAULT_INSTANCE = new BaseDialectNamingAdaptor();
+
+	public BaseDialectNamingAdaptor() {
```
gunnarmorling: Should it be private?

core/src/main/java/org/hibernate/ogm/dialect/spi/DialectNamingAdaptor.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.dialect.spi;
+
+/**
+ * The Dialect can validate or adjust the names which will be used to actually persist into the database "tables".
```
gunnarmorling: I would put it a bit more generically: "... the names of structures in the datastores", leaving the door open for future amendment. Also can you add a note that one should not implement this interface directly but rather extend the base class?

core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
```diff
@@ -190,10 +190,12 @@
 			throw log.getIdentityGenerationStrategyNotSupportedException( getEntityName() );
 		}
 
-		tableName = persistentClass.getTable().getQualifiedName(
-				factory.getDialect(),
-				factory.getSettings().getDefaultCatalogName(),
-				factory.getSettings().getDefaultSchemaName()
+		tableName = gridDialect.getNamingAdaptor().makeValidTableName(
```
gunnarmorling: Some more business around table names is going on further below in that constructor. Should that contract be applied to those as well? I'm not sure, as this stuff might be "dead" and not used for OGM atm.

core/src/test/java/org/hibernate/ogm/backendtck/innertypes/InnerClassFindTest.java
```diff
@@ -0,0 +1,64 @@
...
+ * See also org.hibernate.ogm.datastore.mongodb.test.datastore.CollectionNamingValidationTest in the MongoDB module.
+ *
+ * @author Davide D'Alto
+ * @author Sanne Grinovero
+ */
+@TestForIssue(jiraKey = "OGM-265")
+public class InnerClassFindTest extends JpaTestCase {
+
+	@Test
+	public void testInnerClassFind() throws Exception {
```
gunnarmorling: Do we have a test for an adaptor which actually changes the name? That'd help to make sure the contract is used everwhere it should be.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/dialect/impl/MongoDBNamingAdaptor.java
```diff
@@ -0,0 +1,41 @@
...
+
+	private static final Log log = LoggerFactory.getLogger();
+
+	private MongoDBNamingAdaptor() {
+		//use the singleton INSTANCE
+	}
+
+	@Override
+	public String makeValidTableName(String requestedName) {
+		if ( StringHelper.isEmpty( requestedName ) ) {
```
gunnarmorling: How about adding a `assertParameterNotEmpty()` to our `Contracts` class?

core/src/main/java/org/hibernate/ogm/dialect/spi/BaseDialectNamingAdaptor.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.dialect.spi;
+
+public class BaseDialectNamingAdaptor implements DialectNamingAdaptor {
```
gunnarmorling: As this is an SPI element, could you add a short comment? Thanks!

core/src/main/java/org/hibernate/ogm/dialect/spi/BaseDialectNamingAdaptor.java
Sanne: Being the "base" one, I'd expect people to extend this one.
That will allow us to add more contracts down the road for improved validations without breaking the API.

core/src/test/java/org/hibernate/ogm/backendtck/innertypes/InnerClassFindTest.java
gunnarmorling: E.g. it would be interesting to see if it works if a table is renamed which is an association target.

Sanne: the issue was about giving a proper warning to people using MongoDB with inner classes.
I think we should add the additional ones according to per-case priority, compared to other tasks.. so I'd not spend too much time on the others at this point as we don't have a strong reason.

core/src/main/java/org/hibernate/ogm/dialect/spi/BaseDialectNamingAdaptor.java
gunnarmorling: Ah yes, sure. "protected" then maybe.

The public constructor together with the `DEFAULT_INSTANCE` constant made me stumble.

core/src/test/java/org/hibernate/ogm/backendtck/innertypes/InnerClassFindTest.java
gunnarmorling: > so I'd not spend too much time on the others

WDYM?

I'm not talking about adding further methods to the contract but making sure it works as expected actually. We cannot add an extension point not knowing whether it's going to work actually.

For the issue at hand, it could be done more simpler altogether, just add the check to `MongoDBSchemaInitializer` as we discussed. But if we go down the proposed road, it needs to be tested.

hferentschik: > Do we have a test for an adaptor which actually changes the name? That'd help to make sure the contract is used everwhere it should be.

+1 I am also curious to see that changing/adapting a name actually works.

Sanne: I'm not sure I understand what you want me to do.
I should add a new dialect which does some transformation on this, for the sake of testing all of the TCK test cases with an hypothetical feature?

Sanne: The contract of the service is straight-forward. No doubt that it works.

It's entirely possible that it should be invoked in other places as well, but that's an implementation detail: if some code doesn't use that *constant* which we're patching today, it should be fixed to use that constant. That's not about testing this contract is it?

gunnarmorling: It is not a hypothetical feature, the contract is there. So yes, it needs to be tested. It is exposed to dialect implementors which may want to make use of it.

I don't think running the entire TCK with that customization is needed. I'd do a test persisting/loading a single entity with a customized table name and another one where a customized table is at the end of an association. This should cover the most important cases.

If you think testing for that is not feasible let's go a step back and do the simpler fix as suggested. Note that I like the proposed approach (should e.g. also be useful for CouchDB), but I am against releasing untested SPIs to the wild.

core/src/main/java/org/hibernate/ogm/dialect/spi/BaseDialectNamingAdaptor.java
Sanne: +1 right

core/src/main/java/org/hibernate/ogm/dialect/spi/DialectNamingAdaptor.java
Sanne: +1 nice

core/src/test/java/org/hibernate/ogm/backendtck/innertypes/InnerClassFindTest.java
hferentschik: > The contract of the service is straight-forward. No doubt that it works.

How can you be sure? You are not testing the case where you actually change the name from the original. That's what the contract "promises", but atm it just throws exceptions in all cases. 

You cold for example, just append "Foo" to a otherwise valid name. IMO it still should work with the table name becoming <orignialTableName>Foo.

gunnarmorling: > That's not about testing this contract is it?

It's about making sure that the SPI does what it promises. This includes having tests which show that the SPI is invoked at the right places. At least the most-standard scenarios should be covered, to avoid a huge WT* moment for users wishing to provide an SPI implementation and finding out it it's not invoked in a common case.

Sanne: Ok I wouldn't like to release an untested SPI either, I'm just not sure of what you think is not tested here. The SPI itself is tested for sure.. how you use it that's a different thing ;-)

Is the concern that various blocks of code should be invoking it - but actually aren't - for the case of table *rename* ?
That's a good point, I see if I can do something to cover for that.

gunnarmorling: > Is the concern that various blocks of code should be invoking it - but actually aren't - for the case of table rename ?

Yes, exactly. As I said, "... help to make sure the contract is used everwhere it should be" ;)

> That's a good point, I see if I can do something to cover for that.

Cool, thanks!

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/dialect/impl/MongoDBNamingAdaptor.java
Sanne: done, thanks I didn't know about this `Contract` class

Sanne:
I've pushed a new commit with the suggested javadoc and validation improvements, as a separate commit to make it easier to see the diff for you all. I'll squash it all together later.

Sanne:
(still needing the new tests though)

gunnarmorling:
@Sanne, any news on this front? Do you still plan to add a commit for these tests?

Sanne:
@gunnarmorling yes I have a draft and it's failing :)
I'm currently giving higher priority to the parser reviews, then I'll resume this. Let me close the PR for now..

PR #498, opened 2015-03-04
OGM-171 Automate setting the FIND_BY_ID strategy on fulltext queries

Sanne:
https://hibernate.atlassian.net/browse/OGM-171

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/cfg/impl/HibernateSearchIntegration.java
+++ b/core/src/main/java/org/hibernate/ogm/cfg/impl/HibernateSearchIntegration.java
@@ -1,0 +1,84 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.cfg.impl;
+
+import java.util.Map;
+import java.util.Properties;
+
+import org.hibernate.cfg.Configuration;
+
+/**
+ * Hibernate Search is an optional dependency, so while we need refer to its constants
+ * to make sure its configured correctly for OGM by default, we don't want to create a
+ * hard dependency. This class should be used as a helper to centralise that.
+ *
+ * @author Sanne Grinovero
+ */
+public final class HibernateSearchIntegration {
+
+	private static final boolean searchIsAvailable = isSearchAvailable();
+	private static String MASSINDEXER_PROPERTY_KEY = null;
+	private static String MASSINDEXER_PROPERTY_VALUE = null;
+	private static String RETRIEVALSTRATEGY_PROPERTY_KEY = null;
+	private static String RETRIEVALSTRATEGY_PROPERTY_VALUE = null;
+
+	static {
+		if ( searchIsAvailable ) {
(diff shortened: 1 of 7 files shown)


gunnarmorling:
Ah, very nice. Have been waiting for that one for a long time :)

core/src/main/java/org/hibernate/ogm/cfg/impl/HibernateSearchIntegration.java
```diff
@@ -0,0 +1,84 @@
...
+ * Hibernate Search is an optional dependency, so while we need refer to its constants
+ * to make sure its configured correctly for OGM by default, we don't want to create a
+ * hard dependency. This class should be used as a helper to centralise that.
+ *
+ * @author Sanne Grinovero
+ */
+public final class HibernateSearchIntegration {
+
+	private static final boolean searchIsAvailable = isSearchAvailable();
+	private static String MASSINDEXER_PROPERTY_KEY = null;
```
gunnarmorling: Why is it that you put the keys and values as constants here?

Sanne: If you see the commit history, in the first commit I just fix the issue w/o refactoring.
The result is the same string constants spread all over the code.

You then have two alternatives:
 - either define the constants somewhere like in org.hibernate.ogm.cfg.impl.InternalProperties - as Strings
 - this solution

I've preferred this one as it's type safe to depend on the actual constant, so the project wouldn't compile in case the changes name in Hibernate Search.
But also, Hibernate Search is not a mandatory dependency: for example Neo4J needs to exclude it.. so there needs to be some special care on making this all trigger only if Search is actually around.

gunnarmorling: Yes, I like the centralization part.

Only I'm wondering why you are not working with the actual constants (`o.h.s.c.Environment.DATABASE_RETRIEVAL_METHOD` etc.) within this class directly but rather go through this indirection. All the usages are guarded by `if( isSearchAvailable )` so I don't see what you gain by that?

>  Hibernate Search is not a mandatory dependency: for example Neo4J needs to exclude it.

Of course. But in which way do these constants help? Also I don't think you need to avoid importing these classes. You'd only get a CNFE upon actually using them (which you nicely guarded already).

Sanne: I tried initially, but that failed miserably in the Neo4J tests.. the extra guard is needed.

gunnarmorling: How odd. It works for me just as well if I remove all those constants and use the respective values inline (leaving the searchIsAvailable guard in place of course).

Anyways, I'll merge as is, it's fine. Though I'm a bit curious why this did not work for you.

gunnarmorling:
Merged. Thanks, @Sanne!

core/src/main/java/org/hibernate/ogm/cfg/impl/HibernateSearchIntegration.java
Sanne: ah I misunderstood that, I had tried in a different way.
Happy to return on this and polish, but why would you not extract constants? they are repeated all over.

gunnarmorling: No need to change. I wouldn't have done it for the keys (they are constants already), but I don't really mind.

PR #499, opened 2015-03-04
Integration tests cleanup and some dependency upgrades

Sanne:
- https://hibernate.atlassian.net/browse/OGM-758
 - https://hibernate.atlassian.net/browse/OGM-759
 - https://hibernate.atlassian.net/browse/OGM-760


pom.xml
```diff
@@ -56,6 +56,11 @@
         <embeddedMongoDbLogDir>${project.build.directory}/embeddedMongoDb/logs</embeddedMongoDbLogDir>
         <!-- Not using default port to reduce chances of shutting down an external MongoDB instance -->
         <embeddedMongoDbPort>27018</embeddedMongoDbPort>
+
+        <!-- Since we require users to import a specific Hibernate Search module explicitly,
+         all integration tests need to know which version that shall be.
+         So the property needs to be defined in the global parent pom -->
+        <hibernate-search.module.slot>5.1</hibernate-search.module.slot>
```
gunnarmorling: Why is it that 5.1 is used now? Is there a specific need?

If we explicitly need 5.1, I think there should be a note in the reference guide regarding the set-up in WildFly, specifically that the HSEARCH 5.1 module ZIP needs to be put into the installation.

This used to be part of the OGM ZIP (just repackaging the original HSEARCH ZIP) for the sake of convinience, but I guess this was removed at some point when we used the WF stock HSEARCH. We also might re-establish that mechanism.

integrationtest/mongodb/src/test/resources/arquillian.xml
```diff
@@ -1,28 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
gunnarmorling: Is the arquillian.xml from "testcase" applied now?

gunnarmorling:
Looks good to me apart from those remarks.

integrationtest/mongodb/src/test/resources/arquillian.xml
Sanne: yes .. it always was .. with some luck :)
as I say on the JIRA, there was a duplicate resource on classpath, making it pseudo-random behaviour. Luckily most of these resource were dubplicate in content too, but not all of them.

pom.xml
Sanne: it was 5.0, and since the JIRA is about moving to 5.1 I changed it.
Also the versions were hardcoded in the tests, now they are properly injected from the build.

The aspect on module distribution doesn't change.

integrationtest/mongodb/src/test/resources/arquillian.xml
gunnarmorling: Cool. Nice to get rid of it.

pom.xml
gunnarmorling: Thanks, Sanne!

PR #501, opened 2015-03-05
OGM-465 - PREVIEW 2 - SPI for obtaining applied operations of current flush cycle

gunnarmorling:
Hi @emmanuelbernard, opening the next round for the error handler. What has changed compared to the first PR:

* applied operations are stored per transaction now (via TX wrapper `ErrorHandlerEnabledTransaction`)
* new method `ErrorHandler#onFailedOperation()` which allows to continue or abort upon a failed op
* extracted generic flush-cycle state manager which now is used for exposing the TX error handler where it's needed. But it could also be used for the batching stuff (so basically one generic `ThreadLocal` rather specific ones, which raises a clear error if used out of flushes)

This addresses now the use cases 1.1, 2.1 and 2.1.1 from [OGM-465](https://hibernate.atlassian.net/browse/OGM-465). IMO that makes for a good first step functional-wise, as it removes the major obstacle for people right now: knowing what has been executed in case of failures. I think All the retry/compensating etc. stuff could be done separately.

I have not done much polishing, so please ignore minor issues such as comments etc. 

Probably the best is to checkout `ErrorSpiTest` to see what's possible and how it is used.

Thanks!


core/src/main/java/org/hibernate/ogm/exception/spi/ErrorHandlingStrategy.java
```diff
@@ -0,0 +1,16 @@
...
+ */
+package org.hibernate.ogm.exception.spi;
+
+
+/**
+ * @author Gunnar Morling
+ *
+ */
+public enum ErrorHandlingStrategy {
+	ABORT, CONTINUE;
```
gunnarmorling: Another value could be `RETRY`, but I don't think it's needed in the first iteration. It would repeatedly invoke the failed op (and the error handler) until it either succeeds or the error handler returns ABORT/CONTINUE.

emmanuelbernard:
Hum this PR is really hard to follow in this scaffolding state :/ I'm reviewing classes that are gone a few commits later, Not sure how to address it.

emmanuelbernard:
Still not finished, my try later tonight or tomorrow

emmanuelbernard:
I'm kind of concerned about the amount of objects created each time (one operation object, one context object, the list of operations. This will defeat all the work we did to improve the calls to grid dialect and that was showing up in the profiler.
I don't have a significantly better idea unfortunately.

It feels like we could push this API as experimental anyways. It's worth chatting with sanne tomorrow maybe he will have some interesting compromise idea.

core/src/main/java/org/hibernate/ogm/exception/spi/ErrorHandlingStrategy.java
emmanuelbernard: Why would you return "call me back" instead of retrying yourself in a loop + sleep?

emmanuelbernard: I guess one benefit is that the retry # + sleep could be generally configurable.

emmanuelbernard:
I think I'm done with the review. The feedback came in a bit of incoherent order but so was your PR so that's fair treatment ;)

Lots of comments but generally:

1. the flush state collector (you call it manager) is nice even though it needs either name or/and doc polish
2. I'm concerned about the number of extra objects created: I proposed some ideas but that remains an open question
3. the batch treatment could probably be improved with a smarter `OperationsQueue` that could also collection partial success / failure.

Some proposals can wait a post beta release, even final. It depends on your energy.

emmanuelbernard:
By the way the tests fails at Neo4J for me.

```
Running org.hibernate.ogm.backendtck.exception.ErrorSpiTest
Tests run: 4, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.209 sec <<< FAILURE! - in org.hibernate.ogm.backendtck.exception.ErrorSpiTest
subsequentOperationsArePerformedForErrorHandlingStrategyContinue(org.hibernate.ogm.backendtck.exception.ErrorSpiTest)  Time elapsed: 0.032 sec  <<< ERROR!
org.hibernate.TransactionException: commit failed
        at org.neo4j.kernel.impl.transaction.TxManager.rollbackCommit(TxManager.java:629)
        at org.neo4j.kernel.impl.transaction.TxManager.commit(TxManager.java:390)
        at org.hibernate.ogm.transaction.impl.JTATransactionManagerTransaction.doCommit(JTATransactionManagerTransaction.java:119)
        at org.hibernate.engine.transaction.spi.AbstractTransactionImpl.commit(AbstractTransactionImpl.java:180)
        at org.hibernate.ogm.transaction.impl.ForwardingTransactionImplementor.commit(ForwardingTransactionImplementor.java:78)
        at org.hibernate.ogm.backendtck.exception.ErrorSpiTest.subsequentOperationsArePerformedForErrorHandlingStrategyContinue(ErrorSpiTest.java:321)

```

emmanuelbernard:
For the record, here is an idea that would  _potentially_ limit pressure on memory.

Create an `Operations` object that stores the necessary data representing applied operations in a (set of) array.

```
class Operations implements Iterable<GridDialectOperation> {
    Object[] state = new Object[100]; //todo make it a set of arrays to scale on big lists
    int index = 0;
    int numberOfOperations = 0;
    applyOperation(OperationType operationType, Object... operationState) {
        // todo use the Slf4J trick with overloaded methods (OT) (OT, O), (OT, O,O), (OT, O,O, O...)
        state[index++] = operationType;
        for(Object object : operationState) { state[index++] = object; }
        numberOfOperations++;
    }

    int getNumberOfOperations() { return numberOfOperations; }

    // todo  iterable / iterator implementation

    private GridDialectOperation getOperation(index startOfState) {
        OperationType operationType = (OperationType) state[startOfState++];
        if ( operationType == INSERT_OR_UPDATE_TUPLE) {
            return new InsertOrUpdateTupleImpl(state, startOfState);
        }
        // todo other operationType

}
```

We would only instantiate the applied operation objects when the user needs them.

Maybe it's worth changing `ErrorHandler.RollbackContext.getAppliedOperations()` to return such an iterable + size `Operations` object to accommodate that potential improvement?

core/src/main/java/org/hibernate/ogm/exception/spi/ErrorHandlingStrategy.java
gunnarmorling: Right, it'd be simpler to use. You wouldn't have to implement that in each handler. And we could provide more advanced strategies such as exponential back-off etc.

gunnarmorling:
> Hum this PR is really hard to follow in this scaffolding state :/ I'm reviewing classes that are gone a few commits later, Not sure how to address it.

Understood. Things evolved a bit over time. I tried to go back and alter earlier commits based on my latest findings, but at some point the rebasing just wasn't managable anymore.

gunnarmorling:
> By the way the tests fails at Neo4J for me.

Yes, for me, too. Need to disable at on this backend. It is caused by the fact that while we ignore the exception and happily continue, Neo4j does not forget. When trying to commit it notices that something went wrong previously. I think the `CONTINUE` strategy only should be used on non-transactional stores. Should add a note.

gunnarmorling:
> the flush state collector (you call it manager) is nice even though it needs either name or/and doc polish

Agreed. I wanted to get it out quickly and get your opinion. Polishing comes next :)

> I'm concerned about the number of extra objects created: I proposed some ideas but that remains an open question

Yes, same here. Not sure what to do with this. The only way I see is to pass applied ops not as object but via dedicated methods to the error handler: onAppliedInsert(EntityKey, Tuple) etc. Makes the contract much broader but this would avoid the object allocation on our side and also altogether if people just want to log in a file or so. If we expose the operation impl types on the SPI, people still could assemble objects from that if needed.

> the batch treatment could probably be improved with a smarter OperationsQueue that could also collection partial success / failure.

Yeah. Can we handle that separately? Batching is a sore point IMO altogether.

emmanuelbernard:
> Yes, same here. Not sure what to do with this. The only way I see is to pass applied ops not as object but via dedicated methods to the error handler: onAppliedInsert(EntityKey, Tuple) etc. Makes the contract much broader but this would avoid the object allocation on our side and also altogether if people just want to log in a file or so. If we expose the operation impl types on the SPI, people still could assemble objects from that if needed.

I did think about that but I don't think that solves the crux of the problem. It sure offer the option to not collect operations if the {{ErrorHandler}} does not need them. Let's go with what we have (or the alternative proposal with a dedicated `Operations` object) and we can revisit later.

> > the batch treatment could probably be improved with a smarter OperationsQueue that could also collection partial success / failure.

> Yeah. Can we handle that separately? Batching is a sore point IMO altogether.

yes, just collect the feedback in a jira.

gunnarmorling:
> Maybe it's worth changing ErrorHandler.RollbackContext.getAppliedOperations() to return such an iterable

I've change it to return `Iterable` now. Didn't add the size method, we can do it if needed.

PR #503, opened 2015-03-05
OGM-762 Composite IDs used in tests should have an equals and hashcode d...

Sanne:
...efined

- https://hibernate.atlassian.net/browse/OGM-762


gunnarmorling:
Good catch. Rebased and applied.

PR #504, opened 2015-03-06
OGM-718 Make sure RESOURCE_LOCAL persistence unit do not require a transaction manager

hferentschik:


core/src/main/java/org/hibernate/ogm/datastore/spi/DatastoreProvider.java
```diff
@@ -46,4 +47,17 @@
...
+	 *
+	 * When transaction emulation is used, transactions only demarcate a unit of work. The transaction emulation will
+	 * make sure that at commit time all required changes are flushed, but there are otherwise no true transaction,
+	 * in particular rollback, semantics.
+	 *
+	 * @return {@code true} if this dialect allows transaction emulation, {@code false} otherwise.
+	 *
+	 */
+	@Experimental("This contract might evolve into something which differentiates in more detail various transactional capabilities (see OGM-763)")
+	boolean allowsTransactionEmulation();
```
gunnarmorling: I guess it's ok for now. I still would have preferred to return an object here which would give us better odds for evolving the thing in a compatible manner. But `@Experimental` gives us a free ride :)

core/src/main/java/org/hibernate/ogm/transaction/impl/EmulatedLocalTransaction.java
```diff
@@ -0,0 +1,129 @@
...
+		// nothing to do here
+	}
+
+	@Override
+	protected void doRollback() throws TransactionException {
+	}
+
+	@Override
+	public boolean isInitiator() {
+		return isActive();
```
gunnarmorling: Is it right to delegate to isActive() here? Might not make a difference in practice, but it seems confusing. Maybe always returning true would be better?

core/src/test/java/org/hibernate/ogm/backendtck/jpa/JPAAndJTAViaContainerAPITest.java
```diff
@@ -1,48 +0,0 @@
...
-import org.junit.Test;
-import org.hibernate.ogm.utils.jpa.JpaTestCase;
-
-import static org.fest.assertions.Assertions.assertThat;
-
-/**
- * Test that PersistenceProvider#createContainerEntityManagerFactory work properly in a JTA environment
- * @author Emmanuel Bernard &lt;emmanuel@hibernate.org&gt;
- */
-public class JPAAndJTAViaContainerAPITest extends JpaTestCase {
```
gunnarmorling: This one is not needed any more at all? It seems not to do what the description said anyways?

core/src/main/java/org/hibernate/ogm/datastore/spi/DatastoreProvider.java
```diff
@@ -46,4 +47,17 @@
 	 * @return the schema definer type
 	 */
 	Class<? extends SchemaDefiner> getSchemaDefinerType();
+
+	/**
+	 * Whether this dialect allows emulation of transactions.
```
gunnarmorling: s/dialect/datastore (provider)/

core/src/test/java/org/hibernate/ogm/utils/test/SkipByGridDialectSelfJpaTest.java
```diff
@@ -6,7 +6,7 @@
  */
 package org.hibernate.ogm.utils.test;
 
-import junit.framework.Assert;
+import org.junit.Assert;
```
gunnarmorling: Good catch!

gunnarmorling:
@hferentschik, looks good to me. Only one minor question and typo left (see inline). Otherwise it's good to merge.

core/src/main/java/org/hibernate/ogm/datastore/spi/DatastoreProvider.java
hferentschik: Right. I think our `TransactionCapabilties` idea might be the way forward in the end. But this way we have something we can use until we know exactly what we need.

core/src/main/java/org/hibernate/ogm/transaction/impl/EmulatedLocalTransaction.java
hferentschik: this is from the JDBC transaction code. I guess always returning true might work as well. I have to strong feeling for either solution.

core/src/test/java/org/hibernate/ogm/backendtck/jpa/JPAAndJTAViaContainerAPITest.java
hferentschik: Right, one test became obsolete in the end. It was pretty much identical to another of the tests based in this directory. Also, as you say the name was not correct either way.

hferentschik:
Fixed type, changed impl of `EmulatedLocalTransaction#isInitiator` and rebased.

gunnarmorling:
Merged. Many thanks, @hferentschik!

hferentschik:
My pleasure :-)

PR #506, opened 2015-03-09
OGM-412 Align the use of house keeping method calls between ORMs EntityM...

hferentschik:
...anagerImpl/SessionImpl with the respective OgmEntityManager/OgmSessionImpl

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/jpa/impl/OgmEntityManagerFactory.java
+++ b/core/src/main/java/org/hibernate/ogm/jpa/impl/OgmEntityManagerFactory.java
@@ -21,12 +21,14 @@ import javax.persistence.metamodel.Metamodel;
 import org.hibernate.SessionFactory;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.jpa.HibernateEntityManagerFactory;
+import org.hibernate.jpa.internal.EntityManagerImpl;
 import org.hibernate.jpa.internal.metamodel.EntityTypeImpl;
 import org.hibernate.ogm.hibernatecore.impl.OgmSessionFactoryImpl;
 
 /**
- * Delegate most work to the underlying EntityManagerFactory.
- * REturn an OgmEntityManager to cope with query operations
+ * An OGM {@code EntityManagerFactory} which delegates most work to the underlying ORM {@code EntityManagerFactory}.
+ *
+ * Note: Returns an {@code OgmEntityManager} to cope with query operations.
  *
  * @author Emmanuel Bernard &lt;emmanuel@hibernate.org&gt;
  */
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/hibernate/ogm/jpa/impl/OgmEntityManagerFactory.java
```diff
@@ -39,22 +41,26 @@ public OgmEntityManagerFactory(EntityManagerFactory hibernateEmf) {
 
 	@Override
 	public EntityManager createEntityManager() {
-		return new OgmEntityManager( this, hibernateEmf.createEntityManager() );
+		return new OgmEntityManager( this, hibernateEmf.createEntityManager().unwrap( EntityManagerImpl.class ) );
```
Sanne: There is no handling for the specific implementation _EntityManagerImpl.class_ in the unwrap() implementation of our EntityManager.
This happens to work as unwrap() is able to cast it to _EntityManager_ (the interface), but this only works if the actual implementation happens to be an _EntityManagerImpl_. So.. shouldn't you rather cast explicitly?

Sanne:
I'd like to merge something like this, but it feels like a temporary solution as we can't really guarantee that these method calls stay in synch, and it's hard to test.
For the long term should we also pursue having a nice reusable decorator in ORM codebase? that would also:
 - allow us to stay away from the internal implementation classes
 - make ORM's code more readable in this regard

The one problem I see, is that if we improve on this in ORM's code, we'll need to delegate the unwrap() call too, and the trick you do in the above comment would break.

Sanne:
BTW I don't see my comment above as a blocker: just wondering how to improve in next.gen ORM.

If you think the _unwrap_ catch is a good compromise for the time being, I'll merge it.

core/src/main/java/org/hibernate/ogm/jpa/impl/OgmEntityManagerFactory.java
hferentschik: Unwrapping is what one should do, but I agree the unwrap method is fishy. Unwrapping should check against `EntityManagetImpl` explicitly. In fact it should probably only be possible to unwrap to `org.hibernate.jpa.spi.AbstractEntityManagerImpl

hferentschik:
> I'd like to merge something like this,

do it then ;-)

> but it feels like a temporary solution as we can't really guarantee that these method calls stay in synch, and it's hard to test.

Sure, there is no guarantee that it won't stay in sync, but short of probably quite extensive surgery in the ORM/OGM code base I am not sure how you would solve that. I am not even sure how a "delegate" in ORM would help. 

> If you think the unwrap catch is a good compromise for the time being,

Not sure what you mean, but I actually removed the unwrap and use casts now.

hferentschik:
Pushed an updated version using casts instead of `unwrap`.

Sanne:
thanks, merged!

You think we should enhance the unwrap implementation to "extract" implementation types?

hferentschik:
> You think we should enhance the unwrap implementation to "extract" implementation types?

Yes, but that's a ORM issue right?

Sanne:
Sure, just wondering if you think we should follow up on that. I'm not sure about the intentions of the _unwrap_ method. Personally I think I like it that you can only extract an interface, but you seem to have a different opinion? Open an ORM issue if you can make a good argument for it, since 5 is coming soon..

hferentschik:
> Open an ORM issue if you can make a good argument for it, since 5 is coming soon..

FYI - HHH-9665

PR #509, opened 2015-03-11
OGM-465 "ErrorHandler SPI with a vengeance" - Director's Cut

gunnarmorling:
@emmanuelbernard Hopefully the last and final PR for OGM-465.

~~I need to push one more commit (or separate PR) for the doc update, but apart from that it's done from my side.~~ Also contains the doc update now.

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/cfg/OgmProperties.java
+++ b/core/src/main/java/org/hibernate/ogm/cfg/OgmProperties.java
@@ -6,6 +6,8 @@
  */
 package org.hibernate.ogm.cfg;
 
+import org.hibernate.ogm.failure.ErrorHandler;
+
 /**
  * Common properties for configuring NoSql datastores via {@code persistence.xml} or
  * {@link OgmConfiguration}.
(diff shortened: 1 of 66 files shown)


core/src/main/java/org/hibernate/ogm/exception/spi/ErrorHandler.java
```diff
@@ -0,0 +1,76 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.exception.spi;
```
gunnarmorling: @emmanuelbernard, given the definition of "SPI" we use, should this actually not rather be "API"? It's expected to be implemented and used by OGM users, not dialect authors or integrators. I think we should drop the "spi" from the package names.

emmanuelbernard:
OK I'll look, likely starting at 17:00 PM

emmanuelbernard:
But I am not necessarily happy about the package name:

- `failure`
- `compensation`

+1 for dropping `spi`

gunnarmorling:
Renamed to "failure" and force-pushed.

emmanuelbernard:
I could not give it my full attention but it looks good overall. I've put specific comments. I think after thinking about it we will rethink the doc but that can wait.

gunnarmorling:
@emmanuelbernard Reworked some commits to add the TODO around batching, enable that one test and apply some changes to the docs as per your comments.

gunnarmorling:
Rebased and applied by @emmanuelbernard, closing.

PR #510, opened 2015-03-11
OGM-265 Validate collection names used for tables in MongoDB

Sanne:
We've thrown the previous proposal for a generic validation contract away, this is the simplified form and is MongoDB specific.

 - https://hibernate.atlassian.net/browse/OGM-265

The better naming validation should be performed in Hibernate ORM, for example to validate to not use database specific keywords.


gunnarmorling:
Rebased and applied. Thanks, @Sanne!

PR #511, opened 2015-03-11
OGM-749 Creating a ruby script for creating hibernate.org release file a...

hferentschik:
...nd updating changelog.txt

Some notes:
* This is based on the ruby snipplet which is already used in the Jenkins release just to create the release file for hibernate.org. Since updating the `changelog.txt` also requires accessing Jira I combined everything into a single script. 
* I placed everything into a folder of the actual source tree (`src/main/release-scripts`). IMO this has the benefits of having the release "knowledge" under version control as well and it allows for easier testing locally. See also OGM-769. From the Jenkins side one just executes the script locally to the workspace (instead of typing all the commands into the Jenkins shell command text-field directly)

Thoughts?

BTW, I created a Jenkins test job 'hibernate-ogm-release-test' to just test the execution of this script and it works fine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,9 +27,6 @@ bin/
 #build
 target/
 
-#custom scripts
-*.sh
-
 #test output
 test-output/
 transaction.log
--- a/src/main/release-scripts/Gemfile
+++ b/src/main/release-scripts/Gemfile
@@ -1,0 +1,4 @@
+source 'http://rubygems.org'
+
+gem 'choice'
+gem 'git'
--- a/src/main/release-scripts/Gemfile.lock
+++ b/src/main/release-scripts/Gemfile.lock
@@ -1,0 +1,13 @@
+GEM
+  remote: http://rubygems.org/
+  specs:
+    choice (0.2.0)
+    git (1.2.9.1)
+
+PLATFORMS
+  java
+  ruby
+
+DEPENDENCIES
+  choice
+  git
--- a/src/main/release-scripts/create_tag.sh
+++ b/src/main/release-scripts/create_tag.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+RELEASE_VERSION=$1
+
+git commit -a -m "[Jenkins release job] Preparing release $RELEASE_VERSION"
+git tag $RELEASE_VERSION
+
--- a/src/main/release-scripts/pre-release.rb
+++ b/src/main/release-scripts/pre-release.rb
@@ -1,0 +1,227 @@
+#!/usr/bin/env ruby
+# encoding: UTF-8
+
+########################################################################################################################
+# The purpose of this tool is to handle several pre release tasks. These tasks are:
+# - Update the data in readme.md
+# - Update the changelog.txt (using JIRA's REST API to get the required information)
+# - Print a release file template for the hibernate.org website
+########################################################################################################################
+
+require 'rubygems'
+require 'bundler/setup'
+
+require 'json'
+require 'net/https'
+require 'choice'
+require 'git'
+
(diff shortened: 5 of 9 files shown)


src/main/release-scripts/changelog-and-release-file-handling.rb
```diff
@@ -0,0 +1,172 @@
...
+  option :website_release, :required => false do
+    short '-w'
+    long '--website-release-file'
+    desc 'If specified the hibernate.org release file for the specified release will be created'
+  end
+
+  option :update_changelog, :required => false do
+    short '-c'
+    long '--update-change-log=<path to changelog>'
+    desc 'If specified the hibernate.org release file for the specified release will be created'
```
gunnarmorling: release file = changelog file

src/main/release-scripts/changelog-and-release-file-handling.rb
```diff
@@ -0,0 +1,172 @@
...
+  end
+
+  separator 'Common:'
+
+  option :help do
```
gunnarmorling: Could you display a sentence describing the tool's purpose?

src/main/release-scripts/changelog-and-release-file-handling.rb
```diff
@@ -0,0 +1,172 @@
...
+  puts 'date: ' + jira_release_info['releaseDate']
+  puts 'stable: ' + (version.include?('Final') ? 'true' : 'false')
+  puts 'announcement_url: <TBD>'
+  puts 'summary: ' + (jira_release_info['description'] or '<TBD>')
+  puts 'displayed: true'
+  puts 'EOF'
+  puts '====='
+end
+
+def create_changelog_update(release_version)
```
gunnarmorling: A short description would be nice, specifically of the format of change log records.

gunnarmorling:
> Since updating the changelog.txt also requires accessing Jira I combined everything into a single script. 

+1

> I placed everything into a folder of the actual source tree (src/main/release-scripts)

Yes, that's very good. As you say in OGM-769, everything should go in there eventually.

gunnarmorling:
In addition to updating the changelog, could you also do the same for date and version in the readme file?

gunnarmorling:
@hferentschik, nice stuff! With this in place, we can start and use a component in JIRA for each backend we have.

src/main/release-scripts/changelog-and-release-file-handling.rb
hferentschik: Ok

src/main/release-scripts/changelog-and-release-file-handling.rb
hferentschik: Good idea

src/main/release-scripts/changelog-and-release-file-handling.rb
hferentschik: +1

hferentschik:
Sure.The good thing is that in the end I can just move this to Valitator as well :-)

hferentschik:
I applied the suggested fixes to the ruby script and added also a commit for OGM-769 which adds the various shell snipplets to version control.

Left to do would be now to update the release build on the CI server. Not sure whether someone feels for doing it for today's release already or whether we do this later!?

P.S. I left the different echo steps out from the scripts. I think I would keep this type of echoing in the Jenkins job.

gunnarmorling:
Merged, added one commit for fixing a typo in one of the file names.

>  added also a commit for OGM-769 which adds the various shell snipplets to version control.

That's great.
> Not sure whether someone feels for doing it for today's release already or whether we do this later!?

I'd say let's do this separately, but it's @DavideD 's call as he will be doing the release.

hferentschik:
Thanks :-)

gunnarmorling:
Thank you, @hferentschik!

PR #512, opened 2015-03-12
[4.2] OGM-778 Running TCK tests from test JAR itself rather than unpacking

gunnarmorling:
Turns out, it also scans the classified JAR when specifying the dependency name.

### DIFF
--- a/couchdb/pom.xml
+++ b/couchdb/pom.xml
@@ -121,10 +121,10 @@
                 <configuration>
                     <forkMode>once</forkMode>
                     <skipExec>${skipUnitTests}</skipExec>
+                    <dependenciesToScan>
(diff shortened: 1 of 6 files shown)


Sanne:
I'm very glad it works but I'm not understanding how :-)

Davide can testify that I had an identical branch as I've shown it to him last week, but then I have thrown it away as when adding _dependenciesToScan_ and pointing to _org.hibernate.ogm:hibernate-ogm-core_ it would - as expected - not take the test jar.

Sanne:
I ported your commit to branch _master_, it doesn't work there. What's different?
(See my branch _OGM-778-master_)

I'd love we get this to work but I'd rather not rely on some bug, as it's not supposed to work..

Sanne:
ok. Gunnar pointed out in chat that the difference is master didn't receive the surefire upgrade. merging...

PR #515, opened 2015-03-27
Don't register initiator for ORM's configuration service

gunnarmorling:


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/515

PR #518, opened 2015-04-13
OGM-122 cassandra backend

jhalliday:
Phase one of the cassandra backend - basic connection, schema, types. Should handle single entities, but no association support yet.

### DIFF
--- a/cassandra/pom.xml
+++ b/cassandra/pom.xml
@@ -20,8 +20,8 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <environmentVariables>
-                        <cassandra_HOSTNAME>${env.cassandra_HOSTNAME}</cassandra_HOSTNAME>
-                        <cassandra_PORT>${env.cassandra_PORT}</cassandra_PORT>
+                        <CASSANDRA_HOSTNAME>${env.CASSANDRA_HOSTNAME}</CASSANDRA_HOSTNAME>
+                        <CASSANDRA_PORT>${env.CASSANDRA_PORT}</CASSANDRA_PORT>
                     </environmentVariables>
                     <forkMode>once</forkMode>
 
--- a/cassandra/pom.xml
+++ b/cassandra/pom.xml
@@ -1,0 +1,193 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>hibernate-ogm-parent</artifactId>
+        <groupId>org.hibernate.ogm</groupId>
+        <version>4.2.0-SNAPSHOT</version>
+        <relativePath>..</relativePath>
+    </parent>
+
+    <artifactId>hibernate-ogm-cassandra</artifactId>
+
+    <name>Hibernate OGM for Cassandra</name>
+    <description>Persist objects in Cassandra</description>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <environmentVariables>
+                        <cassandra_HOSTNAME>${env.cassandra_HOSTNAME}</cassandra_HOSTNAME>
+                        <cassandra_PORT>${env.cassandra_PORT}</cassandra_PORT>
+                    </environmentVariables>
+                    <forkMode>once</forkMode>
+
+                    <includes>
+                        <include>**/BuiltInTypeTest*</include>
+                    </includes>
+
+                    <dependenciesToScan>
+                        <dependency>org.hibernate.ogm:hibernate-ogm-core</dependency>
+                    </dependenciesToScan>
+
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.bsc.maven</groupId>
+                <artifactId>maven-processor-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <artifactId>maven-dependency-plugin</artifactId>
+            </plugin>
+
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>animal-sniffer-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>check-java-api-signature</id>
+                        <phase>verify</phase>
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>${project.groupId}</groupId>
+                <artifactId>hibernate-ogm-bom</artifactId>
+                <version>${project.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.hibernate.ogm</groupId>
+            <artifactId>hibernate-ogm-core</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.datastax.cassandra</groupId>
+            <artifactId>cassandra-driver-core</artifactId>
+            <version>2.1.3</version>
+        </dependency>
+
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate.ogm</groupId>
+            <artifactId>hibernate-ogm-core</artifactId>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-search-orm</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.jbossts</groupId>
+            <artifactId>jbossjta</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.easytesting</groupId>
+            <artifactId>fest-assert</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.shrinkwrap</groupId>
+            <artifactId>shrinkwrap-impl-base</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.hibernate.hql</groupId>
+            <artifactId>hibernate-hql-lucene</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-search-orm</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-search-engine</artifactId>
+        </dependency>
+
+    </dependencies>
+
+    <profiles>
+        <profile>
+            <id>embedded-cassandra</id>
+            <activation>
+                <property>
+                    <name>!useExternalcassandra</name>
+                </property>
+            </activation>
+            <build>
+                <plugins>
+
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <environmentVariables>
+                                <cassandra_PORT>${embeddedcassandraPort}</cassandra_PORT>
+                                <cassandra_HOSTNAME>${embeddedcassandraBindIp}</cassandra_HOSTNAME>
+                            </environmentVariables>
+                            <forkMode>once</forkMode>
+                        </configuration>
+                    </plugin>
+
+                    <!-- http://mojo.codehaus.org/cassandra-maven-plugin/ -->
+                    <plugin>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>cassandra-maven-plugin</artifactId>
+                        <version>2.0.0-1</version>
+                        <configuration>
+                            <startNativeTransport>true</startNativeTransport>
+                        </configuration>
+                        <executions>
+                            <!--<execution>-->
+                                <!--<goals>-->
+                                    <!--<goal>start</goal>-->
+                                    <!--<goal>stop</goal>-->
+                                <!--</goals>-->
+                            <!--</execution>-->
+                        </executions>
+                        <dependencies>
+                            <!-- use to override the version of cassandra used by the plugin. -->
+                            <!-- but due to class changes in cassandra you can't currently go to 2.1:
+                            https://jira.codehaus.org/browse/MCASSANDRA-38 -->
+                            <dependency>
+                                <groupId>org.apache.cassandra</groupId>
+                                <artifactId>cassandra-all</artifactId>
+                                <version>2.1.4</version>
+                            </dependency>
+                        </dependencies>
+                    </plugin>
+
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+</project>
--- a/cassandra/pom.xml
+++ b/cassandra/pom.xml
@@ -133,6 +133,12 @@
             <artifactId>hibernate-search-engine</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>io.takari.junit</groupId>
+            <artifactId>takari-cpsuite</artifactId>
+            <scope>test</scope>
+        </dependency>
+
     </dependencies>
 
     <profiles>
--- a/cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
+++ b/cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
@@ -1,0 +1,329 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.datastore.cassandra;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+import com.datastax.driver.core.BoundStatement;
+import com.datastax.driver.core.DataType;
+import com.datastax.driver.core.PreparedStatement;
+import com.datastax.driver.core.ProtocolVersion;
+import com.datastax.driver.core.RegularStatement;
+import com.datastax.driver.core.ResultSet;
+import com.datastax.driver.core.Row;
+import com.datastax.driver.core.Session;
+import com.datastax.driver.core.exceptions.DriverException;
+import com.datastax.driver.core.querybuilder.Delete;
+import com.datastax.driver.core.querybuilder.Insert;
+import com.datastax.driver.core.querybuilder.QueryBuilder;
+import com.datastax.driver.core.querybuilder.Select;
+
+import org.hibernate.HibernateException;
+import org.hibernate.LockMode;
+import org.hibernate.dialect.lock.LockingStrategy;
+import org.hibernate.ogm.datastore.cassandra.impl.CassandraDatastoreProvider;
+import org.hibernate.ogm.datastore.cassandra.impl.CassandraTypeMapper;
+import org.hibernate.ogm.datastore.cassandra.logging.impl.Log;
+import org.hibernate.ogm.datastore.cassandra.logging.impl.LoggerFactory;
+import org.hibernate.ogm.datastore.cassandra.model.impl.ResultSetTupleSnapshot;
+
+import org.hibernate.ogm.datastore.map.impl.MapTupleSnapshot;
+
+import org.hibernate.ogm.dialect.batch.spi.BatchableGridDialect;
+import org.hibernate.ogm.dialect.batch.spi.InsertOrUpdateAssociationOperation;
+import org.hibernate.ogm.dialect.batch.spi.InsertOrUpdateTupleOperation;
+import org.hibernate.ogm.dialect.batch.spi.Operation;
+import org.hibernate.ogm.dialect.batch.spi.OperationsQueue;
+import org.hibernate.ogm.dialect.batch.spi.RemoveAssociationOperation;
+import org.hibernate.ogm.dialect.batch.spi.RemoveTupleOperation;
+
+import org.hibernate.ogm.dialect.spi.AssociationContext;
+import org.hibernate.ogm.dialect.spi.AssociationTypeContext;
+import org.hibernate.ogm.dialect.spi.DuplicateInsertPreventionStrategy;
+import org.hibernate.ogm.dialect.spi.ModelConsumer;
+import org.hibernate.ogm.dialect.spi.NextValueRequest;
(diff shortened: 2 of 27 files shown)


gunnarmorling:
@jhalliday Going to look into this now.

Could you also provide a commit/PR with a new chapter on the Cassandra backend for the reference guide? Especially the description of mappings (e.g. analogous to the [MongoDB chapter](http://docs.jboss.org/hibernate/ogm/4.2/reference/en-US/html_single/#ogm-mongodb-storage-principles) will be helpful.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraDatastoreProvider.java
```diff
@@ -0,0 +1,266 @@
...
+
+	@Override
+	public void start() {
+		if ( cluster == null ) {
+			try {
+				log.connectingToCassandra( config.getHost(), config.getPort() );
+
+				cluster = new Cluster.Builder()
+						.addContactPoint( config.getHost() )
+						.withPort( config.getPort() )
```
gunnarmorling: support for credentials would be great, too.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraDatastoreProvider.java
```diff
@@ -0,0 +1,266 @@
...
+		sequenceHandler = null;
+	}
+
+	public void removeKeyspace() {
+		Session bootstrapSession = cluster.connect();
+		bootstrapSession.execute( "DROP KEYSPACE " + config.getDatabaseName() );
+		bootstrapSession.close();
+	}
+
+	// test harness
```
gunnarmorling: It would be cool if this stuff could be moved to `CassandraTestHelper`. That'd also avoid the need for `metaDataCache` here.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraSequenceHandler.java
```diff
@@ -0,0 +1,155 @@
...
+
+	private final CassandraDatastoreProvider provider;
+
+	public CassandraSequenceHandler(CassandraDatastoreProvider provider) {
+		this.provider = provider;
+	}
+
+	public void createSequence(String generatorsKey, CassandraDatastoreProvider datastoreProvider) {
+
+		// ogm wiring doesn't allow us to get at the real col names :-(
```
gunnarmorling: What exactly is missing? Have you checked what's available on `IdSourceKeyMetadata`?

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraSequenceHandler.java
```diff
@@ -0,0 +1,155 @@
...
+
+		PreparedStatement preparedStatement = provider.getSession().prepare( select.toString() );
+		BoundStatement boundStatement = preparedStatement.bind( sequenceName );
+
+		ResultSet resultSet;
+		try {
+			resultSet = provider.getSession().execute( boundStatement );
+		}
+		catch (DriverException e) {
+			System.out.println( e.toString() );
```
gunnarmorling: I guess that can go away :)

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraSequenceHandler.java
```diff
@@ -0,0 +1,155 @@
...
+		columnTypes.add( "bigint" );
+		datastoreProvider.createColumnFamilyIfNeeded( generatorsKey, primaryKeyName, columnNames, columnTypes );
+	}
+
+	private Long nextValueSelect(String tableName, String sequenceName) {
+
+		Statement select = select().column( "sequence_value" ).from( "\"" + tableName + "\"" )
+				.where( eq( "sequence_name", QueryBuilder.bindMarker() ) );
+
+		PreparedStatement preparedStatement = provider.getSession().prepare( select.toString() );
```
gunnarmorling: Would it make sense to keep the prepared statements around for re-use?

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraSequenceHandler.java
```diff
@@ -0,0 +1,155 @@
...
+		}
+		catch (DriverException e) {
+			System.out.println( e.toString() );
+			throw e;
+		}
+
+		return resultSet.one().getBool( 0 );
+	}
+
+	public Number nextValue(NextValueRequest request) {
```
gunnarmorling: Did you look into [Cassandra counters](http://www.datastax.com/dev/blog/whats-new-in-cassandra-2-1-a-better-implementation-of-counters) by any chance? Seems like we could use those to represent sequences?

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/model/impl/ResultSetTupleSnapshot.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Wrap a java-driver Row, handling column name and type mappings.
+ *
+ * @author Jonathan Halliday
+ */
+public class ResultSetTupleSnapshot implements TupleSnapshot {
+
+	private final Row row;
+	private Map<String, Integer> columnNames = new HashMap<String, Integer>();
+	private ProtocolVersion protocolVersion;
```
gunnarmorling: Is this needed as a member here? Or could we obtain it from some shared constant?

cassandra/pom.xml
```diff
@@ -0,0 +1,193 @@
...
+    <description>Persist objects in Cassandra</description>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <environmentVariables>
+                        <cassandra_HOSTNAME>${env.cassandra_HOSTNAME}</cassandra_HOSTNAME>
```
gunnarmorling: Setting this to another value than localhost seems to take no effect (I am running Cassandra in a VM with another IP). Probably this needs to be pushed into a system variable similar to e.g. done in `MongoDBTestHelper`. Could you also make it all upper-case to be consistent with the env variables for the other backends we have? Thx!

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/type/impl/CassandraCalendarDateType.java
```diff
@@ -0,0 +1,39 @@
...
+import org.hibernate.MappingException;
+import org.hibernate.engine.spi.Mapping;
+import org.hibernate.ogm.type.impl.AbstractGenericBasicType;
+import org.hibernate.type.descriptor.java.CalendarDateTypeDescriptor;
+
+import java.util.Calendar;
+import java.util.Date;
+
+/**
+ * Cassandra doesn't do Calendar, but it does do Date, so convert them.
```
gunnarmorling: What happens to the TZ information of calendars? As per [this doc](http://docs.datastax.com/en/cql/3.0/cql/cql_reference/timestamp_type_r.html), Cassandra's time stamps support storing a TZ offset. I guess that'd require to put forward calendars as formatted strings rather than `Date`.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
```diff
@@ -0,0 +1,329 @@
...
+import static com.datastax.driver.core.querybuilder.QueryBuilder.insertInto;
+import static com.datastax.driver.core.querybuilder.QueryBuilder.select;
+import static com.datastax.driver.core.querybuilder.QueryBuilder.quote;
+
+/**
+ * Dialect implementation using CQL3 over Cassandra's native transport via java-driver.
+ *
+ * @author Jonathan Halliday
+ */
+public class CassandraDialect implements BatchableGridDialect {
```
gunnarmorling: This can extend `BaseGridDialect`, providing default/no-op implementations for a few methods.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
```diff
@@ -0,0 +1,329 @@
...
+
+	@Override
+	public GridType overrideType(Type type) {
+		return CassandraTypeMapper.INSTANCE.overrideType( type );
+	}
+
+	@Override
+	public void executeBatch(OperationsQueue queue) {
+
+		// placeholder until we get around to using logged batch for these.
```
gunnarmorling: You can skip the method altogether by implementing `GridDialect` rather than `BatchableGridDialect`.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
```diff
@@ -0,0 +1,329 @@
...
+					deleteOps.add( op );
+					break;
+				default:
+					throw new HibernateException( "TupleOperation not supported: " + op.getType() );
+			}
+		}
+
+		if ( deleteOps.size() > 0 ) {
+
+			Delete.Selection deleteSelection = delete();
```
gunnarmorling: REMOVE / PUT_NULL refer to a single column only. So this should rather yield an UPDATE than a DELETE statement.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
```diff
@@ -0,0 +1,329 @@
...
+			for ( int i = 1; i < key.getColumnNames().length; i++ ) {
+				deleteWhere = deleteWhere.and( eq( quote( key.getColumnNames()[i] ), QueryBuilder.bindMarker() ) );
+			}
+
+			bindAndExecute( key.getColumnValues(), delete );
+		}
+
+		if ( updateOps.size() > 0 ) {
+
+			Insert insert = insertInto( quote( key.getTable() ) );
```
gunnarmorling: Ah, so INSERT in Cassandra has UPSERT semantics, right? Probably worth adding a short note.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
```diff
@@ -0,0 +1,329 @@
...
+			PreparedStatement preparedStatement = session.prepare( statement );
+			BoundStatement boundStatement = new BoundStatement( preparedStatement );
+			for ( int i = 0; i < columnValues.length; i++ ) {
+				DataType dataType = preparedStatement.getVariables().getType( i );
+				boundStatement.setBytesUnsafe( i, dataType.serialize( columnValues[i], protocolVersion ) );
+			}
+			return session.execute( boundStatement );
+		}
+		catch (DriverException e) {
+			log.failToExecuteCQL( statement.getQueryString(), e );
```
gunnarmorling: I'd rather re-throw that exception instead swallowing it. An NPE will be raised subsequently.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraDatastoreProvider.java
```diff
@@ -0,0 +1,266 @@
...
+		implements Startable, Stoppable, ServiceRegistryAwareService, Configurable {
+
+	private static final Log log = LoggerFactory.getLogger();
+
+	private ServiceRegistryImplementor serviceRegistry;
+
+	private CassandraConfiguration config;
+
+	private Cluster cluster;
+	private Session session;
```
gunnarmorling: I tried to stop and re-start my Cassandra server while the test is running. It seems we need some kind of re-connect mechanism, I continued to get exceptions from the driver also if the server was up again. Probably a new session needs to be fired up in this case.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraDatastoreProvider.java
```diff
@@ -0,0 +1,266 @@
...
+	private CassandraSequenceHandler sequenceHandler;
+
+	private final Set<String> entityTableNames = new HashSet<String>();
```
gunnarmorling: This never seems to be read. Is it some kind of left-over?

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
```diff
@@ -0,0 +1,329 @@
...
+	public LockingStrategy getLockingStrategy(Lockable lockable, LockMode lockMode) {
+		//Cassandra essentially has no workable lock strategy unless you use external tools like
+		// ZooKeeper or fancy cql overlays like waitchain.
+		return null;
+	}
+
+	private ResultSet bindAndExecute(Object[] columnValues, RegularStatement statement) {
+
+		try {
+			PreparedStatement preparedStatement = session.prepare( statement );
```
gunnarmorling: Should the PS be cached? Some sort of cache for the current session and its statements might make sense.

cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/utils/CassandraTestHelper.java
```diff
@@ -0,0 +1,143 @@
...
+import static com.datastax.driver.core.querybuilder.QueryBuilder.eq;
+import static com.datastax.driver.core.querybuilder.QueryBuilder.quote;
+import static com.datastax.driver.core.querybuilder.QueryBuilder.select;
+
+/**
+ * Utility methods for test suite support.
+ *
+ * @author Jonathan Halliday
+ */
+public class CassandraTestHelper implements TestableGridDialect {
```
gunnarmorling: Could you add an element for Cassandra to `GridDialectType`? That construct is not perfect, but it allows us to skip specific tests on certain backends.

gunnarmorling:
@jhalliday Done with the review. Many thanks for working in this! Looking forward to association handling, too :)

jhalliday:
@gunnarmorling cheers, that little lot will keep me busy a while, but I don't see anything requiring major rework. phew.   Discussion preferred here or on the mailing lists?

gunnarmorling:
> Discussion preferred here or on the mailing lists?

I would say let's discuss specific remarks right here inline just as needed
and take more general discussions to the mailing list (e.g. strategies for
mapping associations or things like that).

gunnarmorling:
@jhalliday, could you also add an equivalent to `MongoDBBackendTckHelper` to that backend? That's kind of useful for running single tests from the TCK against a specific backend (it needs the _io.takari.junit:takari-cpsuite_ dependency in the POM). Thx!

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraSequenceHandler.java
jhalliday: The full info is indeed available at table creation time, but since it's not available at query time (i.e. in the NextValueRequest) we're left with the options of a) keeping it in some lookup structure on the sequencehandler class or b) just using hardcoded values that we know a-priori. The current impl is the latter option. let me know if you want to switch it.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraSequenceHandler.java
jhalliday: derp. gone.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraDatastoreProvider.java
jhalliday: methods moved. metaDataCache has to be somewhere - I left it on the datastore because the alternative means linking production code to the test helper i.e. the table creation code would have to populate the cache in the test helper.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraSequenceHandler.java
jhalliday: Absolutely. That one belongs in the latter half of the 'first make it work, then make it fast' process though...  Actually I'm of the opinion that since the functionality is not application (or in this case middleware) specific it really belongs in the driver layer. I'll probably try to get something pushed upstream to that and fall back to sticking it in the OGM module if I can't. For now the warning are ugly but at least early adopters have something to kick the tyres on.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraSequenceHandler.java
jhalliday: Counters in cassandra are somewhat weird beasts that don't work the same way as everything else. In particular, you can't do CAS ('lightweight transactions' in cassandra speak) on them. It's like an AtomicLong you can't put a synchronised block around, so you can't read and update in a way that would allow you to build seq ids on them.  UUID is really the way to go for creating synthetic id values in cassandra, I don't expect many users to utilise the equivalent of 'auto generate col seq num' approach as they would in a relational database, but at least with this clunky approach they have the option if they really want to.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/model/impl/ResultSetTupleSnapshot.java
jhalliday: It's not a constant, it's runtime depending on the cluster you're connecting to. If it's not here then it's in the dialect and all the tuples need a backpointer to that instead. Really a design flaw in the Row/DataType - it should know how to deserialize itself, but can't really cope with dynamic typing. It's app friendly, not framework friendly. /sigh.  We'll just have to live with it unless we can get a change pushed upstream.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/type/impl/CassandraCalendarDateType.java
jhalliday: Internal storage in cassandra 2.1 is actually long from epoch, just as inside java's Date.  The TZ handling doc is just about how the server interprets Strings to convert them. We're sending it longs instead, so handling the TZ is our problem.  Note that things are in a state of flux though - this code may need revision for https://issues.apache.org/jira/browse/CASSANDRA-7523

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
jhalliday: In its current form yes. That's one of the chunks of experimental stuff I've not stripped out - it should probably be confined to a branch for now. Eventually I want to use atomic batches to ensure certain invariants in the data storage.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
jhalliday: yup. The general idea in cassandra is you never read before write, so you can't tell which you need. Hence they behave (more or less) the same. Once upon a time there were some weird inconsistencies in the way tables with only primary key columns behaved, but due to other limitations we can only work on versions recent enough not to have those issues anyhow.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
jhalliday: done.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraDatastoreProvider.java
jhalliday: hmm, weird, the driver should deal with that since it is designed for talking to a cluster in which nodes come and go routinely. Maybe some edge case in which it's assuming the cluster as a whole is robust enough that at least one node is reachable.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraDatastoreProvider.java
jhalliday: Done.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraSequenceHandler.java
gunnarmorling: You can get `IdSourceKey` and `IdSourceKeyMetadata`from `NextValueRequest`. That does not cut it?

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/type/impl/CassandraCalendarDateType.java
gunnarmorling: > Internal storage in cassandra 2.1 is actually long from epoch, just as inside java's Date [...] We're sending it longs instead, so handling the TZ is our problem

Yes, that's what I mean, couldn't we pass String (and store) instead in order to convey the TZ information? I find silently loosing the TZ information from calendars a bit troubling. At least a warning should be raised, that `Calendar` properties can be stored but will be read back without the TZ information.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraSequenceHandler.java
jhalliday: hmm yes the IdSourceKey has only the key col not the value col, but I could chain through it to the metadata which has both.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/type/impl/CassandraCalendarDateType.java
jhalliday: yup, at the cost of giving up any server side query ability, since CQL3 lacks the date manipulation functions that SQL has. May still be worth it though.  I'm inclined to go with what we've got for now and revisit it once the new types are available in CQL.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraSequenceHandler.java
jhalliday: done.

jhalliday:
ok, I think we've reached the point of diminishing returns on this lot. Time to look at the test suite refactoring...

gunnarmorling:
Jenkins, retest this please.

core/src/main/java/org/hibernate/ogm/datastore/impl/AvailableDatastoreProvider.java
```diff
@@ -20,7 +20,8 @@
 	MONGODB( "org.hibernate.ogm.datastore.mongodb.impl.MongoDBDatastoreProvider" ),
 	FONGO( "org.hibernate.ogm.datastore.mongodb.impl.FongoDBDatastoreProvider" ),
 	NEO4J_EMBEDDED( "org.hibernate.ogm.datastore.neo4j.impl.Neo4jDatastoreProvider" ),
-	COUCHDB_EXPERIMENTAL( "org.hibernate.ogm.datastore.couchdb.impl.CouchDBDatastoreProvider" );
+	COUCHDB_EXPERIMENTAL( "org.hibernate.ogm.datastore.couchdb.impl.CouchDBDatastoreProvider" ),
+	CASSANDRA( "org.hibernate.ogm.datastore.cassandra.impl.CassandraDatastoreProvider" );
```
gunnarmorling: I will rename this into `CASSANDRA_EXPERIMENTAL` to reflect the current status (as we did for CouchDB).

gunnarmorling:
Rebased and merged :) Added two more commits for some clean-up + minor improvements. Thanks, @jhalliday; Great to have this in!

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
gunnarmorling: Got it now; Unlike with SQL, DELETE will only delete single columns from the row in this case.

PR #519, opened 2015-04-14
OGM-797 Updating plug-in for starting MongoDB so build works on JDK 9

gunnarmorling:


gunnarmorling:
Merged myself, as it's a trivial fix.

PR #521, opened 2015-04-29
OGM-783 Double embedded returned

DavideD:
https://hibernate.atlassian.net/browse/OGM-783

I think I've finally fixed the query with Neo4j.
The only problem remained is that embeddedble inside collections are not mapped correctly and therefore some of the tests don't pass.

This error should be same as  [OGM-669](https://hibernate.atlassian.net/browse/OGM-669) and [OGM-801](https://hibernate.atlassian.net/browse/OGM-801) and should be solved when we will move to ORM 5.

What do you think, should we try to solve them for this version or should we wait untill we move to ORM 5?

I'm quite surprised they pass for MongoDB

### DIFF
--- a/core/src/test/java/org/hibernate/ogm/backendtck/queries/QueriesWithEmbeddedCollectionTest.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/queries/QueriesWithEmbeddedCollectionTest.java
@@ -15,7 +15,9 @@ import java.util.List;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
 import org.hibernate.Transaction;
+import org.hibernate.ogm.utils.GridDialectType;
 import org.hibernate.ogm.utils.OgmTestCase;
+import org.hibernate.ogm.utils.SkipByGridDialect;
 import org.hibernate.ogm.utils.TestSessionFactory;
 import org.junit.After;
 import org.junit.AfterClass;
@@ -99,30 +101,35 @@ public class QueriesWithEmbeddedCollectionTest extends OgmTestCase {
 	}
 
 	@Test
+	@SkipByGridDialect(value = GridDialectType.NEO4J)
 	public void testBetweenOperatorWithEmbeddedInEmbeddedCollection() throws Exception {
 		List<?> result = session.createQuery( "FROM WithEmbedded e JOIN e.anEmbeddedCollection c WHERE c.embedded.embeddedInteger BETWEEN -100 AND 100" ).list();
 		assertThat( result ).onProperty( "id" ).containsOnly( 1L );
 	}
 
 	@Test
+	@SkipByGridDialect(value = GridDialectType.NEO4J)
 	public void testLikeOperatorWithEmbeddedInEmbeddedCollection() throws Exception {
 		List<?> result = session.createQuery( "FROM WithEmbedded e JOIN e.anEmbeddedCollection c WHERE c.embedded.embeddedString LIKE 'string[1%'" ).list();
 		assertThat( result ).onProperty( "id" ).containsOnly( 1L );
 	}
 
 	@Test
+	@SkipByGridDialect(value = GridDialectType.NEO4J)
 	public void testEqualOperatorWithEmbeddedInEmbeddedCollectionForString() throws Exception {
 		List<?> result = session.createQuery( "FROM WithEmbedded e JOIN e.anEmbeddedCollection c WHERE c.embedded.embeddedString = 'string[1][0]'" ).list();
 		assertThat( result ).onProperty( "id" ).containsOnly( 1L );
 	}
 
 	@Test
+	@SkipByGridDialect(value = GridDialectType.NEO4J)
 	public void testEqualOperatorWithEmbeddedInEmbeddedCollectionForInteger() throws Exception {
 		List<?> result = session.createQuery( "FROM WithEmbedded e JOIN e.anEmbeddedCollection c WHERE c.embedded.embeddedInteger = 10" ).list();
 		assertThat( result ).onProperty( "id" ).containsOnly( 1L );
 	}
 
 	@Test
+	@SkipByGridDialect(value = GridDialectType.NEO4J)
 	public void testConjunctionOperatorEqualOperatorWithEmbeddedInEmbeddedCollection() throws Exception {
 		List<?> result = session.createQuery( "FROM WithEmbedded e JOIN e.anEmbeddedCollection c WHERE c.item = 'item[1]' AND c.embedded.embeddedString IN ('string[1][0]')" ).list();
 		assertThat( result ).onProperty( "id" ).containsOnly( 1L );
--- a/core/src/test/java/org/hibernate/ogm/backendtck/queries/QueriesWithEmbeddedCollectionTest.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/queries/QueriesWithEmbeddedCollectionTest.java
@@ -8,6 +8,7 @@ package org.hibernate.ogm.backendtck.queries;
 
 import static org.hibernate.ogm.utils.OgmAssertions.assertThat;
 
+import java.util.ArrayList;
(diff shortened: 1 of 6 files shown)


core/src/test/java/org/hibernate/ogm/backendtck/queries/QueriesWithEmbeddedCollectionTest.java
```diff
@@ -200,6 +220,66 @@ private static void persist(Object... entities) {
...
+			if ( !( result instanceof Object[] ) ) {
+				throw new IllegalArgumentException( "No projection result: " + result );
+			}
+			projectionResults.add( ProjectionResult.forArray( (Object[]) result ) );
+		}
+
+		return projectionResults;
+	}
+
+	private static class ProjectionResult {
```
gunnarmorling: I think I've seen that before somewhere else? Maybe you can extract it into test-utils?

core/src/test/java/org/hibernate/ogm/backendtck/queries/QueriesWithEmbeddedCollectionTest.java
```diff
@@ -93,35 +96,40 @@ public void testEqualEmbeddedCollectionWithEmbeddableInCollectionWhereClause() t
 
 	@Test
 	public void testConjunctionOperatorWithEmbeddedInEmbeddedCollection() throws Exception {
-		List<?> result = session.createQuery( "FROM WithEmbedded e JOIN e.anEmbeddedCollection c WHERE c.item = 'item[1]' AND c.anotherItem IN ('secondItem[0]')" ).list();
+		List<?> result = session.createQuery( "FROM WithEmbedded e JOIN e.anEmbeddedCollection c WHERE c.item = 'item[0]' AND c.anotherItem IN ('secondItem[0]')" ).list();
 		assertThat( result ).onProperty( "id" ).containsOnly( 1L );
 	}
 
 	@Test
+	@SkipByGridDialect(value = GridDialectType.NEO4J)
```
gunnarmorling: Why are these excluded? A short explanation in `comment()` would be nice.

DavideD: I think the error is the same as [OGM-801](https://hibernate.atlassian.net/browse/OGM-801), the nested embedded in a collection skip one level while the cypher query assumes that the mapping is correct. This causes an exception since they cannot found the element.

The error I fixed with this patch made this tests pass but now that it's fixed they don't work anymore. Not sure why it is working for the other datastores.

core/src/test/java/org/hibernate/ogm/backendtck/queries/QueriesWithEmbeddedCollectionTest.java
DavideD: ok

core/src/test/java/org/hibernate/ogm/backendtck/queries/QueriesWithEmbeddedCollectionTest.java
DavideD: I'll add a comment.
I was just wondering if we should try to fix this or way for ORM 5

gunnarmorling: If it is relatively easily fixable prior to ORM 5, we can do it. I am
looking into this issue as we speak.

gurkerl83:
The proposed pull request is working. Unfortunately it is limited to embedded. Is there a way to query entities based on **partial** composite key information through **embedded[Ids]**.

```
public void embeddedIdQuery() throws Exception {
	
	EntityManager entityManager = emf.createEntityManager();

	OgmSession session = (OgmSession) HibernateUtils.getHibernateSession(entityManager);
	Transaction transaction = session.beginTransaction();
	
	try {
		
        // only the firstName information is available for query. 
		List<?> result1 = session.createQuery( "from Person e where *e.id.firstName* = 'Emmanuel'" ).list();
		
        // the entire composite key elements have to be available in order to identify a entity
		PersonPK emmanuelPk = new PersonPK();
		emmanuelPk.setFirstName( "Emmanuel" );
		emmanuelPk.setLastName( "Bernard" );
		
		Object result2 = session.get(Person.class, emmanuelPk);
		
		transaction.commit();
	}
	finally {
		session.clear();
		session.close();
	}
}

@Entity
@Indexed
public class Person {

	@EmbeddedId
	@FieldBridge(impl = PersonPKBridge.class)
	@DocumentId
	private PersonPK id;

	private String favoriteColor;

	public PersonPK getId() {
		return id;
	}

	public void setId(PersonPK id) {
		this.id = id;
	}

	@Field
	public String getFavoriteColor() {
		return favoriteColor;
	}

	public void setFavoriteColor(String favoriteColor) {
		this.favoriteColor = favoriteColor;
	}
}

@Embeddable
public class PersonPK implements Serializable {

	private static final long serialVersionUID = 1L;

	private String firstName;
	private String lastName;

	public String getLastName() {
		return lastName;
	}

	public void setLastName(String lastName) {
		this.lastName = lastName;
	}

	public String getFirstName() {
		return firstName;
	}

	public void setFirstName(String firstName) {
		this.firstName = firstName;
	}

	@Override
	public boolean equals(Object o) {
		if ( this == o ) {
			return true;
		}
		if ( !( o instanceof PersonPK ) ) {
			return false;
		}
		PersonPK personPK = (PersonPK) o;

		if ( firstName != null ? !firstName.equals( personPK.firstName ) : personPK.firstName != null ) {
			return false;
		}
		if ( lastName != null ? !lastName.equals( personPK.lastName ) : personPK.lastName != null ) {
			return false;
		}

		return true;
	}

	@Override
	public int hashCode() {
		int result;
		result = ( firstName != null ? firstName.hashCode() : 0 );
		result = 31 * result + ( lastName != null ? lastName.hashCode() : 0 );
		return result;
	}
}

public class PersonPKBridge implements TwoWayFieldBridge {

	@Override
	public Object get(String name, Document document) {
		PersonPK id = new PersonPK();
		IndexableField field = document.getField( name + ".firstName" );
		id.setFirstName( field.stringValue() );
		field = document.getField( name + ".lastName" );
		id.setLastName( field.stringValue() );
		return id;
	}

	@Override
	public String objectToString(Object object) {
		PersonPK id = (PersonPK) object;
		StringBuilder sb = new StringBuilder();
		sb.append( id.getFirstName() ).append( " " ).append( id.getLastName() );
		return sb.toString();
	}

	@Override
	public void set(String name, Object value, Document document, LuceneOptions luceneOptions) {
		PersonPK id = (PersonPK) value;

		//store each property in a unique field
		luceneOptions.addFieldToDocument( name + ".firstName", id.getFirstName(), document );

		luceneOptions.addFieldToDocument( name + ".lastName", id.getLastName(), document );

		//store the unique string representation in the named field
		luceneOptions.addFieldToDocument( name, objectToString( id ), document );

	}

}

public void embeddedIdPersist() {
	PersonPK emmanuelPk = new PersonPK();
	emmanuelPk.setFirstName( "Emmanuel" );
	emmanuelPk.setLastName( "Bernard" );
	Person emmanuel = new Person();
	emmanuel.setFavoriteColor( "Blue" );
	emmanuel.setId( emmanuelPk );

	PersonPK petePk = new PersonPK();
	petePk.setFirstName( "Pete" );
	petePk.setLastName( "Drexler" );
	Person pete = new Person();
	pete.setFavoriteColor( "Green" );
	pete.setId( petePk );

	persist(emmanuel, pete);
}

```

DavideD:
I'm closing this pull request for now and I'm going to send a new one which integrate @gunnarmorling work on OGM-801.

@gurkerl83 Thanks for letting us know the error on embedded id, I will have a look at it as a different issue.

DavideD:
Replaced by https://github.com/hibernate/hibernate-ogm/pull/531

PR #523, opened 2015-04-30
OGM-249 First draft of CouchBase backend

ewa01s:
I Implemented CRUD for tuples - first phase of backend for CouchBase, is it possible to assign task ([OGM-249](https://hibernate.atlassian.net/browse/OGM-249)) to me ([ewastawicka](https://hibernate.atlassian.net/secure/ViewProfile.jspa?name=ewastawicka))?


gunnarmorling:
@ewa01s, many thanks for following up on this! If you haven't done so, could you sign the JBoss [contributor license agreement](https://cla.jboss.org/index.seam)? We need that so we can integrate your work.

gunnarmorling:
ok to test.

gunnarmorling:
@ewa01s, you are assigned to the right group in JIRA now, the issue is assigned to you :)

gunnarmorling:
Going to check it out in the afternoon.

couchbase/src/main/java/org/hibernate/ogm/datastore/couchbase/dialect/backend/json/impl/EntityDocument.java
```diff
@@ -0,0 +1,234 @@
...
+ *
+ * Implementation note: The entity's properties are stored in a map, with embedded properties being represented by
+ * dot-separated property paths. When (de-)serializing this document from/to JSON via Jackson, this flat representation
+ * is converted into a hierarchical representation using nested maps (if embedded properties are present).
+ *
+ * @author Stawicka Ewa
+ */
+@JsonSerialize(include = Inclusion.NON_NULL)
+@JsonTypeName(EntityDocument.TYPE_NAME)
+public class EntityDocument {
```
gunnarmorling: This looks very similar to the implementation in the _couchdb_ module. Maybe we could extract a common implementation and use it for both?

ewa01s:
> This looks very similar to the implementation in the couchdb module. Maybe we could extract a common implementation and use it for both?

I tried, the problem for me was how to keep revision field in all inheritied couchdb classes and not in couchbase classes, is there a better solution than [b2c2910](https://github.com/ewa01s/hibernate-ogm/commit/b2c2910d1772a05a96bb666b81bcf0bb0d5c848c)?

couchbase/src/main/java/org/hibernate/ogm/datastore/couchbase/dialect/backend/json/impl/CouchBaseDocument.java
```diff
@@ -0,0 +1,108 @@
...
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectReader;
+
+/**
+ * @author Stawicka Ewa
+ */
+public class CouchBaseDocument extends AbstractDocument {
```
gunnarmorling: So what is the difference to a CouchDB document?

couchbase/src/main/java/org/hibernate/ogm/datastore/couchbase/dialect/type/impl/CouchBaseTypeContributor.java
```diff
@@ -0,0 +1 @@
...

public class CouchBaseTypeContributor implements TypeContributor {

	@Override
	public void contribute(TypeContributions arg0, ServiceRegistry arg1) {
		// TODO Auto-generated method stub

	}

}
```
gunnarmorling: Is that one even needed?

couchbase/src/main/java/org/hibernate/ogm/datastore/couchbase/impl/CouchBaseDatastoreProvider.java
```diff
@@ -0,0 +1,103 @@
...
+
+	@Override
+	public Class<? extends SchemaDefiner> getSchemaDefinerType() {
+		// TODO Auto-generated method stub
+		return null;
+	}
+
+	@Override
+	public boolean allowsTransactionEmulation() {
+		// TODO Auto-generated method stub
```
gunnarmorling: This should be true; That we it can be used with `RESOURCE_LOCAL` transaction type instead of requiring a JTA implementation.

couchbase/src/main/java/org/hibernate/ogm/datastore/couchbase/logging/impl/LoggerFactory.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+package org.hibernate.ogm.datastore.couchbase.logging.impl;
+import org.jboss.logging.Logger;
+
+/**
+ * @author Stawicka Ewa
+ */
+public class LoggerFactory {
+	public static Log getLogger() {
+		return Logger.getMessageLogger( Log.class, "CouchBase" );
```
gunnarmorling: Better pass the caller class name as a category, that's what we do for (most of) the other backends. E.g. check out the logger factory in core.

couchbase/src/main/java/org/hibernate/ogm/datastore/couchbase/util/impl/DatabaseIdentifier.java
```diff
@@ -0,0 +1,68 @@
...
+public class DatabaseIdentifier {
+	private static final String PROTOCOL = "http";
+	private static final String SLASH = "/";
+	private final String host;
+	private final int port;
+	private final String databaseName;
+	private final String password;
+	private final URI serverUri;
+	private final URI databaseUri;
+	public DatabaseIdentifier(String host, int port, String databaseName, String password)
```
gunnarmorling: Seems like its copied from the CouchDB backend? It'd be nice if we could unify it.

couchbase/src/main/java/org/hibernate/ogm/datastore/couchbase/logging/impl/Log.java
```diff
@@ -0,0 +1,36 @@
...
+import org.jboss.logging.annotations.Message;
+import org.jboss.logging.annotations.MessageLogger;
+
+/**
+ * @author Stawicka Ewa
+ */
+@MessageLogger(projectCode = "OGM")
+public interface Log extends org.hibernate.ogm.util.impl.Log {
+	@LogMessage(level = INFO)
+	@Message(id = 1901, value = "Connecting to CouchBase at %s")
```
gunnarmorling: You can start with 1701; Currently 1601 - 1700 is the upermost interval in use (by Cassandra).

couchbase/src/test/java/org/hibernate/ogm/datastore/couchbase/utils/CouchBaseTestHelper.java
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	private static boolean isNotNull(String property) {
+		return property != null && property.length() > 0 && ! "null".equals( property );
+	}
+
+	@Override
+	public long getNumberOfEntities(SessionFactory sessionFactory) {
+		// TODO Auto-generated method stub
+		return 0;
```
gunnarmorling: That should be implemented to make some of the tests work.

gunnarmorling:
Nice work, @ewa01s :) Many thanks! Some general thoughts/questions:

* It would be great if we could avoid duplications between CouchDB and CouchBase as far as possible
* Each class should have a class-level comment describing its purpose, important design decisions and alike
* How does CouchBase deal with id sequences? Is there such thing at all? Or should one preferably work with UUIDs for instance?
* Could you exclude all those tests of the backend TCK not yet passing? You can e.g. check out the Cassandra POM file which explicitely includes only a single test atm. (I guess some more will be passing in your case). So we have a build which actually succeeds, we then can include more and more tests of the suite as more features (e.g. associations) get implemented.

couchbase/src/main/java/org/hibernate/ogm/datastore/couchbase/dialect/backend/json/impl/CouchBaseDocument.java
ewa01s: CouchBase doesn't have support for versioning documents, so I think it doean't make sense to keep revision field in the instances of CouchBase documents.

ewa01s: That problem doesn't exist anymore since I applied another solution for creating object for CouchBase Java Api.

couchbase/src/main/java/org/hibernate/ogm/datastore/couchbase/dialect/type/impl/CouchBaseTypeContributor.java
ewa01s: No ;)

Sanne:
I just noticed there's a new commit: @ewa01s sorry we didn't notice as github won't notify anyone if you add a commit on an existing pull request.
What's the update? I see ci.hibernate.org are quite happy, did you address all of @gunnarmorling 's suggestions?

ewa01s:
@Sanne thanks for pointing me! Next time when I add a commit I will add a comment as well.

- [x] I changed approach of connecting, now CouchBaseDialect works on CouchBase Java API models (JsonDocument, JsonObject) so we don't need Jackson, and APIs from couchdb module anymore :)
- [x] I excluded tests which failure

>did you address all of @gunnarmorling 's suggestions?

Not yet, I'm still working on:
- [ ] change logger factory
- [ ] move DatabaseIdentifier and Identifier to common module
- [ ] getNumberOfEntities
- [ ] better documentations of classes

>How does CouchBase deal with id sequences? Is there such thing at all? Or should one preferably work with UUIDs for instance?

Yes, it has [Atomic operations](http://docs.couchbase.com/developer/java-2.1/documents-atomic.html), we can make use of that.

I would also like to develop, in this pr, assosiations, we should have then quite powerful module.

gunnarmorling:
@ewa01s Glad to hear that you are making progress. Wrt. to associations I'd say let's focus on everything else in this PR and then have another one for association handling. No need to do everything at once, and the larger a single PR gets, the tougher it becomes to review it :)

That being said, I'd expect quite a few tests to fail due to the missing assoc support. Can you exclude those in the POM temporarily similar to how its done for Cassandra? That way a `mvn clean install` should pass without failure. You can then re-enable those tests with the assoc handling PR.

ewa01s:
@gunnarmorling, ok, associations will be implemented later. I've already excluded test which failure, currently 20 of integration tests work for build of <i>couchbase</i>

ewa01s:
I implemented API for id sequences,
I changed LoggerFactory,
I moved method createId from CouchDBDatastore to the Identifier. I moved Identifier to core module, and then replaced fromColumnValues, escapeCharsValuesUsedAsColumnValuesSeparator with methods from StringHelper, any issues not to do that?

Still on the list:
- [ ] Implement counting number of entities
- [ ] Ability do create database (i.e bucket) if requested

DavideD:
@ewa01s Because of a small change in the DocumentStoreConfiguration the current code does not compile with the latest master.  Could You fix it? You only need to change the qya you read the host and the port.

It should be something like this:
```
String host = configuration.getHosts().getFirst().getHost();
Integer port = configuration.getHosts().getFirst().getPort();
```

Thanks

ewa01s:
@DavideD thanks, I corrected it.

gunnarmorling:
Hey @ewa01s, I don't expect you have any plans to continue work on this dialect?

I'm going to close this pull request for now in order to clean up a bit. Should you want to work again on this, that'd be great of course. I'd suggest to get in touch over the mailing list or the chat on IRC so we can discuss next steps. Thanks!

PR #526, opened 2015-05-06
OGM-791 MongoDB Driver 2.13 | MongoDB 3.0 credentials | Fongo update

emmanuelbernard:


pom.xml
```diff
@@ -51,7 +51,7 @@
         <skipUnitTests>false</skipUnitTests>
 
         <!-- The following properties configure the embedded MongoDB instance used for tests -->
-        <embeddedMongoDbVersion>2.6.5</embeddedMongoDbVersion>
+        <embeddedMongoDbVersion>3.0.2</embeddedMongoDbVersion>
```
gunnarmorling: Should we have another profile for 2.6 so we can run tests against it on CI? It's a more general question I guess, should we only test against latest major versions of backends or also previous ones.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/AuthenticationMechanismType.java
```diff
@@ -39,7 +44,25 @@ public MongoCredential createCredential(String username, String databaseName, St
 		public MongoCredential createCredential(String username, String databaseName, String password) {
 			return MongoCredential.createMongoX509Credential( username );
 		}
-	};
+	}
+	,
+	SCRAM_SHA_1 {
```
gunnarmorling: Could you update the docs of `MongoDBProperties.AUTHENTICATION_MECHANISM` to state the new default value? I don't think we need to list the actual values there, a ref to the enum is good enough.

pom.xml
emmanuelbernard: I'm inclined not to. I did test it manually.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/AuthenticationMechanismType.java
emmanuelbernard: ok. Which reminds me I did not look at the documentation either

pom.xml
DavideD: I think as well, we should test only the latest versions. At least for now.
If at some point we receive many requests from the users about using specific versions we might change approach.

emmanuelbernard:
@gunnarmorling Update with the documentation changes.
I also anticipated the `WriteConcern` deprecation of MongoDB driver 3 with a  soft deprecation.

gunnarmorling:
Merged; Thanks.

PR #527, opened 2015-05-07
test changes for cassandra backend

jhalliday:
Exclude tests that cassandra can't support, which in some cases requires moving them to separate files to prevent annotated POJOs being processed if they would result in unsupported mappings.


gunnarmorling:
Hey, could you do a rebase of your branch so it only contains the actual commit to be pushed? We usually work with rebases instead of merges in order to keep the project history linear and thus more easily digestable.

jhalliday:
hmm, so apparently git is not going to cooperate happily with that plan. time for a new one...

PR #528, opened 2015-05-07
test changes for cassandra backend

jhalliday:
Exclude tests that cassandra can't support, which in some cases requires moving them to separate files to prevent annotated POJOs being processed if they would result in unsupported mappings.


gunnarmorling:
Rebased and applied. Thanks! Btw. I took the liberty and altered the commit messages so they have a short description on the first line (as this is what is rendered prominently in many places in the GitHub UI).

PR #531, opened 2015-05-11
Ogm OGM-783 Mapping embeddables in Neo4j

DavideD:
https://hibernate.atlassian.net/browse/OGM-783

This issue was initially about returning the right number of embedded from a query but I've found other problems related to the mapping of embedded elements. It ended up to be bigger than I've initially thought. Most of the changes are in the test part. The commit that changes the dialect is this one: https://github.com/hibernate/hibernate-ogm/commit/c30d8da90a4b9abde1900a6b598484138c6e1162

This pull request fixes:
* the creation of embedded associations, there where some errors in the way properties were mapped when dealing with embedded collections inside of an embedded property.

* [OGM-783](https://hibernate.atlassian.net/browse/OGM-783): the reason multiple embeddable get returned is because of query generated uses the OPTIONAL MATCH clause instead of the MATCH one when the property appear in the WHERE clause of the HQL query

* @gunnarmorling fix for [OGM-801](https://hibernate.atlassian.net/browse/OGM-801)

* A change to the entities used for tests with embeddable. The current solution is to use entities called WithEmbeddable, AnotherEmbeddable, and so on. This is caused of confusion and I've changed them to a set of classes used to store a simple multi-path story game.

* Refactoring of the tests for the mapping of entities and associations in Neo4j. I've created a simple DSL that can be used to define nodes and relationships that should be more concise, readable and understandable than the previous approach.


gunnarmorling:
@DavideD Could you take a look at the build failure? Seems to be a stale JavaDoc link in `AliasResolver`.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/embeddable/EmbeddableMappingTest.java
```diff
@@ -205,13 +205,17 @@ public void testEmbeddableCollection() throws Exception {
...
 		// If the value is not big enough, it gets converted as integer
 		Long id = Long.MAX_VALUE;
-		WithEmbedded with = new WithEmbedded( id, null );
-		with.setAnEmbeddable( new AnEmbeddable( "embedded 1", new AnotherEmbeddable( "string 1", 1 ) ) );
-		with.setYetAnotherEmbeddable( new AnEmbeddable( "embedded 2" ) );
-		with.setAnEmbeddedCollection( Arrays.asList( new EmbeddedCollectionItem( "item[0]", "secondItem[0]", null ), new EmbeddedCollectionItem( "item[1]", null, new AnotherEmbeddable( "string[1][0]", 10 ) ) ) );
-		with.setAnotherEmbeddedCollection( Arrays.asList( new EmbeddedCollectionItem( "another[0]", null, null ), new EmbeddedCollectionItem( "another[1]", null, null ) ) );
-
-		session.persist( with );
+		StoryGame story = new StoryGame( id, null );
```
gunnarmorling: I like the new test :)

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jAssociationQueries.java
```diff
@@ -252,4 +275,343 @@ private EntityKey getEntityKey(AssociationKey associationKey, RowKey rowKey) {
...
+	 * @param embeddedKey the {@link EntityKey} identifying the embedded component
+	 * @return the created {@link Relationship} that represents the association
+	 */
+	public Relationship createRelationshipForEmbeddedAssociation(ExecutionEngine executionEngine, AssociationKey associationKey, EntityKey embeddedKey) {
+		String query = initCreateEmbeddedAssociationdQuery( associationKey, embeddedKey );
+		Object[] queryValues = createRelationshipForEmbeddedQueryValues( associationKey, embeddedKey );
+		return executeQuery( executionEngine, query, queryValues );
+	}
+
+	private String initCreateEmbeddedAssociationdQuery(AssociationKey associationKey, EntityKey embeddedKey) {
```
gunnarmorling: Association _d_ Query

gunnarmorling: How comes this one is not cached as the others?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jAssociationQueries.java
```diff
@@ -38,6 +45,13 @@ public Neo4jAssociationQueries(EntityKeyMetadata ownerEntityKeyMetadata, Associa
 		this.removeAssociationQuery = initRemoveAssociationQuery( ownerEntityKeyMetadata, associationKeyMetadata );
 		this.removeAssociationRowQuery = initRemoveAssociationRowQuery( ownerEntityKeyMetadata, associationKeyMetadata );
 		this.findRelationshipQuery = initFindRelationshipQuery( ownerEntityKeyMetadata, associationKeyMetadata );
+		this.matchOwnerEntityNode = initMatchOwnerEntityNode( ownerEntityKeyMetadata );
+	}
+
+	private String initMatchOwnerEntityNode(EntityKeyMetadata ownerEntityKeyMetadata) {
```
gunnarmorling: Should be static.

gunnarmorling:
> the creation of embedded associations, there where some errors in the way properties were mapped when dealing with embedded collections inside of an embedded property.

Could you give some more details around this? The changes in `Neo4jAssociationQueries` seem quite involved :) Btw. for the migration notes we should have an explanation of what has changed wrt. to the persistent format (maybe some before/after comparison of affected mappings). Doesn't have to be much, just a small example maybe helping users (and me ;) understand the change.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jAssociationSnapshot.java
```diff
@@ -63,6 +68,28 @@ public int size() {
 	}
 
 	private static Iterable<Relationship> relationships(Node ownerNode, AssociationKey associationKey, String relationshipType) {
+		if ( relationshipType.contains( "." ) ) {
```
gunnarmorling: What is the actual semantics of that. Maybe push the condition to method with a descriptive name (think you even did it at another place).

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jAssociationSnapshot.java
```diff
@@ -63,6 +68,28 @@ public int size() {
...
+			String[] pathToAssociation = EMBEDDED_FIELDNAME_SEPARATOR.split( relationshipType );
+			Node nextNode = ownerNode;
+			for ( int i = 0; i < pathToAssociation.length; i++ ) {
+				String splitType = pathToAssociation[i];
```
gunnarmorling: Couldn't you do that via Cypher? E.g. look for arbitrary intermediate relationships/nodes followed by the actual relationship type you are looking for?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jAssociationSnapshot.java
DavideD: Ok, I forgot to add a more expressive method. I'll update it.

Anyway this check if the collection is inside an embedded property by looking for a dot in the collection role

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jAssociationQueries.java
DavideD: ok

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jAssociationSnapshot.java
DavideD: Make sense, althought I suspect this logic should not be here in the first place. I see what I can do

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jAssociationQueries.java
DavideD: The query might be different if some values are null because it might not need to create intermediate nodes.
I'll find an example.

DavideD:
Force pushed:
* Rebased
* Fixed the syntax error in Neo4jAssociationQueries#initCreateEmbeddedAssociationdQuery
* Fixed the javadoc

Without this patch an embedded inside an embedded collection get store in the wrong way.
So, if we use the following mapping as an example
```
@Entity
public class Order {
        ...
	@Embedded
	private ShippingAddress shippingAddress;
...
}

@Embeddable
public class ShippingAddress {

	@Embedded
	private PhoneNumber phone;
...
}

@Embeddable
public class PhoneNumber {

	private String main;

	@ElementCollection
	private List<String> alternatives;

}
```
The association `shippingAddress.phone.alternatives`, is mapped as

```
(order:Order:ENTITY) -[:shippingAddress.phone.alternatives]-> (:EMBEDDED { shippingAddress.phone.alternatives: "ajsfh"})
```

instead of

```
(order:Order:ENTITY) -[:shippingAddress] ->(:EMBEDDED) -[:phone]-> (:EMBEDDED) -[:alternatives]-> (:EMBEDDED {alternatives: "akslh"})
```

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/mapping/UnidirectionalManyToOneTest.java
```diff
@@ -50,35 +53,28 @@ public void prepareDb() throws Exception {
...
-		params.put( "jug", jugProperties );
-		params.put( "e", emmanuelProperties );
-		params.put( "j", jeromeProperties );
-
-		assertExpectedMapping( "jug", jugNode, params );
-		assertExpectedMapping( "e", emmanuelNode, params );
-		assertExpectedMapping( "j", jeromeNode, params );
-		assertExpectedMapping( "r", jugNode + " - [r:memberOf] - " + emmanuelNode, params );
-		assertExpectedMapping( "r", jugNode + " - [r:memberOf] - " + jeromeNode, params );
+		NodeForGraphAssertions jugNode = node( "jug", JUG.class.getSimpleName(), ENTITY.name() )
```
gunnarmorling: FWIW, I'd prefer a style more closely to Cypher, without any parameters etc.:

    assertThatNodeExists("(jug:JUG:ENTITY {jug_id: summer_camp, name: JUG Summer Camp })");
    assertThatRelationshipExists("(jug:JUG:ENTITY {jug_id: summer_camp} - [r:memberOf] - > ...)");

Much more readable IMO. Also more reliable as you actually write/read the expected persistent form.

But no need to change from my side, just to share my view :)

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/mapping/MapTest.java
```diff
@@ -53,62 +56,48 @@ public void prepareDb() throws Exception {
...
+				.property( "nicknames", "idrA" );
+
+		NodeForGraphAssertions nickNode2 = node( "nick2", "Nicks", EMBEDDED.name() )
+				.property( "nicknames", "day[9]" );
+
+		getTransactionManager().begin();
+		ExecutionEngine executionEngine = createExecutionEngine();
+
+		assertThatOnlyTheseNodesExist( executionEngine
+				, userNode
```
gunnarmorling: The leading comma looks weird.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jAssociationQueries.java
```diff
@@ -96,17 +120,22 @@ private static String initRemoveAssociationQuery(EntityKeyMetadata ownerEntityKe
...
 		appendRelationshipType( queryBuilder, associationKeyMetadata );
 		queryBuilder.append( "]" );
 		if ( associationKeyMetadata.getAssociationKind() == AssociationKind.EMBEDDED_COLLECTION ) {
 			queryBuilder.append( " - (e:" );
 			queryBuilder.append( EMBEDDED );
-			queryBuilder.append( ") DELETE r, e" );
+			queryBuilder.append( ")" );
+			queryBuilder.append( " WITH a,e" );
+			queryBuilder.append( "  MATCH path=(e) -[*0..]-> (:EMBEDDED) ");
+			queryBuilder.append( "  FOREACH ( r IN relationships(path) | DELETE r )" );
```
gunnarmorling: Does this cover embeddables within embeddables, too?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jEntityQueries.java
```diff
@@ -54,6 +57,62 @@ public Neo4jEntityQueries(EntityKeyMetadata entityKeyMetadata) {
...
 		this.removeEmbdeddedElementQuery = initRemoveEmbdeddedElementQuery( entityKeyMetadata );
+		this.findAssociationPartialQuery = initFindAssociationPartialQuery( entityKeyMetadata );
+	}
+
+	/*
+	 * Example:
+	 *
+	 * MATCH (owner:ENTITY:AccountOwner {id: {0}})
+	 */
+	private static String initFindAssociationPartialQuery(EntityKeyMetadata entityKeyMetadata) {
```
gunnarmorling: How does it differ from `initFindEntityQuery()`?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jEntityQueries.java
```diff
@@ -54,6 +57,62 @@ public Neo4jEntityQueries(EntityKeyMetadata entityKeyMetadata) {
...
+	 *
+	 * MATCH (owner:ENTITY:Car {`carId.maker`: {0}, `carId.model`: {1}}) -[r:tires]- ()
+	 * RETURN r
+	 *
+	 * or for embedded associations:
+	 *
+	 * MATCH (owner:ENTITY:StoryGame {id: {0}}) -[:evilBranch]-> (:EMBEDDED) -[r:additionalEndings]-> (:EMBEDDED)
+	 * RETURN r
+	 */
+	private String completeFindAssociationQuery(String relationshipType) {
```
gunnarmorling: Shouldn't this be on `Neo4jAssociationQueries`?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jEntityQueries.java
gunnarmorling: Btw. `Neo4jAssociationQueries#initMatchOwnerEntityNode()` looks the same, too?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jTupleAssociationSnapshot.java
```diff
@@ -59,7 +105,50 @@ public Neo4jTupleAssociationSnapshot(Relationship relationship, AssociationKey a
...
+			}
+			return ownerNode;
+		}
+		else {
+			return relationship.getStartNode();
+		}
+	}
+
+	private static boolean isEmbedded(String targetColumnName) {
+		return targetColumnName.contains( "." );
```
gunnarmorling: Use `EmbeddedHelper.isPartOfEmbedded(String)`?

gunnarmorling:
@DavideD, Added some more comments. Looks good in general, I think we can merge it if you address those.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jTupleAssociationSnapshot.java
DavideD: +1

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jEntityQueries.java
DavideD: Initially I've created it there but then I've moved it here because sometimes you need the inverse association of the one you are considering and this query depends mainly on the entityKeyMetadata.

Now that I think about it I could see if I can create a function like findAssociationOnTheMainSide(..) when I have the associationKey.

I'll have a look

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jAssociationQueries.java
DavideD: Yes,
this part:
```
MATCH path=(e) -[*0..]-> (:EMBEDDED) 
```

`e` is already defined as an EMBEDDED in the previous part of the query, while the syntax `-[*0..]` select any path tha connect `e` to a (:EMBEDDED) with any number of relationships.

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/mapping/UnidirectionalManyToOneTest.java
DavideD: I wanted to know which properties you expected to find in the node and give more information about the properties on the element. In particular I don't want unexpected properties in the node or relationship that I matched. Now that I think about it I don't check the labels.

With the cypher query, you only know if the pattern you are looking for has a match or not.

DavideD:
@gunnarmorling I've applied the remarks.
In the end I've decided to leave findAssociation in Neo4jEntityQueries because for *ToOne scenarios at the moment we don't create queries for associations and it was starting to become compilcated to create an associationKey for it.
I'll check again in the future, but for now I would leave it like this

PR #532, opened 2015-05-11
OGM-122 exclude or refactor tests not supported by cassandra backend

jhalliday:


core/src/test/java/org/hibernate/ogm/backendtck/optimisticlocking/OptimisticLockingExtraTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Test for detecting concurrent updates by dialects which support atomic find/update semantics or have their own
+ * optimistic locking scheme.
+ *
+ * @author Gunnar Morling
+ */
+@SkipByGridDialect(
+		value = { GridDialectType.CASSANDRA },
+		comment = "Galaxy.stars list - bag semantics unsupported (no primary key)"
```
DavideD: Can we avoid to add the name of the class in the comment? It's very easy to forget about it if we rename forget about something.

It think, saying bag semantic not supported or collection of embeddable not supported is enough since our entity classes are aimed to test a specific scenario.

gunnarmorling:
Rebased and applied. Thanks, @jhalliday.

PR #533, opened 2015-05-11
Support WildFly 9

Sanne:
This has been tested successfully with WildFly 9.0.0.CR2-SNAPSHOT, but the PR refers to 9.0.0.CR1 (both work).

Some points of interest:
 - I had to better hide the Lucene version used by Neo4J as classloader precedence in WF9 would otherwise expose it to the wrong Lucene version
 - the custom Lucene module uses the 'org.hibernate.ogm.neo4j.lucene' module id to prevent slot conflicts with other projects (this being the one provided by OGM)
 - turns out some slot definitions were wrong 
 - this does replace (remove) the support for WildFly 8, as agreed on the mailing list.
 - the artifact id for Arquillian changed

The download/unpacking of the Hibernate Search and Infinispan modules for integration tests - not needed in this specific version - is simply commented out as I don't think it's worth to fight Maven to automate that: we'll probably need that soon again, and it's going to be impossible to forget about that w/o having tests to complain.

### DIFF
--- a/documentation/manual/pom.xml
+++ b/documentation/manual/pom.xml
@@ -91,6 +91,8 @@
                         <hibernate-orm-version>${hibernateVersion}</hibernate-orm-version>
                         <hibernate-search-version>${hibernateSearchVersion}</hibernate-search-version>
                         <hibernate-search-module-slot>${hibernate-search.module.slot}</hibernate-search-module-slot>
+                        <hibernate-search-major-minor-version>${hibernate-search-major-minor-version}</hibernate-search-major-minor-version>
+                        <hibernate-search-full-module-id>${hibernate-search-full-module-id}</hibernate-search-full-module-id>
                         <infinispan-version>${infinispanVersion}</infinispan-version>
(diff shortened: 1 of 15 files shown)


gunnarmorling:
Looks good to me. Only the integration test for CouchDB [is failing](http://ci.hibernate.org/job/hibernate-ogm-PR/556/testReport/junit/org.hibernate.ogm.test.integration.jboss/CouchDBModuleMemberRegistrationIT/org_hibernate_ogm_test_integration_jboss_CouchDBModuleMemberRegistrationIT/). Could you take a look at that one? Apart from that it's good to merge.

Sanne:
thanks, I'm looking. I guess that one didn't run locally as I don't remember installing couchdb on my new system (fresh install last weekend).

gunnarmorling:
Yes, we compile the CouchDB-related things but don't run tests, unless COUCHDB_HOSTNAME is set.

gunnarmorling:
Same for Cassandra btw now. You may want to install that as well :)

Sanne:
Added a commit to fix the CouchDB modules

PR #534, opened 2015-05-12
OGM-806 Renaming error handler API

gunnarmorling:
This renames affected packages and some references in the documentation. Note that I have left the main contract, `ErrorHandler`, unchanged because that's what it really does, it handles errors. I am not super-happy with the new name because we do not yet have any means of automated compensation, but I guess it's better to do now than later. And users can try to  compensate manually.


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/534

Thanks

PR #535, opened 2015-05-12
OGM-807 Adding clarification to reference guide

gunnarmorling:


Sanne:
I've merged it as it seems correct, but aesthetically this big red box is getting huge..

PR #536, opened 2015-05-12
OGM-826 Don't run @BeforeClass/@AfterClass methods if entire test cla…

gunnarmorling:
…ss is skipped


Sanne:
thanks, integrated.

PR #537, opened 2015-05-13
OGM-717 Adding support for ORM's different boolean mapping styles

gunnarmorling:


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/537

Thanks

PR #539, opened 2015-05-15
OGM-662

emmanuelbernard:
Probably can be merged but it's more to get feedback that this pull request is opened.
Known remaining tasks:
- update the documentation
- add support from the programmatic API

But worse case that can wait until after the next release.

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/cfg/OgmProperties.java
+++ b/core/src/main/java/org/hibernate/ogm/cfg/OgmProperties.java
@@ -50,12 +50,19 @@ public interface OgmProperties {
 
 	/**
 	 * Property for setting the host name to connect to. Accepts {@code String}.
+	 * Accepts a comma separated list of host / ports.
+	 * Note that for IPv6, the host must be surrounded by square bracket if a port is defined: [2001:db8::ff00:42:8329]:123
+	 *
+	 * For example
+	 * www.example.com, www2.example.com:123, 192.0.2.1, 192.0.2.2:123, 2001:db8::ff00:42:8329, [2001:db8::ff00:42:8329]:123
 	 */
 	String HOST = "hibernate.ogm.datastore.host";
 
 	/**
 	 * Property for setting the port number of the database to connect to. Accepts {@code int}.
+	 * @deprecated ignored when multiple hosts are defined in {@link #HOST}, prefer the {@code host:port} approach
 	 */
+	@Deprecated
 	String PORT = "hibernate.ogm.datastore.port";
 
 	/**
--- a/core/src/main/java/org/hibernate/ogm/cfg/impl/HostParser.java
+++ b/core/src/main/java/org/hibernate/ogm/cfg/impl/HostParser.java
@@ -1,0 +1,111 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.cfg.impl;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.hibernate.ogm.cfg.spi.Hosts;
+import org.hibernate.ogm.util.impl.Log;
+import org.hibernate.ogm.util.impl.LoggerFactory;
+
+/**
+ * Parser for the host property.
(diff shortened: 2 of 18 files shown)


core/src/main/java/org/hibernate/ogm/util/configurationreader/impl/SimplePropertyReaderContext.java
```diff
@@ -96,6 +99,26 @@ else if ( configuredValue instanceof Number ) {
 		}
 	}
 
+	private Integer getAsInteger() {
```
gunnarmorling: `getAsInt()` could be implemented on top of this:

    Integer value = getAsInteger();
    return value != null ? value : 0;

core/src/main/java/org/hibernate/ogm/cfg/OgmProperties.java
```diff
@@ -50,12 +50,18 @@
 
 	/**
 	 * Property for setting the host name to connect to. Accepts {@code String}.
+	 * Accepts a comma separated list of host / ports.
```
gunnarmorling: I'd explicitly mention the "bracket style" for IPv6.

core/src/main/java/org/hibernate/ogm/cfg/impl/HostParser.java
```diff
@@ -0,0 +1,113 @@
...
+			return Hosts.NO_HOST;
+		}
+		// for each element between commas
+		String[] splits = hostString.split( "," );
+		for (String rawSplit : splits) {
+			// remove whitespaces
+			String split = rawSplit.trim();
+			//
+			Matcher matcher = HOST_AND_PORT_PATTERN.matcher( split );
+			String host = null;
```
gunnarmorling: These two vars are unused.

core/src/main/java/org/hibernate/ogm/cfg/spi/Hosts.java
```diff
@@ -0,0 +1,99 @@
...
+	public Hosts(List<String> hosts, List<Integer> ports) {
+		int length = hosts.size();
+		Contracts.assertTrue( length == ports.size(), "Not the same number of hosts and ports" );
+		this.hostsAndPorts = new ArrayList<HostAndPort>( length );
+		for ( int i = 0 ; i < length; i++ ) {
+			hostsAndPorts.add( new HostAndPort( hosts.get( i ), ports.get( i ) ) );
+		}
+	}
+
+	public boolean hasNoHost() {
```
gunnarmorling: I'd avoid the negated method name, e.g. `isEmpty()`.

gunnarmorling:
Looks good to me overall, added some minor comments.

> add support from the programmatic API

Agreed that this can wait. Btw. I am wondering about the usefulness of giving connection-related settings through the API. It's something you typically would like to change without re-compiling the application, so using properties for that seems better suited to me.

Speaking of it, currently (pre ORM 5) we allow for two approaches for plugging in programmatic API configurations:

* naming a `Configurator` impl class via a property
* through `OgmConfiguration#configureOptionsFor()`

With `Configuration` being gone at some point, only the former would be possible. With the new bootstrap API, we can plug in an OGM-specific `SessionFactoryBuilder` which can be accessed through `unwrap()` and allows for OGM-specific bootstrap APIs. But that'd be "too late" for connection related settings, as `DatastoreProvider` (which is the client of those) isn't an SF-level service. It would work though for "runtime" settings, e.g. to control certain aspects of mapping.

emmanuelbernard:
> Speaking of it, currently (pre ORM 5) we allow for two approaches for plugging in programmatic API configurations:
> 
> naming a Configurator impl class via a property
> through OgmConfiguration#configureOptionsFor()
> With Configuration being gone at some point, only the former would be possible. With the new bootstrap API, we can plug in an OGM-specific SessionFactoryBuilder which can be accessed through unwrap() and allows for OGM-specific bootstrap APIs. But that'd be "too late" for connection related settings, as DatastoreProvider (which is the client of those) isn't an SF-level service. It would work though for "runtime" settings, e.g. to control certain aspects of mapping.

Wouldn't it be better to offer it as a Hibernate `Service` and let the user plug an initiator in its service registry? I am not sure how that would look like but that would be at the right time and more natural I guess. thoughts?

emmanuelbernard:
@gunnarmorling I've update things with your remarks and added a commit with the documentation.
Ready to go on my side.

gunnarmorling:
> Wouldn't it be better to offer it as a Hibernate Service and let the user plug an initiator in its service registry?

Hm, maybe. Although the service/initiator stuff seems a bit tough on users. A configurator class made known through some prop still seems nicer to me.

gunnarmorling:
Merged, thanks.

PR #541, opened 2015-05-15
OGM-717 Fix build for Cassandra

DavideD:
@emmanuelbernard  could you let me know if this works?


gunnarmorling:
Looks good, thanks. Merged it. Btw. Cassandra tests weren't run on CI master (CASSANDRA_HOSTNAME was not set), hence we didn't notice it there. I've just changed that.

PR #542, opened 2015-05-18
OGM-830 Storing multiply generators in one CouchDB document

stawny:


gunnarmorling:
Ok to test.

couchdb/src/main/java/org/hibernate/ogm/datastore/couchdb/dialect/backend/json/impl/SequenceDocument.java
```diff
@@ -64,10 +62,12 @@ public void increase(int increment) {
...
 	@JsonAnySetter
 	void set(String name, Object value) {
-		if ( valueProperty == null ) {
-			valueProperty = name;
-		}
-
 		properties.put( name, value );
 	}
+
+	public void enusreValueProperty(String valueProperty, int initialValue) {
```
gunnarmorling: enusre=ensure

gunnarmorling:
Could you add a test for this situation?

PR #538, opened 2015-05-14
OGM-122 cassandra backend association support

jhalliday:

PR #543, opened 2015-05-18
OGM-597 Supporting typed parameters given via Query#setParameter()

gunnarmorling:


gunnarmorling:
@Sanne That's the second piece of the puzzle.

@emmanuelbernard, @DavideD : This requires the parser to be released first.

Sanne:
my plan was to push #545 directly and this one, but since I've got both Cassandra and couch integration failures I'l falling back to less aggressive integration.. could you merge #545 first, rebase and then we verify the failures are somewhat manageable/expected?

gunnarmorling:
Force-pushed to this branch to pull in 1.2.0.Final of the parser.

Wrt. to the failures, I am also seeing some for Cassandra (see my comments on #538), but Couch is fine for me. What failures do you have there?

emmanuelbernard:
Applied.

Sanne:
Looks like my couch failures are caused by a previous run which didn't cleanup.

PR #544, opened 2015-05-18
OGM-781 Support projections for embedded collections

DavideD:
https://hibernate.atlassian.net/browse/OGM-781

Support projection of properties in embedded collections. This will only work for Neo4j and MongoDB.

The projection of a collection does not work at the moment in hibernate search, it will only return the first element and not all the elements of the collection stored in the index.


emmanuelbernard:
The documentation remark is that puzzles me the most, but aside from that it's likely good to go. Let's chat over this tomorrow morning.

PR #548, opened 2015-05-24
OGM-837 Support queries on composite ids on Neo4j and MongoDB

DavideD:
https://hibernate.atlassian.net/browse/OGM-837

This patch allows the execution of queries on MongoDB and Neo4j on the attributes of an embedded id. For example:
 `FROM News n WHERE n.newsId.title = '...'`

Hibernate Search converts the embedded id using a  FieldBridge, Therefore there is non `newsId.title` in the index at the end (only `newsId`). Maybe we should store both values. haat do you think?

### DIFF
--- a/core/src/test/java/org/hibernate/ogm/backendtck/queries/CompositeIdQueriesTest.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/queries/CompositeIdQueriesTest.java
@@ -1,0 +1,120 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.backendtck.queries;
+
+import static org.fest.assertions.Assertions.assertThat;
+import static org.hibernate.ogm.utils.SessionHelper.asProjectionResults;
+import static org.hibernate.ogm.utils.SessionHelper.delete;
+import static org.hibernate.ogm.utils.SessionHelper.persist;
+
+import java.util.List;
+
+import org.hibernate.Session;
+import org.hibernate.SessionFactory;
+import org.hibernate.Transaction;
+import org.hibernate.ogm.backendtck.id.Label;
+import org.hibernate.ogm.backendtck.id.News;
+import org.hibernate.ogm.backendtck.id.NewsID;
+import org.hibernate.ogm.utils.GridDialectType;
+import org.hibernate.ogm.utils.OgmTestCase;
+import org.hibernate.ogm.utils.SessionHelper.ProjectionResult;
+import org.hibernate.ogm.utils.SkipByGridDialect;
+import org.hibernate.ogm.utils.TestSessionFactory;
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+/**
+ * @author Davide D'Alto
+ */
+@SkipByGridDialect(
+	value = { GridDialectType.CASSANDRA, GridDialectType.COUCHDB, GridDialectType.EHCACHE, GridDialectType.HASHMAP, GridDialectType.INFINISPAN },
+	comment = "Hibernate Search does not store properties of the @EmbeddedId by default in the index, it requires the use of @FieldBridge."
+			+ "It is also not sufficient to add a custom field bridge because the properties of the embedded id won't be recognized as properties of the entity."
+			+ "There is a JIRA to keep track of this: OGM-849")
+public class CompositeIdQueriesTest extends OgmTestCase {
+
+	@TestSessionFactory
+	public static SessionFactory sessions;
+
+	private static final String author = "Guillaume";
+
+	private static final String titleOGM = "How to use Hibernate OGM ?";
+	private static final String titleAboutJUG = "What is a JUG ?";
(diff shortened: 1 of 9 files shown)


gurkerl83:
Hi, thanks for investigating in this! I have successfully tested the pull request on our data model with the neo4j datastore. We have not tested it within mongo due to the fact we are not using it.

gunnarmorling:
> Therefore there is non newsId.title in the index at the end (only newsId). Maybe we should store both values. haat do you think? 

I would expect that you could configure this via `@IndexEmbedded` and possibly `@Field` on the properties of the id type? Have you spoken to our HS gurus about it?

DavideD:
> Therefore there is non newsId.title in the index at the end (only newsId). Maybe we should store both values. What do you think?
> 
> I would expect that you could configure this via @IndexEmbedded and possibly @Field on the properties of the id type? Have you spoken to our HS gurus about it?

@Sanne, @hferentschik, any word of wisdom?

DavideD:
> Hi, thanks for investigating in this! I have successfully tested the pull request on our data model with the neo4j datastore. We have not tested it within mongo due to the fact we are not using it.

@gurkerl83L: Thanks for your feedback, that's much appreciated.

DavideD:
@gunnarmorling The feature you suggested is not implemented at the moment but it is possible to obtian the same result using a field bridge. I'll try it

DavideD:
Fixing this issue for all the datastore would require some changes in the HQL parser project. Would that be ok?

I think I would porceed this way:
- Update the documentation to highlight the limitation
- Merge this pull request and upate the JIRA to mention that it's a MongoDB and Neo4j fix.
- Create a new JIRA for the fix related to the backends using Hibernate Search .

What do you think?

DavideD:
I've updated the documentation.
It will also solve [OGM-846](https://hibernate.atlassian.net/browse/OGM-846).

I think this pull can be merged as it this.
I've tried to fix it fo rLucene but it requires more changes thatn I've initially thought and I'm not even sure how to fix it in the end.

I will create a new issue for it if that's OK.

core/src/test/java/org/hibernate/ogm/backendtck/queries/CompositeIdQueriesTest.java
```diff
@@ -0,0 +1,118 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+/**
+ * @author Davide D'Alto
+ */
+@SkipByGridDialect(
+		value = { GridDialectType.CASSANDRA, GridDialectType.COUCHDB, GridDialectType.EHCACHE, GridDialectType.HASHMAP, GridDialectType.INFINISPAN },
+		comment = "These queries won't work with Hibernate Search because it requires a FieldBridge" )
```
gunnarmorling: I am not quite sure I understand what that means: "because it requires a FieldBridge". Is there an issues for this already? Could you reference it here?

documentation/manual/src/main/asciidoc/en-US/modules/query.asciidoc
```diff
@@ -176,6 +196,16 @@ String query =
      "FROM StoryGame sg JOIN sg.optionalEndings ending WHERE ending.text = 'Happy ending'";
 List<String> endings = session.createQuery( query ).list();
 ----
+
+Projection of embedded id properties will work with Neo4j and MongoDB,
```
gunnarmorling: Is it only projection or also referencing embedded id properties in WHERE clauses?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/query/parsing/impl/Neo4jQueryRendererDelegate.java
```diff
@@ -179,39 +179,56 @@ private void returns(StringBuilder builder, String targetAlias) {
...
+			final List<String> associationPath = propertyHelper.findAssociationPath( targetTypeName, pathWithoutAlias );
+			// Currently, it is possible to nest only one association inside an emebedded
+			if ( associationPath != null ) {
+				List<String> nextPath = new ArrayList<String>( associationPath.size() + 1 );
+				int next = associationPath.size();
+				nextPath.addAll( associationPath );
+				if ( next < pathWithoutAlias.size() - 1 ) {
+					nextPath.add( pathWithoutAlias.get( next ) );
+				}
+				boolean leftJoin = false; // TODO: Not sure how to obtain this information at the moment
```
gunnarmorling: Is there a JIRA issue for this TODO? Should we create one?

DavideD: No, I'll create one

documentation/manual/src/main/asciidoc/en-US/modules/query.asciidoc
DavideD: Right, it needs to be better explained. I'll change it

core/src/test/java/org/hibernate/ogm/backendtck/queries/CompositeIdQueriesTest.java
DavideD: EmbeddedId can be store only using a FieldBridge, this means that we don't know how the value is stored in the index. You cannot project/filter the properties of the emebedded id since they might not exist in the index.

I've created this one: https://hibernate.atlassian.net/browse/HSEARCH-1879

DavideD:
@gunnarmorling: I've changed the latest two commits to update the SkipByGridDIalect message and the documentation. Could you have a quick look and tell me if it is any better, please? Thanks

gunnarmorling:
@DavideD, added two more comments. Good to merge otherwise.

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/548

PR #549, opened 2015-05-25
OGM-838 Skipping the deployment of broken EAP modules

hferentschik:


gunnarmorling:
Rebased and applied. Thanks!

PR #551, opened 2015-05-25
OGM-836 Adding Cassandra to the assembly configuration and adding new…

hferentschik:
… directort output format to help inspecting assembly


gunnarmorling:
Rebased and applied. Thanks, @hferentschik.

PR #552, opened 2015-05-26
OGM-808 Removing maven-processor-plugin and do annottion provcessing …

hferentschik:
…via the maven-compiler plugin in order to have a better IDE integration

With this setup one can import the project into Intellij and rebuild the whole project directly w/o prior command line build. One can also modify the `Log` interfaces and just rebuild in the IDE making for a quick dev cycle.

### DIFF
--- a/cassandra/pom.xml
+++ b/cassandra/pom.xml
@@ -32,13 +32,8 @@
                 </configuration>
             </plugin>
             <plugin>
-                <groupId>org.bsc.maven</groupId>
-                <artifactId>maven-processor-plugin</artifactId>
-            </plugin>
-            <plugin>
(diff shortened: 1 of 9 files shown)


Sanne:
The good news is it works flawlessly in Eclipse too.

Unless I'm missing something about JMH, it seems this broke the performance testsuite though?
Or at least it needs new instructions in `org.hibernate.ogm.perftest.Launcher`

gunnarmorling:
Merged. Good catch with the perf suite, @Sanne. I've added one more commit for updating the launcher instructions.

hferentschik:
Awesome and good catch indeed :-)

Sanne:
Thanks!

PR #553, opened 2015-05-26
Cassandra modules for WildFly 9

Sanne:
- https://hibernate.atlassian.net/browse/OGM-813


modules/wildfly/src/main/modules/ogm/cassandra-drivers/module.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ ~ Hibernate OGM, Domain model persistence for NoSQL datastores
+ ~
+ ~ License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ ~ See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+  -->
+<module xmlns="urn:jboss:module:1.1" name="org.hibernate.ogm.cassandra.drivers" slot="${hibernate.ogm.cassandra.module.slot}">
```
gunnarmorling: Out of interest, what's the movation for having a separate module for the driver? It's different from the other backends, where we package grid dialect and JARs required for the driver in one module.

Sanne: Generally I think we should always do this, as it improves isolation.
For example this way we don't force the user to have an outdated version of Guava.

I already had to apply a similar patch to Neo4J to separate the older Lucene version it uses into an isolated module.

Looking forward I'd hope we wouldn't actually maintain the driver modules - in this case the datastax driver but the rule might apply to all - but have a separate project to build that module. Until then it's better this uses the OGM namespace.

For example the Infinispan integration doesn't ship the Infinispan driver, but depends on the module: much cleaner.

Sanne: BTW while I think we should do it across all dialects, I've avoided that for now as it doesn't seem fit for a CR1->Final upgrade.. it requires some attention as for example you don't want to hide too much, for example to allow using native queries where supported.

gunnarmorling: Ok, sounds reasonable. Let's try and do it for others where applicable as well as of 5 then.

gunnarmorling: Why is it "driver *s*" btw. and not "driver"? It's a single one?

Sanne: > Why is it "driver s" btw. and not "driver"? It's a single one?

Not sure. I think it was "driver's dependencies" in my mind. Want me to change it?

gunnarmorling: That's alright, I will do it.

gunnarmorling:
@Sanne, can you check out my PR https://github.com/Sanne/hibernate-ogm/pull/2 against your repo which is based on this. Thanks!

Sanne:
@gunnarmorling this one should work now.

gunnarmorling:
Cool; Thanks,@Sanne!

PR #556, opened 2015-05-27
OGM-827 Setting the mongodb timeout for tests to 1000ms and refactori…

hferentschik:
…ng MongoDBDatastoreProvider#extractDatabase to remove retry logic and be generally simpler

### DIFF
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
@@ -35,7 +35,6 @@ import com.mongodb.MongoClient;
 import com.mongodb.MongoClientOptions;
 import com.mongodb.MongoCredential;
 import com.mongodb.MongoException;
-import com.mongodb.MongoTimeoutException;
 import com.mongodb.ServerAddress;
 
 /**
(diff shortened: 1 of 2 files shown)


hferentschik:
@emmanuelbernard, is this what you had in mind?

emmanuelbernard:
I've commented in the commit.

hferentschik:
Where do we stand on this one now? I guess I can try to improve a bit on the error handling.

emmanuelbernard:
The version in this PR - 9641063 - does not work as expected as we discussed.
So either:
1. we keep the original code minus the retry logic
2. you rewrite it but with the same semantic: we can think of changing the semantic in 5.0 if we manage find a better one

hferentschik:
@emmanuelbernard, I pushed an update which just changes the timeout in hibernate.properties and just removes the retry logic. I have my doubts about the remaining {{if ( containsDatabase == null ) }} branch  and the whole logic in this method. See also https://hibernate.atlassian.net/browse/OGM-848.

Personally I would re-write the whole method.

gunnarmorling:
> I am wondering how useful the hibernate.ogm.datastore.create_database option is for a production environment. 

Agreed, it mostly makes sense for testing purposes. Eventually, we plan to support the `hibernate.hbm2ddl.auto` option from ORM and remove `hibernate.ogm.datastore.create_database` in favor of that.

gunnarmorling:
Merged; thanks, @hferentschik!

hferentschik:
> Eventually, we plan to support the hibernate.hbm2ddl.auto option from ORM and remove hibernate.ogm.datastore.create_database in favor of that.

+1

PR #557, opened 2015-05-28
OGM-835 Disabling tests in QueriesWithEmbeddedCollectionTest which fa…

hferentschik:
…il for Fongo

Disabling the tests for now. We are most likely hitting a bug/limitation of Fongo. We could try to isolate a test case for Fongo and create an issue there.


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/557

DavideD:
Thanks

PR #559, opened 2015-07-20
Elastic Search First Shot

wokier:
Created the basic stuff.
without associations.
Inspired by the couchdb implementation.
Needs an elastic search instance running, in order to run the tests.
Use ELASTICSEARCH_HOSTNAMES variable to pass server and port.
Reviews and Feedback are welcome.
The pull request contains other commits, by mistake.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -19,7 +19,7 @@
     <artifactId>hibernate-ogm-core</artifactId>
     <packaging>jar</packaging>
 
-    <name>Hibernate Object Grid Mapper</name>
+    <name>Hibernate Object Grid Mapper Core</name>
     <description>Persist data in a grid rather than a relational database</description>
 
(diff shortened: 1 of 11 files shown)


emmanuelbernard:
Sweet! We might be a bit slow to answer but we will definitively have a look at it asap.

emmanuelbernard:
One thing though @wokier, there are lots of code style changes on existing classes. Could you try to keep these changes in separated commits.
In this case they won't even pass our checkstyle rules, you need to run `mvn clean install` and see how it crashes in flames :) The checkstyle rules are under `src` somewhere.
I'm not sure of your kunfu level on git but we can try and help you to undo some of these unwanted changes.

wokier:
I made some cleanup and created a new pull request #561

PR #560, opened 2015-07-22
Contribution for a Redis module

mp911de:
Contains support for JSON-based value and association mapping, TTL support, in-entity and external association storage, no transaction support, no query support (Redis has no query language, just Lua scripting), Wildfly/EAP support,  initial docs

Test results:
```
Tests run: 264, Failures: 4, Errors: 2, Skipped: 17
```

Signed the CLA for Hibernate right now. Have fun and let's start a discussion how we can bring this into Hibernate OGM

Known issues with this PR:
* Some tests fail (need your help to figure out, why this is)
* Version conflict with Google's Guava. Cassandra specifies a 14ish version, lettuce needs 17 (did not check for compatibility yet)


redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
```diff
@@ -0,0 +1,546 @@
...
+	}
+
+	@Override
+	public Tuple getTuple(EntityKey key, TupleContext tupleContext) {
+
+		Entity entity = getEntity( key, tupleContext.getOptionsContext() );
+		if ( entity != null ) {
+			return new Tuple( new RedisTupleSnapshot( entity.getProperties() ) );
+		}
+		else if ( isInTheQueue( key, tupleContext ) ) {
```
gunnarmorling: Are you actually doing some kind of batching? Or is it just a left-over from MongoDB? I am quite unhappy with how the batching is done in the MongoDB backend, so I'd prefer if we could start for Redis without this.

mp911de: This code is c'n'p from MongoDB because I assumed it was needed. I can remove the queueing part. Can you explain to me why you don't like how it's done?

redis/src/main/java/org/hibernate/ogm/datastore/redis/options/impl/AssociationStorageOption.java
```diff
@@ -0,0 +1,30 @@
...
+import org.hibernate.ogm.options.spi.UniqueOption;
+import org.hibernate.ogm.util.configurationreader.spi.ConfigurationPropertyReader;
+
+/**
+ * Represents the type of association storage as configured via the API or annotations for a given element.
+ *
+ * @author Gunnar Morling
+ * @author Mark Paluch
+ */
+public class AssociationStorageOption extends UniqueOption<AssociationStorageType> {
```
gunnarmorling: Could `org.hibernate.ogm.datastore.document.options.spi.AssociationStorageOption` be re-used instead?

gunnarmorling: Could `org.hibernate.ogm.datastore.document.options.spi.AssociationStorageOption` be re-used instead?

redis/src/main/java/org/hibernate/ogm/datastore/redis/options/impl/TTLOption.java
```diff
@@ -0,0 +1,28 @@
...
+import org.hibernate.ogm.options.spi.UniqueOption;
+import org.hibernate.ogm.util.configurationreader.spi.ConfigurationPropertyReader;
+
+/**
+ * Option for specifying the <a href="http://redis.io/commands/pexpireat">expiry</a> in
+ * Redis.
+ *
+ * @author Mark Paluch
+ */
+public class TTLOption extends UniqueOption<Long> {
```
gunnarmorling: Very nice you already support that :)

redis/src/main/java/org/hibernate/ogm/datastore/redis/options/navigation/RedisGlobalContext.java
```diff
@@ -0,0 +1,19 @@
...
+package org.hibernate.ogm.datastore.redis.options.navigation;
+
+import org.hibernate.ogm.datastore.keyvalue.options.navigation.KeyValueStoreGlobalContext;
+import org.hibernate.ogm.datastore.redis.navigation.RedisEntityContext;
+
+/**
+ * Allows to configure Redis-specific options applying on a global level. These options may be overridden for single
+ * entities or properties.
+ *
+ * @author Gunnar Morling
```
gunnarmorling: I didn't even know :) Better add your own name.

redis/src/main/java/org/hibernate/ogm/datastore/redis/options/impl/AssociationStorageOption.java
mp911de: Ok from my side but that's up to you. `AssociationStorageOption` has a `document` in its package name, was not sure about that and wanted to see, what's your opinion about that. Will change it.

redis/src/main/java/org/hibernate/ogm/datastore/redis/options/navigation/RedisGlobalContext.java
mp911de: Will do.

core/src/test/java/org/hibernate/ogm/backendtck/associations/storageconfiguration/AssociationStorageConfiguredProgrammaticallyTest.java
```diff
@@ -31,7 +31,7 @@
  * @author Gunnar Morling
  */
 @SkipByGridDialect(
-		value = { GridDialectType.EHCACHE, GridDialectType.HASHMAP, GridDialectType.INFINISPAN, GridDialectType.NEO4J, GridDialectType.CASSANDRA },
+		value = { GridDialectType.EHCACHE, GridDialectType.HASHMAP, GridDialectType.INFINISPAN, GridDialectType.NEO4J, GridDialectType.CASSANDRA, GridDialectType.REDIS },
```
gunnarmorling: Why is it excluded for Redis? The "association storage" option seems to be supported?

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
```diff
@@ -0,0 +1,1946 @@
+[[ogm-redis]]
```
gunnarmorling: Wow, you even added docs already. That's fantastic!

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
```diff
@@ -0,0 +1,1946 @@
...
+
+* Storing entities as JSON documents
+* Storing entities as a bunch of String-based key-value pairs
+
+[NOTE]
+====
+Support for Redis is considered an EXPERIMENTAL feature as of this release.
+In particular you should be prepared for possible changes to the persistent representation of mapped objects in future releases.
+
+Also be aware of the fact that partial updates are unsupported at the moment
```
gunnarmorling: Partial updates a pro for the hash strategy I guess.

core/src/test/java/org/hibernate/ogm/backendtck/associations/storageconfiguration/AssociationStorageConfiguredProgrammaticallyTest.java
mp911de: 2 reasons: I do now know, whether I have to implement sth to support programatically config and the test fails with a class cast exception: 

```
org.hibernate.ogm.datastore.redis.options.navigation.impl.RedisGlobalContextImpl_$$_jvst662_2 cannot be cast to org.hibernate.ogm.datastore.document.options.navigation.DocumentStoreGlobalContext
```

I can see I if can fix the `ClassCastException`by using a different GlobalContext type.

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
mp911de: Yeah, but I didn't take removals/set to null operations into account yet. My bad.

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
mp911de: Need your advice on the representation of keys. I gave it a try, but I'm not really happy with the outcome. Don't have a better idea until now.

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
```diff
@@ -0,0 +1,1946 @@
...
+    private String description;
+
+    //getters, setters ...
+}
+----
+
+[source, JSON]
+----
+{
+    "version": "1",
```
gunnarmorling: "1" or 1?

mp911de: 1

Will change that. Found also some mismatches in the CouchDB doc, that's the base for the Redis docs. I'll contribute the CouchDB fixes later on.

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
```diff
@@ -0,0 +1,1946 @@
...
+
+===== Identifier generation strategies
+
+You can assign id values yourself or let Hibernate OGM generate the value using the
+[classname]`@GeneratedValue` annotation.
+
+Two main strategies are supported:
+
+1. <<redis-table-id-generation-strategy, TABLE>>
+2. <<redis-sequence-id-generation-strategy, SEQUENCE>>
```
gunnarmorling: Would a sequence more "naturally" mapped to a top-level key instead? It also should help avoiding contention when accessing several sequences in parallel.

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
```diff
@@ -0,0 +1,1946 @@
...
+
+    // getters, setters ...
+}
+----
+
+[source, JSON]
+----
+Key: User:user_001
+{ 
+  "addresses" : [
```
gunnarmorling: I think a more natural mapping would be

    "addresses" : [
        { "work" : "address_001" },
        { "home" : "address_002" }
    ]

But the proposed mapping is the same we use for Mongo/Couch so it's alright. There is an OGM issue somewhere to alter the map mapping for document stores.

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
mp911de: You mean the same style as the entities were pulled up from hashes to keys?

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
```diff
@@ -0,0 +1,1946 @@
...
+We recommend that you still use transaction demarcations with Hibernate OGM
+to trigger the flush operation transparently (on commit).
+But do not consider rollback as a possibility, this won't work.
+
+=== Queries
+
+Hibernate OGM is a work in progress
+and we are actively working on JP-QL query support.
+
+In the mean time, you have two strategies to query entities stored by Hibernate OGM:
```
gunnarmorling: What's the other one :)

redis/src/main/java/org/hibernate/ogm/datastore/redis/navigation/RedisEntityContext.java
```diff
@@ -0,0 +1,19 @@
...
+import org.hibernate.ogm.datastore.keyvalue.options.navigation.KeyValueStoreEntityContext;
+import org.hibernate.ogm.datastore.redis.options.navigation.RedisPropertyContext;
+
+/**
+ * Allows to configure Redis-specific options applying on a global level. These options may be overridden for single
+ * properties.
+ *
+ * @author Gunnar Morling
+ */
+public interface RedisEntityContext extends KeyValueStoreEntityContext<RedisEntityContext, RedisPropertyContext> {
```
gunnarmorling: Should be in the same package as global and property context.

redis/src/main/java/org/hibernate/ogm/datastore/redis/options/TTL.java
```diff
@@ -0,0 +1,50 @@
...
+	/**
+	 * Specifies the expiry duration after persisting the annotated entity or
+	 * property.
+	 *
+	 * @return the expiry duration.
+	 */
+	long value();
+
+	/**
+	 * Speciefies the unit of duration for expiring the key. Defaults to {@link java.util.concurrent.TimeUnit#SECONDS}
```
gunnarmorling: "Specifies"

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
```diff
@@ -0,0 +1,546 @@
...
+ */
+public class RedisDialect extends BaseGridDialect {
+
+	public static final String IDENTIFIERS = "Identifiers";
+	public static final byte[] IDENTIFIERS_AS_BYTES = toBytes( IDENTIFIERS );
+	public static final String ASSOCIATIONS = "Associations";
+
+	protected final Map<EntityStorageType, EntityStorageStrategy> entityStorageStrategies;
+
+	private final RedisDatastoreProvider provider;
```
gunnarmorling: Seems not used.

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
```diff
@@ -0,0 +1,546 @@
...
+ *
+ * @author Mark Paluch
+ */
+public class RedisDialect extends BaseGridDialect {
+
+	public static final String IDENTIFIERS = "Identifiers";
+	public static final byte[] IDENTIFIERS_AS_BYTES = toBytes( IDENTIFIERS );
+	public static final String ASSOCIATIONS = "Associations";
+
+	protected final Map<EntityStorageType, EntityStorageStrategy> entityStorageStrategies;
```
gunnarmorling: Can you put the `@Immutable` annotation for documentation purposes? Dialects are accessed concurrently, so they must be thread-safe. I've started to use that annotation to document fields save for this despite their declared type not telling this on first sight.

redis/src/main/java/org/hibernate/ogm/datastore/redis/dialect/model/impl/DocumentBasedAssociation.java
```diff
@@ -0,0 +1,42 @@
...
+import org.hibernate.ogm.datastore.redis.dialect.value.Association;
+import org.hibernate.ogm.datastore.redis.dialect.value.StructuredValue;
+
+
+/**
+ * A {@link RedisAssociation} backed by an {@link Association}.
+ *
+ * @author Gunnar Morling
+ */
+class DocumentBasedAssociation extends RedisAssociation {
```
gunnarmorling: The classes in this package look very similar to what we have for CouchDB. Do you think this could be extracted into shared classes living in core?

redis/src/main/java/org/hibernate/ogm/datastore/redis/dialect/model/impl/RedisAssociationRowFactory.java
```diff
@@ -0,0 +1,94 @@
...
+			this( null, null );
+		}
+
+		public RedisAssociationRowAccessor(String[] prefixedColumns, String prefix) {
+			this.prefix = prefix;
+			if ( prefix != null ) {
+				this.prefixedColumns = Arrays.asList( prefixedColumns );
+			}
+			else {
+				this.prefixedColumns = new ArrayList<String>( 0 );
```
gunnarmorling: `Collections.emptyList`?

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
```diff
@@ -0,0 +1,546 @@
...
+	@Override
+	public Tuple getTuple(EntityKey key, TupleContext tupleContext) {
+
+		Entity entity = getEntity( key, tupleContext.getOptionsContext() );
+		if ( entity != null ) {
+			return new Tuple( new RedisTupleSnapshot( entity.getProperties() ) );
+		}
+		else if ( isInTheQueue( key, tupleContext ) ) {
+			// The key has not been inserted in the db but it is in the queue
+			return new Tuple( new RedisTupleSnapshot( entity.getProperties() ) );
```
gunnarmorling: `entity` can only be null at this point, so an NPE is due when running this branch.

redis/src/main/java/org/hibernate/ogm/datastore/redis/impl/json/JsonSerializationStrategy.java
```diff
@@ -0,0 +1,64 @@
...
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * Serialize payload as JSON using the Jackson Mapper.
+ *
+ * @author Mark Paluch
+ */
+public class JsonSerializationStrategy implements SerializationStrategy {
```
gunnarmorling: What other strategy impls are there besides this one? I'd only establish an interface once there are actual several impls.

redis/src/main/java/org/hibernate/ogm/datastore/redis/dialect/value/Entity.java
```diff
@@ -0,0 +1,166 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonSerialize.Inclusion;
+
+/**
+ * Contains the information related to a {@link Tuple}. The use of this class is to serialize and deserialize the JSON
+ * stored in Redis; Documents have the following structure:
+ * <p/>
+ * <pre>
+ * {@code
+ * {
+ *     "$type": "entity",
```
gunnarmorling: That one actually is not there, right?

redis/src/main/java/org/hibernate/ogm/datastore/redis/impl/RedisDatastoreProvider.java
```diff
@@ -0,0 +1,119 @@
...
+
+	@Override
+	public Class<? extends GridDialect> getDefaultDialect() {
+		return RedisDialect.class;
+	}
+
+	@Override
+	public void configure(Map configurationValues) {
+
+		OptionsService optionsService = serviceRegistry.getService( OptionsService.class );
```
gunnarmorling: It's not used.

redis/src/main/java/org/hibernate/ogm/datastore/redis/impl/RedisDatastoreProvider.java
```diff
@@ -0,0 +1,119 @@
...
+
+		this.config.initConfiguration( configurationValues );
+	}
+
+
+	@Override
+	public void start() {
+		try {
+			Hosts.HostAndPort hostAndPort = config.getHosts().getFirst();
+			RedisURI.Builder builder = new RedisURI.Builder().redis( hostAndPort.getHost(), hostAndPort.getPort() );
```
gunnarmorling: `redis()` should be accessed as static method.

redis/src/main/java/org/hibernate/ogm/datastore/redis/impl/RedisConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+	private long timeout = DEFAULT_TIMEOUT;
+	private boolean ssl = false;
+
+
+	public RedisConfiguration(ConfigurationPropertyReader propertyReader) {
+		String host = propertyReader.property( OgmProperties.HOST, String.class )
+				.withDefault( DEFAULT_HOST )
+				.getValue();
+
+		Integer port = propertyReader.property( OgmProperties.PORT, Integer.class )
```
gunnarmorling: We deprecated PORT recently (HOST is a (list of) host/port tuple(s) now). We should not add it to new data stores.

redis/src/main/java/org/hibernate/ogm/datastore/redis/impl/RedisConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+	public boolean isSsl() {
+		return ssl;
+	}
+
+	/**
+	 * Initialize the internal values form the given {@link Map}.
+	 *
+	 * @param configurationMap The values to use as configuration
+	 */
+	public void initConfiguration(Map<?, ?> configurationMap) {
```
gunnarmorling: How comes that's not done in the constructor as the others?

redis/src/main/java/org/hibernate/ogm/datastore/redis/impl/RedisConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+				.getValue();
+
+		Integer port = propertyReader.property( OgmProperties.PORT, Integer.class )
+				.withValidator( Validators.PORT )
+				.withDefault( null )
+				.getValue();
+
+		hosts = HostParser.parse( host, port, DEFAULT_PORT );
+
+		databaseNumber = propertyReader.property( OgmProperties.PORT, Integer.class )
```
gunnarmorling: Is it right do initialize that with the port? A redis-specific parameter for this seems better suited.

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
gunnarmorling: What I mean is, your dialect does not implement `BatchableGridDialect`, so items will never be put in to the operations queue.

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
```diff
@@ -0,0 +1,546 @@
...
+			connection.hset( IDENTIFIERS_AS_BYTES, key, toBytes( Long.toString( request.getInitialValue() ) ) );
+			return request.getInitialValue();
+		}
+
+		return connection.hincrby( IDENTIFIERS_AS_BYTES, key, request.getIncrement() );
+	}
+
+	@Override
+	public void forEachTuple(final ModelConsumer consumer, EntityKeyMetadata... entityKeyMetadatas) {
+
```
gunnarmorling: We usually don't do empty lines after the method name. I don't care too much, it's just a uncommon thing which makes your flow stop for a sec when reading the code.

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
```diff
@@ -0,0 +1,546 @@
...
+
+		for ( EntityKeyMetadata entityKeyMetadata : entityKeyMetadatas ) {
+
+			KeyScanCursor<byte[]> cursor = null;
+			String pattern = entityKeyMetadata.getTable() + ":*";
+			do {
+
+				if ( cursor != null ) {
+					cursor = connection.scan( cursor, ScanArgs.Builder.matches( pattern ) );
```
gunnarmorling: Could you pull the pattern matching out of the while loop?

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
```diff
@@ -0,0 +1,546 @@
...
+				}
+				else {
+					cursor = connection.scan( ScanArgs.Builder.matches( pattern ) );
+				}
+
+				for ( byte[] key : cursor.getKeys() ) {
+
+					String type = connection.type( key );
+					EntityStorageStrategy entityStorageStrategy;
+					if ( "hash".equalsIgnoreCase( type ) ) {
```
gunnarmorling: How about `static EntityStorageType forType(String type)` declared on `EntityStorageType`?

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
```diff
@@ -0,0 +1,546 @@
...
+				owningEntity.removeAssociation( key.getMetadata().getCollectionRole() );
+				storeEntity( key.getEntityKey(), owningEntity, associationContext );
+			}
+		}
+		else {
+			removeAssociation( key.getEntityKey() );
+		}
+	}
+
+	public Entity getEntity(EntityKey key, OptionsContext optionsContext) {
```
gunnarmorling: Could be privat.

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
```diff
@@ -0,0 +1,546 @@
...
+		return getEntityStorageStrategy( optionsContext ).getEntity( entityId( key ) );
+	}
+
+	public void storeEntity(EntityKey key, Map<String, Object> map, OptionsContext optionsContext) {
+		Entity entityDocument = new Entity();
+		Set<String> keys = new HashSet<>( Arrays.asList( key.getColumnNames() ) );
+
+
+		for ( Map.Entry<String, Object> entry : map.entrySet() ) {
+			if ( keys.contains( entry.getKey() ) ) {
```
gunnarmorling: There is `key.getMetadata().isKeyColumn()`. Saves the set/list allocation.

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
```diff
@@ -0,0 +1,546 @@
...
+		connection.pexpire( associationId, ttl );
+	}
+
+	public void storeAssociation(EntityKey key, Association document) {
+
+		byte[] associationId = associationId( key );
+		connection.del( associationId );
+
+		for ( Object row : document.getRows() ) {
+			connection.rpush( associationId, serializationStrategy.serialize( row ) );
```
gunnarmorling: So this does not use JSON? Or am I missing something. Also I think it'd make sense to use the "transaction" semantics to write all rows of one association in one go?

redis/src/main/java/org/hibernate/ogm/datastore/redis/impl/RedisConfiguration.java
mp911de: That's a bug

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
mp911de: It is JSON but within a Redis List (see `serializationStrategy.serialize` with is JSON for now). I noticed that the CouchDB association is just a List being serialized as JSON. I deconstructed that association so it fits nicely into a Redis List.

We should discuss the transaction issue on a more global level, but yes, wrapping it into a Redis transaction (`MULTI`/`EXEC`) is feasible.

gunnarmorling: Ah no, there is the call to `serialize()` :) Still the question remains whether it could be done within "all or nothing" boundaries.

redis/src/main/java/org/hibernate/ogm/datastore/redis/dialect/value/Entity.java
mp911de: Leftover, will clean it up.

redis/src/main/java/org/hibernate/ogm/datastore/redis/impl/json/JsonSerializationStrategy.java
mp911de: There are a few more serializations that are used more or less often: [Kryo](https://github.com/EsotericSoftware/kryo), [Msgpack](http://msgpack.org/) and [Java](http://docs.oracle.com/javase/7/docs/api/java/io/ObjectOutputStream.html). Didn't want to blow up the initial PR.

redis/src/main/java/org/hibernate/ogm/datastore/redis/dialect/model/impl/DocumentBasedAssociation.java
mp911de: I'll have to take a deeper look.

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
mp911de: How about discussing the transaction issue on IRC and putting the results into a PR comment?

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisProperties.java
```diff
@@ -0,0 +1,51 @@
...
+	public static final String ENTITY_STORE = "hibernate.ogm.datastore.redis.entity_storage";
+
+	/**
+	 * Configuration property for setting the expiry of keys. This value is set in milliseconds.
+	 * <p/>
+	 * Defaults to none.
+	 *
+	 * @see org.hibernate.ogm.datastore.redis.options.TTL
+	 */
+	public static final String EXPIRY = "hibernate.ogm.redis.expiry";
```
gunnarmorling: I'd name it "TTL" everywhere.

gunnarmorling:
@mp911de, done first the first round of review. It looks very good overall,
it's really on a great level for an initial contribution. That's much
appreciated!

The comments I have so far are mostly minor things. The redundancy with the
CouchDB dialect concerns me a bit, it'd be great if we could further
extract common stuff. The other thing is using the MULTI/EXEC approach for
stuff that belongs together (rows of one association, embedded associations
and their "owning" entity).

redis/src/main/java/org/hibernate/ogm/datastore/redis/options/impl/AssociationStorageOption.java
gunnarmorling: I think it makes sense to re-use it. With the JSON mapping strategy, Redis feels rather "document-ish" to me.

core/src/test/java/org/hibernate/ogm/backendtck/associations/storageconfiguration/AssociationStorageConfiguredProgrammaticallyTest.java
gunnarmorling: You'd have to implement `DocumentStoreGlobalContext`. Check out how it's done in the other dialects, I don't think you need to do anything else to make it work then.

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
gunnarmorling: Ok, thanks a lot for that!

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
gunnarmorling: Yes. Seems preferrable, WDYT?

mp911de: Sure, just wanted to make sure that I understand the comment in the right way.

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
mp911de: Result of our discussion: We'll start without transactions in the first step. Transactions require multiple connections/connection pooling. Pooling and transactions will be done in a latter step.

redis/src/main/java/org/hibernate/ogm/datastore/redis/dialect/model/impl/DocumentBasedAssociation.java
mp911de: Where should I put the classes? In `org.hibernate.ogm.datastore.document.association.spi` and `org.hibernate.ogm.datastore.association.spi`?

What about the name? Would go for `AssociationRepresentation`, `EmbeddedAssociationRepresentation` and `DocumentBasedAssociationRepresentation`, each typed with `T` for `T getOwningDocument()`.

But to be honest, the whole construction of the `RedisAssociation` and `CouchDBAssociation` is sort of strange. It's just a bridge between `AssociationRows` and the representation. We should discuss that topic, perhaps we can build something just on top of `AssociationRows`.

mp911de:
Pushed the updated coded according to the feedback.

mp911de:
Fixed missing entity key when loading entities. The only failing test is `CompensationSpiTest`, the reason seems the not supported locking strategies.

gunnarmorling:
@mp911de Great work! I will take another look at the latest changes tomorrow. In between, could you:

* Open an OGM issue and prefix all the commit messages with OGM-123 XYZ...
* Add an integration test (similar to e.g. `EhcacheModuleMemberRegistrationIT`)

? Thanks!

mp911de:
I would reuse https://hibernate.atlassian.net/browse/OGM-194 and extend the description by the features the Redis module supports. Is this OK with you?

gunnarmorling:
Ah yes, sure. That's perfect. Wasn't aware of that issue.

modules/eap/pom.xml
```diff
@@ -65,6 +65,10 @@
             <artifactId>hibernate-ogm-couchdb</artifactId>
```
gunnarmorling: The EAP module ZIP has been removed on master in parallel (it's severely outdated at this point and EAP 6 will not work with ORM 5). Could you thus exclude these changes from the PR? Thanks!

mp911de:
Updated all commits (push -f ugh!) and merged master into redis branch to resolve pom.xml conflict. Will provide the arquillian integration test.

gunnarmorling:
Rebased and applied. Added some commits to clean up some left-overs. Many thanks to you, @mp911de! I am very glad to see Redis support in OGM, looking forward to see that dialect grow and prosper over time :)

itamarhaber:
@mp911de @gunnarmorling kudos - an amazing contrib! :beers:

PR #561, opened 2015-07-23
Elasticsearch

wokier:


core/src/main/java/org/hibernate/ogm/dialect/spi/BaseGridDialect.java
```diff
@@ -44,4 +46,19 @@ public boolean supportsSequences() {
...
+		for (EntityKeyMetadata entityKeyMetadata : entityKeyMetadatas) {
+			List<Tuple> tuples = getTuples(entityKeyMetadata);
+			for (Tuple tuple : tuples) {
+				consumer.consume(tuple);
+			}
+		}
+	}
+
+	// Should be abstract, but high impact on all subclasses
+	protected List<Tuple> getTuples(EntityKeyMetadata entityKeyMetadata) {
```
Sanne: Do you think it's a good idea to encourage new developers to use this approach?

The method should at least be abstract, but then even so the fact that it returns a list of tuples tends to encourage loading all tuples in the JVM heap, which is quite prone to out of memory errors and also very inefficient.

I think you can use this approach on the ElasticSearch dialect while it's work in progress, but I wouldn't encourage this pattern in the parent `GridDialect`.

(besides you're also introducing some checkstyle errors by reformatting unrelated code areas, so better revert all changes to this class)

couchdb/src/main/java/org/hibernate/ogm/datastore/couchdb/CouchDBDialect.java
```diff
@@ -71,9 +71,9 @@ public CouchDBDialect(CouchDBDatastoreProvider provider) {
 
 	@Override
 	public Tuple getTuple(EntityKey key, TupleContext tupleContext) {
-		EntityDocument entity = getDataStore().getEntity( Identifier.createEntityId( key ) );
-		if ( entity != null ) {
-			return new Tuple( new CouchDBTupleSnapshot( entity.getProperties() ) );
+		EntityDocument entity = getDataStore().getEntity(Identifier.createEntityId(key));
```
Sanne: these changes don't seem to be useful, right?

elasticsearch/pom.xml
```diff
@@ -0,0 +1,143 @@
...
+
+	<name>Hibernate OGM for Elastic search as a Database</name>
+	<description>Persist objects in Elastic search</description>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.hibernate.ogm</groupId>
+			<artifactId>hibernate-ogm-core</artifactId>
+			<exclusions>
+				<exclusion>
```
Sanne: Could you add a comment here to explain why the exclusion is needed?

core/src/main/java/org/hibernate/ogm/dialect/spi/BaseGridDialect.java
emmanuelbernard: Note that we are introducing a different `getTuples` method as part of a different improvement. Our version receives a list of `EntityKey` to load.

couchdb/src/main/java/org/hibernate/ogm/datastore/couchdb/CouchDBDialect.java
emmanuelbernard: @wokier we are a bit weird in that way but we put spaces between parenthesis and what's inside parenthesis (except for method definition, catch and for I think). run mvn checkstyle:checkstyle, you will see these rules being raised.

core/src/main/java/org/hibernate/ogm/dialect/spi/BaseGridDialect.java
gunnarmorling: We should not have this method on `BaseDialect`.

elasticsearch/src/main/java/org/hibernate/ogm/datastore/elasticsearch/ElasticSearchDialect.java
```diff
@@ -0,0 +1,195 @@
...
+	@Override
+	public boolean isStoredInEntityStructure(AssociationKeyMetadata associationKeyMetadata, AssociationTypeContext associationTypeContext) {
+		return false;
+	}
+
+	@Override
+	public Number nextValue(NextValueRequest request) {
+		String sequencesIndex = databaseSequencesIndex();
+		if (!sequences.contains(request.getKey().getTable())) {
+			final Map<String, Object> disabled = new HashMap<String, Object>() {
```
gunnarmorling: I would prefer to not use this pattern, as it's creating anonymous classes. In this case, also just `Collections#singletonMap()` can be used.

couchdb/src/main/java/org/hibernate/ogm/datastore/couchdb/CouchDBDialect.java
```diff
@@ -55,9 +55,9 @@
 /**
```
gunnarmorling: Why is the `CouchDBDialect` modified at all? Is there any change besides formatting?

elasticsearch/src/main/java/org/hibernate/ogm/datastore/elasticsearch/ElasticSearchDialect.java
```diff
@@ -0,0 +1,195 @@
...
+	public Tuple getTuple(EntityKey key, TupleContext tupleContext) {
+		GetResponse getResponse = provider.getClient().prepareGet(provider.getDatabase(), key.getTable(), id(key)).setOperationThreaded(false).execute().actionGet();
+		if (getResponse != null) {
+			return new Tuple(new MapTupleSnapshot(getResponse.getSource()));
+		}
+		return null;
+	}
+
+	private String id(EntityKey key) {
+		return String.valueOf(key.getColumnValues()[0]);
```
gunnarmorling: There could be more than one key column (when using a composite id type). You could check out the CouchDB backend for instance, there we also have this issue and address it by concatenating the different column values to one String.

core/src/main/java/org/hibernate/ogm/dialect/spi/BaseGridDialect.java
wokier: ok, go it, well explained. thanks

elasticsearch/src/main/java/org/hibernate/ogm/datastore/elasticsearch/ElasticSearchDialect.java
wokier: In fact, this part i did'nt well understood. Does the Entity Key contains all the key/values of the object, or only the first one, or only those marked as id by annotation ?

gunnarmorling: It contains those column(s) representing the id property of the entity (annotated with `@Id`). There will be more than one column in case of a composite id type (see `@EmbeddedId`).

DavideD:
A bit out of date now, I'm closing this

PR #563, opened 2015-07-23
OGM-703 Upgrade to Neo4j 2.2.3

DavideD:
https://hibernate.atlassian.net/browse/OGM-703

This is the pull request for master.

This should work now, I've also added an integration test for WildFLy that uses JTA and it seems to work.

The main differences for our code are:
* dependency updates: 
* Query are executed using the GraphDb instead of a different object
* Neo4j now has a Neo4jTransactionFactory and Neo4jTransactionImplementor that will delegate all the operations to the configure TransactionFactory by starting/committing transaction with Neo4j.
* I've refactored some tests because now that we don't use a JtaPlatform the transaction wasn't synchronized anymore when a non JTA persistence unit was used.
* I've removed the adapter to the StringLogger class, the new method used to run query does not have a way to pass it and it seems it's not showing all the additional log we had in past


gurkerl83:
Thanks for doing this! Unfortunately I`m not able to test the new transaction approach until before 1th August. Right now we are using the latest version with few custom made modifications of hibernate-ogm able to run it + dependencies within the OSGi Container Karaf.

DavideD:
> I wonder, should you check the JTA transaction status to make sure it is not marked for rollback already somehow?

I added a commit to the pull request

emmanuelbernard:
BTW I did not see any doc change. Is there a need for some?

DavideD:
https://github.com/DavideD/hibernate-ogm/commit/3ca2e3015bd6742fbbe7e5fc7643e3137e5ad715

DavideD:
The commit with the doc update: https://github.com/DavideD/hibernate-ogm/commit/3ca2e3015bd6742fbbe7e5fc7643e3137e5ad715

emmanuelbernard:
Review done, except he doc improvement, I think that's good to go @DavideD . Update the doc and push.

DavideD:
@emmanuelbernard 
I've update the documentation on this commit:
https://github.com/DavideD/hibernate-ogm/commit/f696895379a8f30dc72e20b9d5b0c6cc8352b703

I've added the check for the rollback on this commit:
https://github.com/DavideD/hibernate-ogm/commit/7960e9b277a9df1ee7de31cd45787eb56f4db70c

DavideD:
Added accommit to make sure transaction are closed when expected

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/563

documentation/manual/src/main/asciidoc/en-US/modules/neo4j.asciidoc
```diff
@@ -784,6 +755,22 @@ tx.commit();
...
+That way when the JTA transaction is committed or rollbacked (for example by an EJB CMT or
+explicitly), the Neo4J transaction is also committed or rollbacked.
+This makes for a very nice integration in a Java EE container.
+
+[CAUTION]
+----
+This is NOT a true JTA/XA integration but more a lifecycle alignment:
+changes on more than one datasource won't be executed as a single atomic transaction.
+
+In particular, if the JTA transaction involves multiple resources, Neo4j might commit
```
gunnarmorling: Couldn't we make sure somehow that Neo4j is committed last, something like the "[last resource gambit](https://access.redhat.com/documentation/en-US/JBoss_Enterprise_Application_Platform/5/html/Administration_And_Configuration_Guide/lrco-overview.html)"?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/transaction/impl/Neo4jTransactionImplementor.java
```diff
@@ -0,0 +1,153 @@
...
+import org.hibernate.ogm.datastore.neo4j.impl.Neo4jDatastoreProvider;
+import org.hibernate.ogm.datastore.neo4j.logging.impl.Log;
+import org.hibernate.ogm.datastore.neo4j.logging.impl.LoggerFactory;
+import org.hibernate.ogm.transaction.impl.ForwardingTransactionImplementor;
+import org.neo4j.graphdb.Transaction;
+
+/**
+ * @author Davide D'Alto
+ */
+public class Neo4jTransactionImplementor extends ForwardingTransactionImplementor {
```
gunnarmorling: Could you add a comment about the motivation for having this class? The reader should know what's special about it.

DavideD: Yes, forgot about it

documentation/manual/src/main/asciidoc/en-US/modules/neo4j.asciidoc
DavideD: I don't think it's possible with the current approach.

PR #564, opened 2015-07-24
OGM-767 Add support for @BatchSize in OGM

emmanuelbernard:
Only for entities for now. Will come in a different issue and PR for later:

* collection batching
* subselect fetching

The "legacy" and default batching strategy is not implemented yet as it requires yet another big pile of code duplication and is almost like the padded one. I'm tempted to not implement it and wait for the batch fetching framework improvements to be moved back to ORM to benefit them.
Also no dialect currently make use of `MultigetGridDialect` but that is hopefully easier to do that the core enablement of this PR.

95% because of the legacy batching strategy. Hence feedback season is opened.

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/dialect/impl/GridDialectLogger.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/impl/GridDialectLogger.java
@@ -7,6 +7,8 @@
 package org.hibernate.ogm.dialect.impl;
 
 import java.io.Serializable;
+import java.util.Arrays;
+import java.util.List;
 
 import org.hibernate.engine.spi.QueryParameters;
 import org.hibernate.ogm.dialect.batch.spi.OperationsQueue;
@@ -56,6 +58,14 @@ public class GridDialectLogger extends ForwardingGridDialect<Serializable> {
 	public Tuple getTuple(EntityKey key, TupleContext tupleContext) {
 		log.tracef( "Reading tuple with key %1$s and context %2$s", key, tupleContext );
 		return super.getTuple( key, tupleContext );
+	}
+
+	@Override
+	public List<Tuple> getTuples(EntityKey[] keys, TupleContext tupleContext) {
+		if ( log.isTraceEnabled() ) {
+			log.tracef( "Reading tuples with keys %1$s and context %2$s", Arrays.toString( keys ), tupleContext );
+		}
+		return super.getTuples( keys, tupleContext );
 	}
 
 	@Override
--- a/core/src/main/java/org/hibernate/ogm/loader/impl/OgmLoader.java
+++ b/core/src/main/java/org/hibernate/ogm/loader/impl/OgmLoader.java
@@ -31,6 +31,7 @@ import org.hibernate.event.spi.PreLoadEvent;
 import org.hibernate.internal.util.collections.ArrayHelper;
 import org.hibernate.loader.CollectionAliases;
 import org.hibernate.loader.entity.UniqueEntityLoader;
+import org.hibernate.ogm.dialect.multiget.spi.MultigetGridDialect;
 import org.hibernate.ogm.dialect.spi.GridDialect;
 import org.hibernate.ogm.entityentry.impl.OgmEntityEntryState;
 import org.hibernate.ogm.jdbc.impl.TupleAsMapResultSet;
@@ -53,6 +54,7 @@ import org.hibernate.persister.entity.Loadable;
 import org.hibernate.persister.entity.UniqueKeyLoadable;
 import org.hibernate.pretty.MessageHelper;
 import org.hibernate.proxy.HibernateProxy;
+import org.hibernate.service.spi.ServiceRegistryImplementor;
 import org.hibernate.type.AssociationType;
 import org.hibernate.type.EntityType;
 import org.hibernate.type.Type;
@@ -72,6 +74,7 @@ public class OgmLoader implements UniqueEntityLoader, BatchableEntityLoader {
 	private final LockMode[] defaultLockModes;
 	private final CollectionAliases[] collectionAliases;
 	private final GridDialect gridDialect;
+	private final MultigetGridDialect multigetGridDialect;
 	private final int batchSize;
 
 	/**
@@ -86,7 +89,9 @@ public class OgmLoader implements UniqueEntityLoader, BatchableEntityLoader {
 		this.entityPersisters = new OgmEntityPersister[] {};
 		this.collectionPersisters = collectionPersisters;
 		this.factory = collectionPersisters[0].getFactory();
-		this.gridDialect = this.factory.getServiceRegistry().getService( GridDialect.class );
+		ServiceRegistryImplementor serviceRegistry = this.factory.getServiceRegistry();
+		this.gridDialect = serviceRegistry.getService( GridDialect.class );
+		this.multigetGridDialect = serviceRegistry.getService( MultigetGridDialect.class );
 
 		//NONE, because its the requested lock mode, not the actual!
 		final int fromSize = 1;
@@ -111,7 +116,9 @@ public class OgmLoader implements UniqueEntityLoader, BatchableEntityLoader {
 		this.entityPersisters = entityPersisters;
 		this.collectionPersisters = new OgmCollectionPersister[] {};
 		this.factory = entityPersisters[0].getFactory();
-		this.gridDialect = this.factory.getServiceRegistry().getService( GridDialect.class );
+		ServiceRegistryImplementor serviceRegistry = this.factory.getServiceRegistry();
+		this.gridDialect = serviceRegistry.getService( GridDialect.class );
+		this.multigetGridDialect = serviceRegistry.getService( MultigetGridDialect.class );
 
 		// NONE, because its the requested lock mode, not the actual!
 		final int fromSize = 1;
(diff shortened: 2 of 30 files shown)


gunnarmorling:
Will you send another PR for actually implementing `MultigetDialect` e.g. with MongoDB?

core/src/main/java/org/hibernate/ogm/loader/impl/OgmLoader.java
```diff
@@ -488,10 +551,45 @@ private TupleAsMapResultSet getResultSet(Serializable id, OgmLoadingContext ogmL
...
+					for ( Tuple tuple : multigetGridDialect.getTuples( keys, persister.getTupleContext() ) ) {
+						if ( tuple != null ) {
+							resultset.addTuple( tuple );
+						}
+					}
+				}
+				else {
+					for ( EntityKey entityKey : keys ) {
+						// TODO use the multiload API once introduced
+						Tuple entry = gridDialect.getTuple( entityKey, persister.getTupleContext() );
```
gunnarmorling: Should be doable now, right? I don't see a call to `getTuples()` in the first test method.

core/src/main/java/org/hibernate/ogm/loader/impl/OgmLoader.java
```diff
@@ -360,6 +419,10 @@ else if ( ogmLoadingContext.hasResultSet() ) {
 		return results;
 	}
 
+	private boolean useBatching(QueryParameters qp) {
+		return qp.getPositionalParameterValues().length > 1;
```
gunnarmorling: Could you explain this implementation? How do parameters play into this?

core/src/test/java/org/hibernate/ogm/test/batchfetching/BatchFetchingTest.java
```diff
@@ -0,0 +1,115 @@
...
+		statistics.clear();
+
+		assertEquals( 0, statistics.getEntityStatistics( Floor.class.getName() ).getFetchCount() );
+		for ( Floor currentFloor : tower.getFloors() ) {
+			// load proxies
+			Object entity = session.load( Floor.class, currentFloor.getId() );
+			Hibernate.initialize( entity );
+			assertTrue( Hibernate.isInitialized( entity ) );
+		}
+		assertEquals( 1, statistics.getEntityStatistics( Floor.class.getName() ).getFetchCount() );
```
gunnarmorling: Apparently, still `getTuple()` is used underneath atm. How comes it still reports 1 as fetch count. I am wondering whether this test set-up is based on a wrong assumption?

core/src/test/java/org/hibernate/ogm/test/batchfetching/package-info.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+
+/**
+ * Package containing tests related to batchfetching entity loading.
```
gunnarmorling: Very thorough ;)

core/src/main/java/org/hibernate/ogm/loader/impl/OgmLoader.java
```diff
@@ -488,10 +551,45 @@ private TupleAsMapResultSet getResultSet(Serializable id, OgmLoadingContext ogmL
...
 		if ( getEntityPersisters().length > 0 ) {
 			OgmEntityPersister persister = getEntityPersisters()[0];
-			final EntityKey key = EntityKeyBuilder.fromPersister( persister, id, session );
-			Tuple entry = gridDialect.getTuple( key, persister.getTupleContext() );
-			if ( entry != null ) {
-				resultset.addTuple( entry );
+			if ( useBatching( qp ) ) {
+				// here we expect to receive QueryParameters.positionalParameters full of ids and thus of the same type.
+				// if that's not the case, we are in a bit of a trouble :)
+				// TODO Adding a safeguard but we pay the price for each call, should we remove it?
```
gunnarmorling: In which case the parameter type would be a wrong type? Seems like its a programming error which we should detect during testing/development, so it's not needed to be asserted continuously at runtime?

core/src/test/java/org/hibernate/ogm/test/batchfetching/BatchFetchingTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+		cleanDataset( session, tower );
+		session.close();
+
+	}
+
+	@Test
+	public void testLoadSeveralFloorsFromTower() throws Exception {
+
+		Log log = LoggerFactory.make();
```
gunnarmorling: Not used.

core/src/main/java/org/hibernate/ogm/loader/impl/OgmLoader.java
```diff
@@ -1198,4 +1296,34 @@ else if ( isSpecialOneToOne ) {*/
...
+			qp.setPositionalParameterTypes( types );
+			qp.setPositionalParameterValues( ids );
+			qp.setOptionalObject( optionalObject );
+			qp.setOptionalEntityName( optionalEntityName );
+			qp.setOptionalId( optionalId );
+			qp.setLockOptions( lockOptions );
+			result = doQueryAndInitializeNonLazyCollections( session, qp, OgmLoadingContext.EMPTY_CONTEXT, false );
+		}
+		catch ( Exception e ) {
+			// TODO should we extract from the dialect the query in a String representation (JDBC logs the SQL)?
```
gunnarmorling: I would expect the original dialect exception to contain the required context?

core/src/main/java/org/hibernate/ogm/loader/impl/OgmLoader.java
gunnarmorling: Ok, got it. The method name made me stumble, I thought it's about whether batching is enabled at all or not. Maybe `containsSeveralIds()` or so?

core/src/main/java/org/hibernate/ogm/loader/impl/OgmLoader.java
```diff
@@ -333,6 +360,38 @@ else if ( ogmLoadingContext.hasResultSet() ) {
 		//TODO should we collect List<Object> as result? Not necessary today
 		Object result = null;
 		List<Object> results = new ArrayList<Object>();
+
+		if ( isCollectionLoader ) {
+			// Logic to eliminate the n+1 issue in collection loading when batc fetching is enabled.
```
gunnarmorling: Ok, looks a bit tough :) There is no more explicit way to add the referenced elements to the batch fetching queue?

gunnarmorling:
@emmanuelbernard: Done with the review. My main concern are the remaining to-dos:

* Pushing duplicated stuff back to ORM
* Remaining to-dos in OGM; Could you open OGM JIRA issues for those remaining open after this PR?

What is the "legacy" stuff about? Didn't fully grok that. I'd vote for doing as less as possible on the OGM side and push back the required changes to ORM.

core/src/main/java/org/hibernate/ogm/dialect/impl/GridDialectLogger.java
```diff
@@ -59,6 +61,14 @@ public Tuple getTuple(EntityKey key, TupleContext tupleContext) {
 	}
 
 	@Override
+	public List<Tuple> getTuples(EntityKey[] keys, TupleContext tupleContext) {
+		if ( log.isTraceEnabled() ) {
+			log.tracef( "Reading tuple with keys %1$s and context %2$s", Arrays.toString( keys ), tupleContext );
```
gunnarmorling: "tuples"

emmanuelbernard:
> Will you send another PR for actually implementing MultigetDialect e.g. with MongoDB?

I was planning on other people to give a hand :)

core/src/main/java/org/hibernate/ogm/loader/impl/OgmLoader.java
emmanuelbernard: yes, removing.

core/src/main/java/org/hibernate/ogm/loader/impl/OgmLoader.java
emmanuelbernard: ok, updated.

core/src/test/java/org/hibernate/ogm/test/batchfetching/BatchFetchingTest.java
emmanuelbernard: It tests that the actual batch framework works. The fetch vs load logic is captured much higher at the event logic level AFAIU.

What I am planning to do to align these two is to force the batch size to 1 int he banch framework call if the underlying dialect does not support `MultigetGridDialect`. It's one of the TODOs I've captured in the issue.

core/src/main/java/org/hibernate/ogm/loader/impl/OgmLoader.java
emmanuelbernard: In the ORM case, the query loading the batch of entities might have additional parameters like a discriminator or a @Where clause restriction I think. That's why the `QueryParameters` allows for heterogenous types. In OGM cases at least for now we don't make use of the heterogeneity.

I don't mind removing the check.

core/src/main/java/org/hibernate/ogm/loader/impl/OgmLoader.java
emmanuelbernard: Method extracted.

emmanuelbernard:
@gunnarmorling I've fixed a lot of the remarks and todos form OGM-767. You can review the major changes starting at 91adb76

PR #566, opened 2015-07-28
OGM-866 Isolate internal Neo4J dependencies from conflicting with use…

Sanne:
…r application dependencies

hackaton result - see the reasoning here:
https://hibernate.atlassian.net/browse/OGM-866


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/566

PR #567, opened 2015-07-28
OGM-868 Include git commit in the MANIFEST

DavideD:
https://hibernate.atlassian.net/browse/OGM-868

It will add to the Manifest of the jars an entry called git-commit-id. Example:
```
Git-Commit-Id: f2b416ae402a9101f0fca8521bf4c3c86055cf96                                                   
 ```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -162,6 +162,10 @@
         <defaultGoal>install</defaultGoal>
         <plugins>
             <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>buildnumber-maven-plugin</artifactId>
+            </plugin>
+            <plugin>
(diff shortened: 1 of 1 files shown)


gunnarmorling:
Looks good. Any particular motivation for introducing it?

DavideD:
I don't have a particular scenario in mind, it just feels like it could  be handy to have this information
if some user has a problem with dependencies. 

That said, I added it mainly to see if it works (I was helping @Sanne ); It was easy to add and it might be nice to try it.

DavideD:
Thinking about it,  this is helpful if someone uses SNAPSHOT to experiment with the features we add.
For example, it would be nmice for us to have this kind of information in the hibernate-orm project while we are experimenting with ORM5

gunnarmorling:
Cool :)

PR #568, opened 2015-07-29
OGM-555 More natural representation of maps in MongoDB

gunnarmorling:
For the case of map-typed associations which use a single key column of type `String`, we will use a more natural persistent mapping:

    {
      "_id" : "user_001",
      "addresses" : [
        {
          "work" : "address_001",
          "home" : "address_002"
        }
      ]
    }

For the other cases (composite map key, or single-column key of non-String type), the previous mapping is used:

    {
      "_id" : "user_001",
      "addresses" : [
        {
          "addressType" : 1,
          "addresses_id" : "address_001"
        },
        {
          "addressType" : 2,
          "addresses_id" : "address_002"
        }
      ]
    }

@emmanuelbernard I did not make it configurable but automatically enable the nicer mapping if possible. Do you see a case where it should be able to disable that?

A follow-up PR for CouchDB will follow.

### DIFF
--- a/documentation/manual/src/main/asciidoc/en-US/modules/mongodb.asciidoc
+++ b/documentation/manual/src/main/asciidoc/en-US/modules/mongodb.asciidoc
@@ -118,6 +118,16 @@ Possible strategies are (values of the `org.hibernate.ogm.datastore.mongodb.opti
 
 * `GLOBAL_COLLECTION` (default): stores the association information in a unique MongoDB collection for all associations
 * `COLLECTION_PER_ASSOCIATION` stores the association in a dedicated MongoDB collection per association
+
+hibernate.ogm.datastore.document.map_storage::
+Defines the way OGM stores the contents of map-typed associations in MongoDB.
+The following two strategies exist (values of the `org.hibernate.ogm.datastore.document.options.MapStorageType` enum):
+
+* `BY_KEY`: map-typed associations with a single key column which is of type `String` will be stored as a sub-document,
+organized by the given key; Not applicable for other types of key columns, in which case always `AS_LIST` will be used
+* `AS_LIST`: map-typed associations will be stored as an array containing a sub-document for each map entry.
+All key and value columns will be contained within the array elements
+
 hibernate.ogm.mongodb.write_concern::
 Defines the write concern setting to be applied when issuing writes against the MongoDB datastore.
 Possible settings are (values of the `WriteConcernType` enum):
@@ -214,7 +224,9 @@ When working with the MongoDB backend, you can specify the following settings:
 * the write concern for entities and associations using the `@WriteConcern` annotation
 * the read preference for entities and associations using the `@ReadPreference` annotation
 * a strategy for storing associations using the `@AssociationStorage` and `@AssociationDocumentStorage` annotations
-(refer to <<ogm-mongodb-storage-principles>> to learn more about these options).
+* a strategy for storing the contents of map-typed associations using the `@MapStorage` annotation
+
+Refer to <<mongodb-associations> to learn more about the options related to storing associations.
 
 The following shows an example:
 
@@ -227,6 +239,7 @@ The following shows an example:
 @ReadPreference(ReadPreferenceType.PRIMARY_PREFERRED)
 @AssociationStorage(AssociationStorageType.ASSOCIATION_DOCUMENT)
 @AssociationDocumentStorage(AssociationDocumentStorageType.COLLECTION_PER_ASSOCIATION)
+@MapStorage(MapStorageType.AS_LIST)
 public class Zoo {
 
     @OneToMany
@@ -267,6 +280,7 @@ When working with MongoDB, you can currently configure the following options usi
 * read preference
 * association storage strategy
 * association document storage strategy
+* strategy for storing the contents of map-typed associations
 
 To set these options via the API, you need to create an `OptionConfigurator` implementation
 as shown in the following example:
@@ -285,6 +299,7 @@ public class MyOptionConfigurator extends OptionConfigurator {
             .entity( Zoo.class )
                 .associationStorage( AssociationStorageType.ASSOCIATION_DOCUMENT )
                 .associationDocumentStorage( AssociationDocumentStorageType.COLLECTION_PER_ASSOCIATION )
+                .mapStorage( MapStorageType.ASLIST )
                 .property( "animals", ElementType.FIELD )
                     .associationStorage( AssociationStorageType.IN_ENTITY )
             .entity( Animal.class )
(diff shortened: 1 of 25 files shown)


DavideD:
> did not make it configurable but automatically enable the nicer mapping if possible. Do you see a case where it should be able to disable that?

It's a stretch but with this optimization in place you need to know the type of the key before running a query even if both collections have the same "address" field

In general, I'm concerned about the fact that the user might be surprise to have different behaviours when dealing with different primitive types. In particular, if they decide to update an existing model.

That said, I'm not against the change, since it seems more natural when dealing with strings.

gunnarmorling:
Added another commit for making it configurable whether maps with a single String key column should be mapped in the nicer or the legacy format. Default is the nicer one. For all other key column types this setting is ignored.

gunnarmorling:
Any more feedback on this, @emmanuelbernard, @DavideD et al.?

gunnarmorling:
@DavideD, any thoughts? Speak up today or stay silent forever, I will merge it tomorrow if not hearing back :)

DavideD:
:) I'll have a look but since emmanuel

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -510,23 +513,80 @@ public Association createAssociation(AssociationKey key, AssociationContext asso
...
+		}
+
+		if ( key.getMetadata().getRowKeyIndexColumnNames().length != 1 ) {
+			return false;
+		}
+
+		Object valueOfFirstRow = association.get( association.getKeys().iterator().next() )
+				.get( key.getMetadata().getRowKeyIndexColumnNames()[0] );
+
+		// TODO We may relax this to other single-column key types by conversion through grid type
```
DavideD: Is this TODO associated to some existing JIRA?

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/associations/MapMappingTest.java
```diff
@@ -0,0 +1,380 @@
...
+
+		// clean-up
+		user = (User) session.get( User.class, user.getId() );
+		session.delete( user );
+		session.delete( session.load( Address.class, home.getId() ) );
+		session.delete( session.load( Address.class, work.getId() ) );
+
+		tx.commit();
+		session.close();
+		sessionFactory.close();
```
DavideD: Wouldn't be better to put these in a finally block?

documentation/manual/src/main/asciidoc/en-US/modules/mongodb.asciidoc
```diff
@@ -1972,6 +1992,10 @@ Address collection as JSON in MongoDB
 ----
 ====
 
+In case you want to enforce the list-style represention also for maps with a single key column of type `String`
+(e.g. when reading back data persisted by earlier versions of Hibernate OGM),
+you can use the option `hibernate.ogm.datastore.document.map_storage` to do so.
```
DavideD: You would replace:
```you can use the option `hibernate.ogm.datastore.document.map_storage` to do so.```
with
```you can set the option `hibernate.ogm.datastore.document.map_storage` to the value `AS_LIST`.```

or something similar

File: documentation/manual/src/main/asciidoc/en-US/modules/mongodb.asciidoc [hunk 4]
```diff
@@ -118,6 +118,16 @@ Possible strategies are (values of the `org.hibernate.ogm.datastore.mongodb.opti
 
 * `GLOBAL_COLLECTION` (default): stores the association information in a unique MongoDB collection for all associations
 * `COLLECTION_PER_ASSOCIATION` stores the association in a dedicated MongoDB collection per association
+
+hibernate.ogm.datastore.document.map_storage::
+Defines the way OGM stores the contents of map-typed associations in MongoDB.
+The following two strategies exist (values of the `org.hibernate.ogm.datastore.document.options.MapStorageType` enum):
+
+* `BY_KEY`: map-typed associations with a single key column which is of type `String` will be stored as a sub-document,
+organized by the given key; Not applicable for other types of key columns, in which case always `AS_LIST` will be used
```
DavideD: I wonder if in the future might make sense to provide some kind of Bridge class to convert some types to String

DavideD: and vice-versa

DavideD:
@gunnarmorling 
It looks good to me, I've added some minor comments but they are not blocking.
I think it can be merged after a rebase.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
gunnarmorling: Not yet, as we are not sure whether we really want this or not. I'll ask for feedback in the announcement post. I'd implement it once there is demand for it.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/associations/MapMappingTest.java
gunnarmorling: Hum, I guess yes :) Though I don't really care in tests, it doesn't make a big difference there.

documentation/manual/src/main/asciidoc/en-US/modules/mongodb.asciidoc
gunnarmorling: That bridge class (hierarchy) is already there, `GridType` :) Question is whether we really want it or not.

documentation/manual/src/main/asciidoc/en-US/modules/mongodb.asciidoc
gunnarmorling: Done.

gunnarmorling:
Rebased and applied. Thanks for the review, @DavideD.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/associations/MapMappingTest.java
DavideD: I agree, but I'm trying to get in the habit since it helps to avoid unexpected errors due to something not being clean up. It also set an example for future  contributors.

DavideD: No need change the pull request now, though

PR #562, opened 2015-07-23
OGM-770 (Updating to ORM 5)

gunnarmorling:
Supersedes #524.

@emmanuelbernard Added some more commits to address your remarks and the things we discussed. Also rebased to latest master.

PR #571, opened 2015-07-30
OGM-770 ORM 5

DavideD:
https://hibernate.atlassian.net/browse/OGM-770

Rebase @gunnarmorling  pull request (https://github.com/hibernate/hibernate-ogm/pull/562) to the latest changes and update ORM to CR3.

The integration test are not working at the moment because WildFly is still on ORM CR2


gunnarmorling:
@DavideD When upgrading the WF directory to ORM 5 CR 3 (I will provide a temp. fix for that until we can go to WF 10 Beta1), all integration tests pass except two in Neo4j:

* `Neo4jResourceLocalModuleMemberRegistrationIT#shouldFindPersistedMemberByIdWithQuery`
* `Neo4jResourceLocalModuleMemberRegistrationIT#shouldFindPersistedMemberById()`

Their commit fails, but I think the actual error happens before, causing the TX to be closed when trying to commit. Does this ring a bell?

gunnarmorling:
@DavideD Any success with analyzing the remaining test failures?

DavideD:
Yes, it works now. But I haven't sent a pull request yet.
I cannot do that now. I will send a pull request in the evening.

gunnarmorling:
Cool. No need for another PR, just push another commit to the existing one.

DavideD:
I've addded some commits to the existing pull request.
I've updated the `Datastore` so the the logic to select the correct  `TransactionCoordinatorBuilder` is inside the `BaseDatastoreProvider` but can be overridden by specific datastore when the need arise.
It also allow the removal of the experimental method in the interface.

gunnarmorling:
Thanks, @DavideD. Rebased and applied.

The two TX coordinators look good. I've moved back the decision logic to `TransactionCoordinatorBuilder`, though I think it's better located there as a) not all providers necessarily extend `BaseDatastoreProvider` and b) implementors really should not be confronted with "JDBC"-named APIs. Hope that's alright, let me know in case you see any remaining issues.

DavideD:
I prefer my proposal, the builder received by the datastore is consistent with the real context and it removes the allowemulation method from the interface. the JDBC name in the API could have been easily renamed. The datastore  provider still needs to provide a TransactionCoordinatorBuilder so I don't see why not to delegate to it the creation.

Anyway, I don';t mind to leave like it as it is now. We will probably need to change it anyway when/if we will fix https://hibernate.atlassian.net/browse/OGM-763

PR #572, opened 2015-07-30
OGM-810 Switch to Java MongoDB Driver 3.x

hferentschik:
This is only the first step of the switch. At the moment we are still using the deprecated classes `DB`, `DBCollection` and `DBObject`. They should be used in a second step of this driver migration. I suggest to open a new follow up issue for that, so that this initial work can already be integrated (work on the deprecation removal is in progress, but not ready yet)

### DIFF
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBProperties.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBProperties.java
@@ -6,14 +6,14 @@
  */
 package org.hibernate.ogm.datastore.mongodb;
 
+import com.mongodb.ReadPreference;
+import com.mongodb.WriteConcern;
+
 import org.hibernate.ogm.cfg.OgmConfiguration;
 import org.hibernate.ogm.datastore.document.cfg.DocumentStoreProperties;
 import org.hibernate.ogm.datastore.document.options.AssociationStorageType;
 import org.hibernate.ogm.datastore.mongodb.options.ReadPreferenceType;
 import org.hibernate.ogm.datastore.mongodb.options.WriteConcernType;
-
-import com.mongodb.ReadPreference;
-import com.mongodb.WriteConcern;
 
 /**
  * Properties for configuring the MongoDB datastore via {@code persistence.xml} or {@link OgmConfiguration}.
@@ -56,11 +56,6 @@ public final class MongoDBProperties implements DocumentStoreProperties {
 	public static final String READ_PREFERENCE = "hibernate.ogm.mongodb.read_preference";
 
 	/**
-	 * The timeout used at the connection to the MongoDB instance. This value is set in milliseconds. Defaults to 5000.
-	 */
-	public static final String TIMEOUT = "hibernate.ogm.mongodb.connection_timeout";
-
-	/**
 	 * Configuration property for specifying how to store association documents. Only applicable if
 	 * {@link DocumentStoreProperties#ASSOCIATIONS_STORE} is set to {@link AssociationStorageType#ASSOCIATION_DOCUMENT}.
 	 * Supported values are the {@link org.hibernate.ogm.datastore.mongodb.options.AssociationDocumentStorageType} enum or the String representations of its constants.
@@ -77,6 +72,13 @@ public final class MongoDBProperties implements DocumentStoreProperties {
 	 * @see com.mongodb.MongoCredential
 	 */
 	public static final String AUTHENTICATION_MECHANISM = "hibernate.ogm.mongodb.authentication_mechanism";
+
+	/**
+	 * Property prefix for MongoDB driver settings which needs to be passed on to the driver. Refer to
+	 * the options of {@link com.mongodb.MongoClientOptions.Builder} for a list of available properties.
+	 * All string, int and boolean builder methods can be configured, eg {@code hibernate.ogm.mongodb.driver.maxWaitTime = 1000}.
+	 */
+	public static final String MONGO_DRIVER_SETTINGS_PREFIX = "hibernate.ogm.mongodb.driver";
 
 	private MongoDBProperties() {
 	}
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/configuration/impl/MongoDBConfiguration.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/configuration/impl/MongoDBConfiguration.java
@@ -6,10 +6,18 @@
  */
 package org.hibernate.ogm.datastore.mongodb.configuration.impl;
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
-import org.hibernate.HibernateException;
+import com.mongodb.MongoClientOptions;
+import com.mongodb.MongoCredential;
+import com.mongodb.ReadPreference;
+import com.mongodb.WriteConcern;
+
 import org.hibernate.ogm.cfg.spi.DocumentStoreConfiguration;
 import org.hibernate.ogm.datastore.mongodb.MongoDBProperties;
 import org.hibernate.ogm.datastore.mongodb.impl.MongoDBDatastoreProvider;
@@ -20,41 +28,25 @@ import org.hibernate.ogm.datastore.mongodb.options.impl.ReadPreferenceOption;
 import org.hibernate.ogm.datastore.mongodb.options.impl.WriteConcernOption;
 import org.hibernate.ogm.options.spi.OptionsContext;
 import org.hibernate.ogm.util.configurationreader.spi.ConfigurationPropertyReader;
-import org.hibernate.ogm.util.configurationreader.spi.PropertyValidator;
-
-import com.mongodb.MongoClientOptions;
-import com.mongodb.MongoCredential;
-import com.mongodb.ReadPreference;
-import com.mongodb.WriteConcern;
 
 /**
  * Configuration for {@link MongoDBDatastoreProvider}.
  *
  * @author Guillaume Scheibel &lt;guillaume.scheibel@gmail.com&gt;
  * @author Gunnar Morling
+ * @author Hardy Ferentschik
  */
 public class MongoDBConfiguration extends DocumentStoreConfiguration {
 
 	public static final String DEFAULT_ASSOCIATION_STORE = "Associations";
 
-	/**
-	 * The default value used to set the timeout during the connection to the MongoDB instance This value is set in
-	 * milliseconds.
-	 *
-	 * @see MongoDBProperties#TIMEOUT
-	 */
-	private static final int DEFAULT_TIMEOUT = 5000;
-
 	private static final int DEFAULT_PORT = 27017;
-
 	private static final Log log = LoggerFactory.getLogger();
 
-	private static final TimeoutValidator TIMEOUT_VALIDATOR = new TimeoutValidator();
-
-	private final int timeout;
 	private final WriteConcern writeConcern;
 	private final ReadPreference readPreference;
 	private final AuthenticationMechanismType authenticationMechanism;
+	private final ConfigurationPropertyReader propertyReader;
 
 	/**
 	 * Creates a new {@link MongoDBConfiguration}.
@@ -65,10 +57,7 @@ public class MongoDBConfiguration extends DocumentStoreConfiguration {
 	public MongoDBConfiguration(ConfigurationPropertyReader propertyReader, OptionsContext globalOptions) {
 		super( propertyReader, DEFAULT_PORT );
 
-		this.timeout = propertyReader.property( MongoDBProperties.TIMEOUT, int.class )
-				.withDefault( DEFAULT_TIMEOUT )
-				.withValidator( TIMEOUT_VALIDATOR )
-				.getValue();
+		this.propertyReader = propertyReader;
 		this.authenticationMechanism = propertyReader.property( MongoDBProperties.AUTHENTICATION_MECHANISM, AuthenticationMechanismType.class )
 				.withDefault( AuthenticationMechanismType.BEST )
 				.getValue();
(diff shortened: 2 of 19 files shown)


mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/WriteConcernType.java
```diff
@@ -25,7 +28,7 @@
 	 * @deprecated This WriteConcern is no longer supported by MongoDB
 	 */
 	@Deprecated
-	ERRORS_IGNORED(WriteConcern.ERRORS_IGNORED),
```
gunnarmorling: I would completely remove it. It has been deprecated in OGM 4.2.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/type/impl/CharacterStringType.java
```diff
@@ -0,0 +1,40 @@
...
+import org.hibernate.type.descriptor.java.CharacterTypeDescriptor;
+
+/**
+ * For MongoDB persist a {@link Character} as a {@link String}.
+ * Necessary since MongoDB driver 3 no longer does that for us
+ * https://jira.mongodb.org/browse/JAVA-1804
+ *
+ * @author Emmanuel Bernard emmanuel@hibernate.org
+ */
+public class CharacterStringType extends AbstractGenericBasicType<Character>  {
```
gunnarmorling: Could go to core? E.g. the Hash-based dialect for Redis could re-use it.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBProperties.java
```diff
@@ -57,7 +57,9 @@
 
 	/**
 	 * The timeout used at the connection to the MongoDB instance. This value is set in milliseconds. Defaults to 5000.
+	 * @deprecated Use {@code hibernate.ogm.mongodb.driver.connectTimeout} instead.
 	 */
+	@Deprecated
```
gunnarmorling: Should we remove it right away? It's going to be a new major version, so this sort of breakage seems acceptable to me.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBProperties.java
```diff
@@ -78,6 +80,13 @@
 	 */
 	public static final String AUTHENTICATION_MECHANISM = "hibernate.ogm.mongodb.authentication_mechanism";
 
+	/**
+	 * Property prefix for MongoDB driver settings which needs to ne passed on tp the driver. Refer to
```
gunnarmorling: "to be passed on to the driver"

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/configuration/impl/MongoDBConfiguration.java
```diff
@@ -82,14 +86,62 @@ public MongoDBConfiguration(ConfigurationPropertyReader propertyReader, OptionsC
...
+	private Map<String, Method> createSettingsMap() {
+		Map<String, Method> settingsMap = new HashMap<>();
+
+		Method[] methods = MongoClientOptions.Builder.class.getDeclaredMethods();
+		for ( Method method : methods ) {
+			if ( method.getParameterTypes().length == 1 ) {
+				Class<?> parameterType = method.getParameterTypes()[0];
+				// we just care of string, int and boolean setters
+				if ( String.class.equals( parameterType )
+						|| int.class.equals( parameterType )
```
gunnarmorling: `ReadPreference` and `WriteConcern` could be supported as well that way. WDYT?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/configuration/impl/MongoDBConfiguration.java
```diff
@@ -82,14 +86,62 @@ public MongoDBConfiguration(ConfigurationPropertyReader propertyReader, OptionsC
...
+			if ( property == null ) {
+				continue;
+			}
+
+			Method settingMethod = entry.getValue();
+			try {
+				settingMethod.invoke( optionsBuilder, property );
+			}
+			catch ( InvocationTargetException | IllegalAccessException e ) {
+				// TODO OGM-810 fix error message
```
gunnarmorling: Could be done now?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
```diff
@@ -136,9 +133,6 @@ protected MongoClient createMongoClient(MongoDBConfiguration config) {
 					? new MongoClient( serverAddresses, clientOptions )
 					: new MongoClient( serverAddresses, credentials, clientOptions );
 		}
-		catch (UnknownHostException e) {
-			throw log.mongoOnUnknownHost( config.toString() );
```
gunnarmorling: This log method seems obsolete as well now.

gunnarmorling:
Thanks, @hferentschik! Added some comments, nothing too exciting.

> At the moment we are still using the deprecated classes DB, DBCollection and DBObject

You mean e.g. `com.mongodb.DBObject`? I don't see them marked as deprecated in the 3.0.3 JAR. Do you have any reference to their deprecation and the recommended replacements? Hope we don't need to re-write everything, these classes are really central :)

hferentschik:
> Do you have any reference to their deprecation and the recommended replacements?

Sure - https://www.mongodb.com/blog/post/introducing-30-java-driver

    So for users of the 2.x driver series that are upgrading to 3.0, the driver still has the 
    MongoClient.getDB("myDB") method and, although it has been deprecated, rest easy -- it won't be
    removed in the 3.x series. The deprecation is intended to nudge new applications towards the 
    MongoDatabase/MongoCollection API, which is accessible via the new 
    MongoClient.getDatabase("myDB") method. This way we retain backwards compatibility while still 
    offering an improved API that users can migrate to at their own pace. - See more at:  
    https://www.mongodb.com/blog/post/introducing-30-java-driver#sthash.JDGalGNX.dpuf

So at the moment we still use the deprecated `MongoClient.getDB` call. Not everything underneath is deprecated (I guess they wanted to keep the warnings down), but once you switch to `MongoClient.getDatabase` you are going down a whole rabbit hole of changes. As said, I am somewhere in the middle of this, but wanted to get this in first.

This is how CRUD works now - http://mongodb.github.io/mongo-java-driver/3.0/driver/reference/crud/. I guess `DBObject` was maybe wrong to list, but the changeset becomes quite large.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
hferentschik: It is. I was not sure what policies you have. I can delete the log message in the `Log` class as well, if you like.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/configuration/impl/MongoDBConfiguration.java
hferentschik: sure, I missed this one

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/configuration/impl/MongoDBConfiguration.java
hferentschik: well, there would would have to start instantiating classes. Could be done for sure. just needs a bit more doing.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBProperties.java
hferentschik: +1

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBProperties.java
hferentschik: +1 works for me

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/WriteConcernType.java
hferentschik: ok

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/type/impl/CharacterStringType.java
hferentschik: Ok, no idea. I trust you on this one.

hferentschik:
Forced pushed an update

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/configuration/impl/MongoDBConfiguration.java
gunnarmorling: It'd seem more consistent to me. Right now these two are the only one which are handled in a special way (besides that factory thingy which we don't support for now anyways). Wrt. instantiation, the property reader can deal with enum values, too. The code for parsing the current properties should be using it.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
gunnarmorling: We don't have any written policy, but removing unused stuff seems good to me. Just not should re-use ids for new messages.

gunnarmorling:
>  As said, I am somewhere in the middle of this, but wanted to get this in first.

+1 Let's open a follow-up issue for that.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/configuration/impl/MongoDBConfiguration.java
gunnarmorling: @hferentschik You think you could do that as part of this PR? If not, I'll open another follow-up just for that one, as it can be done separately from (and much quicker than) addressing the deprecation issues.

hferentschik: I guess you are right, especially if the property reader can handle enums

Sent using CloudMagic [https://cloudmagic.com/k/d/mailapp?ct=pi&cv=6.0.64&pv=8.1.3]
On Sat, Aug 1, 2015 at 08:42, hibernate/hibernate-ogm
<reply@reply.github.com>
wrote:
In 
mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/configuration/impl/MongoDBConfiguration.java
[https://github.com/hibernate/hibernate-ogm/pull/572#discussion_r36031315] :

>  		return optionsBuilder.build();
>  	}
>  
> +	private Map<String, Method> createSettingsMap() {
> +		Map<String, Method> settingsMap = new HashMap<>();
> +
> +		Method[] methods = MongoClientOptions.Builder.class.getDeclaredMethods();
> +		for ( Method method : methods ) {
> +			if ( method.getParameterTypes().length == 1 ) {
> +				Class<?> parameterType = method.getParameterTypes()[0];
> +				// we just care of string, int and boolean setters
> +				if ( String.class.equals( parameterType )
> +						|| int.class.equals( parameterType )


It'd seem more consistent to me. Right now these two are the only one which are
handled in a special way (besides that factory thingy which we don't support for
now anyways). Wrt. instantiation, the property reader can deal with enum values,
too. The code for parsing the current properties should be using it.

—
Reply to this email directly or view it on GitHub
[https://github.com/hibernate/hibernate-ogm/pull/572/files#r36031315] .[https://github.com/notifications/beacon/AASFojcXqzuego_pjTGcl51Hdrc5G6axks5ojGHsgaJpZM4FjGq_.gif]

hferentschik: I guess. Depends when you want to merge. I am off next week. I might find some
time somewhere, but not sure. If you want to merge asap, a follow up issue might
be better'

Sent using CloudMagic [https://cloudmagic.com/k/d/mailapp?ct=pi&cv=6.0.64&pv=8.1.3]
On Sat, Aug 1, 2015 at 11:12, hibernate/hibernate-ogm
<reply@reply.github.com>
wrote:
In 
mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/configuration/impl/MongoDBConfiguration.java
[https://github.com/hibernate/hibernate-ogm/pull/572#discussion_r36032088] :

>  		return optionsBuilder.build();
>  	}
>  
> +	private Map<String, Method> createSettingsMap() {
> +		Map<String, Method> settingsMap = new HashMap<>();
> +
> +		Method[] methods = MongoClientOptions.Builder.class.getDeclaredMethods();
> +		for ( Method method : methods ) {
> +			if ( method.getParameterTypes().length == 1 ) {
> +				Class<?> parameterType = method.getParameterTypes()[0];
> +				// we just care of string, int and boolean setters
> +				if ( String.class.equals( parameterType )
> +						|| int.class.equals( parameterType )


@hferentschik [https://github.com/hferentschik] You think you could do that as part of this PR? If not, I'll open another
follow-up just for that one, as it can be done separately from (and much quicker
than) addressing the deprecation issues.

—
Reply to this email directly or view it on GitHub
[https://github.com/hibernate/hibernate-ogm/pull/572/files#r36032088] .[https://github.com/notifications/beacon/AASFoslQVis7S5Ox7Khsgzscgsu1p5a3ks5ojIUTgaJpZM4FjGq_.gif]

hferentschik: > Wrt. instantiation, the property reader can deal with enum values, too. The code for parsing the current properties should be using it.

I had a quick look, the problem is that the driver's `WriteConcern` class is not an emum class, but rather a plain class with some public static instances defining the different write concerns. OGM provides an emum `WriteConcernType` which maps the different `WriteConcern` instances to enum values (same for `ReadPreference`). So there would be some special handling required for write concern and read preference. I guess from a user perspective it would still make sense to have all driver properties under the same prefix, but code wise we would need to handle these two cases explicitly.

hferentschik: Actually there is yet another little complication. `WriteConcernType` is not a 1:1 mapping to the possible `WriteConcern` instances.There is also `WriteConcernType#CUSTOM` which maps to no pre-defined `WriteConcern`, but is supposed to be used in conjunction with `hibernate.ogm.mongodb.write_concern_type`. If we were to support the property `hibernate.ogm.mongodb.driver.writeConcern` we would need to drop `WriteConcernType#CUSTOM`, since the user would expect a 1:1 mapping here. 

Personally I find `hibernate.ogm.mongodb.write_concern_type` a bit odd. I think the best solution would be to just have `hibernate.ogm.mongodb.driver.writeConcern` and handle the cases where the value is a string representation of one of the pre-defined write concern instances, a fully qualified class name or an actual `WriteConcern` instance. 

Thougths?

gunnarmorling: +1 For removing the old OGM properties for that and expose it via the new settings in the "driver" sub-namespace.

> I think the best solution would be to just have hibernate.ogm.mongodb.driver.writeConcern and handle the cases where the value is a string representation of one of the pre-defined write concern instances, a fully qualified class name or an actual WriteConcern instance.

Yep, that sounds reasonable. `WriteConcern` has several constructors allowing for ad-hoc creation of arbitrarily configured instances, but naturally we cannot support that via text-based configuration properties. So users would have to create specific sub-classes of `WriteConcern` in such cases.

Coding-wise it may be not too bad, check out the concept of `ShortNameProvider` as used in `DatastoreProviderInitiator`. There we also read a setting with known (enum-like) values and a FQN as fallback.

Another thing to consider is the annotation-based config and the programmatic config corresponding to these "write concern"/"read preference" properties (which allow for fine-grained control on the entity/property level). The `@WriteConcern` annotation and the API methods (see `MongoDBGlobalContext` et al.) accept the OGM enum. We meant to not expose the MongoDB-specific types in the API directly, but if we want to support custom write concern settings, I don't think it can be avoided.

hferentschik: > Another thing to consider is the annotation-based config and the programmatic config corresponding to these "write concern"/"read preference" properties

Looking at this it makes me wonder whether the old properties should really be replaced by a setting under the "driver" namespace. If we do so, what do we do with the annotation based approach. Letting it use `WriteConcernType` seems inconsistent and switching to a single string property (which probably makes most sense imo) would loose some type safety. 

> which allow for fine-grained control on the entity/property level)

OOI, is this fine grained configuration based on an actual usecase and/or user request or is it just there to prove the concept of the options system?

gunnarmorling:
Rebased and applied. I've opened [OGM-901](https://hibernate.atlassian.net/browse/OGM-901) for deciding on what to do with the write concern / read preference options.

Thanks, @hferentschik!

hferentschik:
Thanks :-)

PR #573, opened 2015-08-03
OGM-885 Add Redis to distribution ZIP

mp911de:
* Add Redis to distribution ZIP
* Fix javadoc inclusion
* add Redis as experimental data store to docs


gunnarmorling:
Rebased and applied. Added one more commit to fix some JavaDoc errors. Thanks, @mp911de!

PR #574, opened 2015-08-04
OGM-887 Don't map elements within embeddable collections using dot names

mp911de:
* Don't map elements within embeddable collections using dot names
* Remove the common prefixes from the JSON representation
* Refactor parts of MongoHelpers into DotPatternMapHelpers

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/dialect/impl/DotPatternMapHelpers.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/impl/DotPatternMapHelpers.java
@@ -1,0 +1,95 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.dialect.impl;
+
+import java.util.Map;
+import java.util.regex.Pattern;
+
+/**
+ * Provides functionality for dealing with (nested) fields of Map documents.
+ *
+ * @author Alan Fitton &lt;alan at eth0.org.uk&gt;
+ * @author Emmanuel Bernard &lt;emmanuel@hibernate.org&gt;
+ * @author Gunnar Morling
+ */
+public class DotPatternMapHelpers {
+
+	private static final Pattern DOT_SEPARATOR_PATTERN = Pattern.compile( "\\." );
+
+	/**
+	 * Remove a column from the Map
+	 *
+	 * @param entity the {@link Map} with the column
+	 * @param column the column to remove
+	 */
+	public static void resetValue(Map<?,?> entity, String column) {
+		// fast path for non-embedded case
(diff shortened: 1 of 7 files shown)


mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/dialect/impl/MongoDBAssociationSnapshot.java
```diff
@@ -46,11 +47,11 @@ public DBObject getQueryObject() {
 
 		if ( storageStrategy == AssociationStorageStrategy.IN_ENTITY ) {
 			if ( associationKey.getMetadata().isOneToOne() ) {
-				Object oneToOneValue = getValueOrNull( document, associationKey.getMetadata().getCollectionRole(), Object.class );
+				Object oneToOneValue = getValueOrNull( (Map<?,?>) document, associationKey.getMetadata().getCollectionRole(), Object.class );
```
gunnarmorling: `DBObject` is not safe to be cast to `Map`. The default implementation `BasicDBObject` indeed is, but it's a slippery slope we'd enter when relying on it. I'd prefer to duplicate that code for now (apparently, MongoDB driver 3 comes with a replacement for `DBObject` et al. anyways, so we may re-visit the topic then).

core/src/main/java/org/hibernate/ogm/dialect/impl/DotPatternMapHelpers.java
```diff
@@ -0,0 +1,95 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.dialect.impl;
```
gunnarmorling: It'd be better located in `org.hibernate.ogm.datastore.document.impl`.

Btw. to give some background, the "*.spi" packages are meant to provide the required types/functionality for implementing dialects. We give some guarantees regarding compatible evolvement there. Other things where we don't feel like committing to that yet, we have under "impl" and still refer them from "our" dialects. The idea is to promote those things to "spi" if they have proven over time. This helper also feels like it's rightly located under "impl" for now and once we see it fits for different use cases / dialects, we can relocate it in a subsequent release.

gunnarmorling:
@mp911de Looks good to me apart from the two remarks above. Could you also rebase it to the latest master? Thanks a lot!

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/dialect/impl/MongoDBAssociationSnapshot.java
mp911de: Ok, will revert the changes to the Mongo code

mp911de:
Thanks for reviewing, I fixed the two issues and rebased the code on the latest master.

gunnarmorling:
Merged; Thanks, @mp911de!

PR #575, opened 2015-08-04
OGM-886 Ship own netty jars to omit connection-refused-bug

mp911de:
Using lettuce 3.2.Final with netty 4.0.26 runs into a deadlock/infinite wait when the connection runs into a connection refused. Ship own netty jars until then.


gunnarmorling:
Thanks, @mp911de. Created [OGM-890](https://hibernate.atlassian.net/browse/OGM-890) for using the WF module once it's updated.

PR #576, opened 2015-08-04
OGM-874 Multi-get for Redis

gunnarmorling:
@mp911de Any thoughts? This adds multi-get support for Redis. That way we save some round trips when e.g. fetching the elements of a lazy collection (see `@BatchSize`). It also provides some fixes around `RedisDialect`.

### DIFF
--- a/redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
+++ b/redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
@@ -84,7 +84,8 @@ public class RedisDialect extends BaseGridDialect implements MultigetGridDialect
 
 	@Override
 	public Tuple getTuple(EntityKey key, TupleContext tupleContext) {
-		Entity entity = getEntity( key );
+		Entity entity = entityStorageStrategy.getEntity( entityId( key ) );
+
 		if ( entity != null ) {
 			return new Tuple( new RedisTupleSnapshot( entity.getProperties() ) );
 		}
@@ -369,15 +370,6 @@ public class RedisDialect extends BaseGridDialect implements MultigetGridDialect
 		}
 	}
 
-	// Retrieve entity and enhance entity data with the entity key
-	private Entity getEntity(EntityKey key) {
-		Entity entity = entityStorageStrategy.getEntity( entityId( key ) );
-		if ( entity != null ) {
-			addIdToEntity( entity, key.getColumnNames(), key.getColumnValues() );
-		}
-		return entity;
-	}
-
 	private void addIdToEntity(Entity entity, String[] columnNames, Object[] columnValues) {
 		for ( int i = 0; i < columnNames.length; i++ ) {
 			entity.set( columnNames[i], columnValues[i] );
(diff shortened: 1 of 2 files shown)


redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
```diff
@@ -534,16 +527,16 @@ public JsonEntityStorageStrategy getEntityStorageStrategy() {
...
+	 * Construct a key based on the key columns:
+	 * Single key: Use the value as key
+	 * Multiple keys: Serialize the key using a JSON map.
+	 */
 	private byte[] keyToBytes(String[] columnNames, Object[] columnValues) {
 		if ( columnNames.length == 1 ) {
 
 			if ( columnValues[0] instanceof CharSequence ) {
-				return columnValues[0].toString().getBytes();
+				return toBytes( columnValues[0].toString() );
```
mp911de: This `toBytes` yields in the same as the one without the `if` block.

redis/src/main/java/org/hibernate/ogm/datastore/redis/impl/json/JsonEntityStorageStrategy.java
```diff
@@ -41,4 +42,33 @@ public void storeEntity(byte[] key, Entity entity, Set<TupleOperation> operation
 
 		connection.set( key, value );
 	}
+
+	public Iterable<Entity> getEntities(byte[][] keys) {
+		final Iterator<byte[]> values = connection.mget( keys ).iterator();
```
mp911de: `MGET` is the perfect fit

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
gunnarmorling: Not sure I am following? The motivation is not to depend on the platform default encoding but always use UTF-8.

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
```diff
@@ -83,7 +84,8 @@ public GridType overrideType(Type type) {
 
 	@Override
 	public Tuple getTuple(EntityKey key, TupleContext tupleContext) {
-		Entity entity = getEntity( key );
+		Entity entity = entityStorageStrategy.getEntity( entityId( key ) );
```
mp911de: Adding Id's back to the entity is only necessary for bulk access such as `getTuples` and `forEachTuple`?

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
mp911de: That's right. I wanted to point out, that the `if (columnValues[0] instanceof CharSequence)` could be removed since both branches of the if call `toBytes( columnValues[0].toString() );` now.

mp911de:
The PR looks very decent.

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
gunnarmorling: Yes, right. I've filed an issue to clarify it in the JavaDocs, the different dialects handle it not consistently atm.

On a related note, I am also thinking about making `EntityKey` an interface and let dialects return specific implementations. That way we may avoid some copying between `EntityKey` and the backend-specific structures.

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
gunnarmorling: Ah, got it now. It's getting to late for me :)

redis/src/main/java/org/hibernate/ogm/datastore/redis/impl/json/JsonEntityStorageStrategy.java
gunnarmorling: Cool, very nice to see this supported by Redis.

gunnarmorling:
Cool, thanks for reviewing it! I'll clean-up the if/else glitch and merge
it.

gunnarmorling:
Rebased and applied.

PR #577, opened 2015-08-07
OGM-880 Keep connection alive between tests

mp911de:
Use TestRedisDatastoreProvider for unit tests to operate on a static held RedisClient instance that is closed using a shutdown hook.


gunnarmorling:
Rebased and applied. Thanks, @mp911de!

PR #578, opened 2015-08-12
update-readme

ZJaffee:
fixed some formatting issues with the readme, let me know if I need to refactor this commit message to fit in with your commit message styling.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -63,9 +63,9 @@ unpack the modules in it and run the tests using Arquillian.
 
 [WARNING]
 ====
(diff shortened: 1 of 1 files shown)


DavideD:
Hi @ZJaffee,
thanks for the patch, there is actuallly something wrong in the rendering of that part.

But this readme it is written using markdown, it should not have html tags.

As far as I can see GitHub does not have warning boxes so I'm not sure how we should render it.
Maybe a table?

I'll close this pull request for now, but feel free to reopen it if you update it , or send a new pull request if you prefer

PR #580, opened 2015-08-19
OGM-891 Use more natural format to store map-typed properties in Redis

mp911de:


core/src/main/java/org/hibernate/ogm/datastore/keyvalue/options/navigation/spi/BaseKeyValueStoreEntityContext.java
```diff
@@ -22,4 +24,13 @@
 	public BaseKeyValueStoreEntityContext(ConfigurationContext context) {
 		super( context );
 	}
+
+	public E mapStorage(MapStorageType mapStorage) {
```
gunnarmorling: I am a bit reluctant about this as a common option for all K/V stores. Can we expose it only for Redis at this point where we really support it? We still can pull it up later on.

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
```diff
@@ -1196,11 +1197,11 @@ Key: User:user_001
 { 
   "addresses" : [
     { 
-      "addresses_KEY" : "work",
+      "addressType" : "work",
```
gunnarmorling: Is that actually right? The example seems not to re-define the map key column? Then by default indeed the `_KEY` suffix is appended.

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
```diff
@@ -345,6 +385,42 @@ else if ( currentTtl != null && currentTtl > 0 ) {
 		return rows;
 	}
 
+	/**
+	 * Whether the rows of the given association should be stored in a hash using the single row key column as key or
+	 * not.
+	 */
+	private boolean organizeByRowKey(
```
gunnarmorling: I guess that could be extracted to a common place as well?

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
```diff
@@ -66,15 +66,6 @@ TTL can be configured on entity and association level (see <<ogm-redis-annotatio
 This property has no default value.
 hibernate.ogm.error_handler::
 The fully-qualified class name, class object or an instance of `ErrorHandler` to get notified upon errors during flushes (see <<ogm-api-error-handler>>)
-hibernate.ogm.datastore.redis.entity_storage::
-Defines the way OGM stores entity information in Redis.
-The following two strategies exist (values of the `org.hibernate.ogm.datastore.redis.options.EntityStorageType` enum):
-* `JSON` (store entities as JSON documents with a key) and
-* `HASH` (store entities within hashes. The fields and values are represented as strings within the hash).
```
gunnarmorling: :)

gunnarmorling:
@mp911de Added some minor remarks, but looks good to me overall.

core/src/main/java/org/hibernate/ogm/datastore/keyvalue/options/navigation/spi/BaseKeyValueStoreEntityContext.java
mp911de: Done.

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
mp911de: Adjusted the example using the resulting JSON of `org.hibernate.ogm.datastore.redis.test.associations.MapMappingTest`

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
mp911de: Moved it to `org.hibernate.ogm.datastore.document.impl.DotPatternMapHelpers`

gunnarmorling:
Rebased and applied. Added two more commits for some minor fix ups. Thanks!

PR #582, opened 2015-08-19
OGM-896 Start Redis and CouchDB integration tests only when required

DavideD:
https://hibernate.atlassian.net/browse/OGM-896


gunnarmorling:
Merged; Thanks, @DavideD!

PR #583, opened 2015-08-21
OGM-810 Update Fongo to version 2.0.2

DavideD:
https://hibernate.atlassian.net/browse/OGM-810

This version will work with   MongoDB Driver 3


DavideD:
@gunnarmorling  THis should fix the issue with fongo

PR #585, opened 2015-08-28
OGM-907 MultiGet  for Neo4j and MongoDB

DavideD:
https://hibernate.atlassian.net/browse/OGM-907

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/dialect/multiget/spi/MultigetGridDialect.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/multiget/spi/MultigetGridDialect.java
@@ -11,6 +11,7 @@ import java.util.List;
 import org.hibernate.ogm.dialect.spi.GridDialect;
 import org.hibernate.ogm.dialect.spi.TupleContext;
 import org.hibernate.ogm.model.key.spi.EntityKey;
+import org.hibernate.ogm.model.key.spi.EntityKeyMetadata;
 import org.hibernate.ogm.model.spi.Tuple;
 
 /**
@@ -24,6 +25,8 @@ public interface MultigetGridDialect extends GridDialect {
 	 * Return the list of tuples with the given columns for a given list of keys.
 	 * The tuples must be returned in the same order as the keys.
 	 * If a key has no matching record, set null to the list entry.
+	 * <p>
+	 * All the keys must have the same {@link EntityKeyMetadata}.
 	 *
 	 * @param keys The array of tuple identifier
 	 * @param tupleContext Contains additional information that might be used to create the tuples
--- a/core/src/main/java/org/hibernate/ogm/dialect/multiget/spi/MultigetGridDialect.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/multiget/spi/MultigetGridDialect.java
@@ -26,7 +26,8 @@ public interface MultigetGridDialect extends GridDialect {
 	 * The tuples must be returned in the same order as the keys.
 	 * If a key has no matching record, set null to the list entry.
 	 * <p>
-	 * All the keys must have the same {@link EntityKeyMetadata}.
+	 * All the keys provided will have the same {@link EntityKeyMetadata}.
+	 * In other words they target the same "table".
 	 *
 	 * @param keys The array of tuple identifier
 	 * @param tupleContext Contains additional information that might be used to create the tuples
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
@@ -60,6 +60,7 @@ import org.hibernate.ogm.dialect.batch.spi.OperationsQueue;
 import org.hibernate.ogm.dialect.batch.spi.RemoveAssociationOperation;
 import org.hibernate.ogm.dialect.batch.spi.RemoveTupleOperation;
 import org.hibernate.ogm.dialect.identity.spi.IdentityColumnAwareGridDialect;
+import org.hibernate.ogm.dialect.multiget.spi.MultigetGridDialect;
 import org.hibernate.ogm.dialect.optimisticlock.spi.OptimisticLockingAwareGridDialect;
 import org.hibernate.ogm.dialect.query.spi.BackendQuery;
 import org.hibernate.ogm.dialect.query.spi.ClosableIterator;
@@ -131,7 +132,7 @@ import com.mongodb.WriteConcern;
  * @author Alan Fitton &lt;alan at eth0.org.uk&gt;
  * @author Emmanuel Bernard &lt;emmanuel@hibernate.org&gt;
  */
-public class MongoDBDialect extends BaseGridDialect implements QueryableGridDialect<MongoDBQueryDescriptor>, BatchableGridDialect, IdentityColumnAwareGridDialect, OptimisticLockingAwareGridDialect {
+public class MongoDBDialect extends BaseGridDialect implements QueryableGridDialect<MongoDBQueryDescriptor>, BatchableGridDialect, IdentityColumnAwareGridDialect, MultigetGridDialect, OptimisticLockingAwareGridDialect {
 
 	public static final String ID_FIELDNAME = "_id";
 	public static final String PROPERTY_SEPARATOR = ".";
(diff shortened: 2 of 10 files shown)


core/src/main/java/org/hibernate/ogm/dialect/multiget/spi/MultigetGridDialect.java
```diff
@@ -10,24 +10,26 @@
...
 	 *
-	 * @param keys The array of tuple identifier
+	 * @param metadata the {@link EntityKeyMetadata} of the expected tuples
+	 * @param values the values of the key columns, each row of the matrix contains one key values
 	 * @param tupleContext Contains additional information that might be used to create the tuples
-	 * @return the list of tuples identified by the keys
+	 * @return the list of tuples identified by the key values
 	 */
-	List<Tuple> getTuples(EntityKey[] keys, TupleContext tupleContext);
+	List<Tuple> getTuples(EntityKeyMetadata metadata, Object[][] values, TupleContext tupleContext);
```
gunnarmorling: My preference would be to leave it as `EntityKey[]`, the two-dimensional array seems not as nice to me.

I wouldn't be worried about the duplication of `EntityKeyMetadata` for the time being. I'll try and remove it from `EntityKey` next week, which would address that duplication here.

DavideD: It's not really about the duplication, the fact is that separating the metadata make it obvious that the keys are all of the same type. While using EntityKey[] make it looks like the key might be of different Entities types.

gunnarmorling: But the explicitly passed EKM makes it apparent that all keys are of that same type. Passing data around as `Object[][]` makes it just unnecessarily hard to understand the stuff, that's what classes are for :) Also that signature does not make clear anymore that this is about entity keys at all, it could just be anything.

DavideD: Ok, I'll revert the change

DavideD:
I've updated the pull, keeping the keys in the signature of MultigetGridDialect

DavideD:
Rebased

emmanuelbernard:
I would prefer if we had a separated specific issue for each dialect implementing multiget, one for mongo, one for neo4j

emmanuelbernard:
@DavideD review done. Minor adjustments and questions.

DavideD:
Thanks, I'm answering them

DavideD:
@emmanuelbernard I've applied the changes:
- update the javadoc
- set read preference in MongoDB
- rename NONE to NOT_IN_THE_DB in the tests 

> I would prefer if we had a separated specific issue for each dialect implementing multiget, one for mongo, one for neo4j

Me too, I will keep that in mind for future issues

DavideD:
Jenkins, test this, please

DavideD:
Jenkins, test this, please

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -154,6 +155,44 @@ public MongoDBDialect(MongoDBDatastoreProvider provider) {
...
+		finally {
+			if ( cursor != null ) {
+				cursor.close();
+			}
+		}
+	}
+
+	private List<Tuple> createListOfTuples(EntityKey[] keys) {
+		List<Tuple> tuples = new ArrayList<>( keys.length );
+		for ( int i = 0; i < keys.length; i++ ) {
```
gunnarmorling: That should not be needed? It'll be nulls anyways.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/Neo4jDialect.java
```diff
@@ -185,6 +188,49 @@ public Tuple getTuple(EntityKey key, TupleContext context) {
...
 
 	@Override
+	public List<Tuple> getTuples(EntityKey[] keys, TupleContext tupleContext) {
+		List<Tuple> tuples = createListOfTuples( keys );
+		// We only supports one metadata for now
+		EntityKeyMetadata metadata = keys[0].getMetadata();
+		ResourceIterator<Node> nodes = entityQueries.get( metadata ).findEntities( dataBase, keys );
+		try {
+			while ( nodes.hasNext() ) {
+				Node node = nodes.next();
```
gunnarmorling: So the nodes returned will not be in the order of the given keys?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
gunnarmorling: Ok, got it now. I think I'd rather add the nulls when iterating the cursor, it seems a bit easier to grok to future readers.

gunnarmorling: It'd also avoid looping once, to give a less style-based argument :) Same for Neo4j btw.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jEntityQueries.java
```diff
@@ -170,6 +185,39 @@ private static String initFindEntitiesQuery(EntityKeyMetadata entityKeyMetadata)
...
+	 * MATCH (n:ENTITY:table)
+	 * WHERE n.id IN {0}
+	 * RETURN n
+	 *
+	 * 2) id is mapped on multiple columns:
+	 *
+	 * MATCH (n:ENTITY:table)
+	 * WHERE
+	 *
+	 * In this case the query depends on how many id we are retrieving and it will completed later
```
gunnarmorling: "will be"

gunnarmorling:
@DavideD Added to more minor remarks, good to go otherwise.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
DavideD: > Ok, got it now. I think I'd rather add the nulls when iterating the cursor,

The number of results returned by the cursor might not match the number of keys, some entities might not be in the db. Not sure how you would do that.

> ... to give a less style-based argument 

Not sure what you mean

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jEntityQueries.java
DavideD: fixed

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
gunnarmorling: Thinking more about it, the two nested loops seem not optimal. If you could ensure the order of results matches the order of keys (e.g. by applying a sort by id), it could be avoided by processing the results like so:

    while ( result.hasNext() ) {
        advance incoming key iterator to matching key
        add null to result list for each advancing without matching
        when find the matching key, add the result
        advance result iterator
    }

That way you'd iterate the keys once and the results once and not the keys once per result.

gunnarmorling: Hey @DavideD, wondering whether you ever followed up on this idea?

DavideD:
I've rebase the pull request to the latest changes and I've applied some refactoring.

> If you could ensure the order of results matches the order of keys (e.g. by applying a sort by id).

@gunnarmorling, as far as I know there is no way to do it in Neo4j. In MongoDB there is a discussion about  adding the feature we would need.

gunnarmorling:
Rebased and applied. Thanks, @DavideD! I hope we can improve on the sorting issue at some point, I've opened [OGM-940](https://hibernate.atlassian.net/projects/OGM/issues/OGM-940) for this.

gunnarmorling:
Also added two more commits for some minor improvements.

PR #586, opened 2015-08-28
OGM-904 Use more natural format to store map-typed properties in CouchDB

mp911de:
- Fix properties with dot path names of embedded documents by converting those to nested objects
- Extract common parts from RedisDialect to DotPatternMapHelpers
- Adjust docs

Hint: I wasn't able to run the full test suite successfully even before the change. Need your advice then.

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/datastore/document/impl/DotPatternMapHelpers.java
+++ b/core/src/main/java/org/hibernate/ogm/datastore/document/impl/DotPatternMapHelpers.java
@@ -133,6 +133,13 @@ public class DotPatternMapHelpers {
 		return associationContext.getAssociationTypeContext().getOptionsContext().getUnique( MapStorageOption.class );
 	}
 
+	/**
+	 * Returns the shared prefix with a trailing dot (.) of the association columns or {@literal null} if there is no shared prefix.
+	 *
+	 * @param associationKey the association key
+	 *
+	 * @return the shared prefix with a trailing dot (.) of the association columns or {@literal null}
+	 */
 	public static String getColumnSharedPrefixOfAssociatedEntityLink(AssociationKey associationKey) {
 		String[] associationKeyColumns = associationKey.getMetadata()
 				.getAssociatedEntityKeyMetadata()
--- a/core/src/main/java/org/hibernate/ogm/datastore/document/impl/DotPatternMapHelpers.java
+++ b/core/src/main/java/org/hibernate/ogm/datastore/document/impl/DotPatternMapHelpers.java
@@ -9,6 +9,7 @@ package org.hibernate.ogm.datastore.document.impl;
 import java.util.Map;
 import java.util.regex.Pattern;
 
+import org.hibernate.ogm.datastore.document.association.spi.impl.DocumentHelpers;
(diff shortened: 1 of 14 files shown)


gunnarmorling:
ok to test

gunnarmorling:
> Hint: I wasn't able to run the full test suite successfully even before the change. Need your advice then.

@mp911de What is the issue with the tests? Any specific error you get?

couchdb/src/test/java/org/hibernate/ogm/datastore/couchdb/test/associations/MapMappingTest.java
```diff
@@ -0,0 +1,374 @@
...
+		Enterprise timberTradingInc = new Enterprise( "enterprise-1", departments );
+
+		session.persist( timberTradingInc );
+		tx.commit();
+		session.clear();
+
+		tx = session.beginTransaction();
+
+		// assert
+		assertDbObject(
```
gunnarmorling: Cool you added this for CouchDB as well. This was really lacking.

core/src/main/java/org/hibernate/ogm/datastore/document/impl/DotPatternMapHelpers.java
```diff
@@ -130,4 +132,15 @@ public static boolean organizeAssociationMapByRowKey(
 	private static MapStorageType getMapStorage(AssociationContext associationContext) {
 		return associationContext.getAssociationTypeContext().getOptionsContext().getUnique( MapStorageOption.class );
 	}
+
+	public static String getColumnSharedPrefixOfAssociatedEntityLink(AssociationKey associationKey) {
```
gunnarmorling: Could you add a description to this method? Also I'd expect it to be used by some of the other dialects? I vaguely remember seeing it somewhere else before.

core/src/main/java/org/hibernate/ogm/datastore/document/impl/DotPatternMapHelpers.java
```diff
@@ -130,4 +132,15 @@ public static boolean organizeAssociationMapByRowKey(
...
+		String[] associationKeyColumns = associationKey.getMetadata()
+				.getAssociatedEntityKeyMetadata()
+				.getAssociationKeyColumns();
+		// we used to check that columns are the same (in an ordered fashion)
```
gunnarmorling: I am not quite following on this comment. Could you elaborate a bit?

mp911de:
@gunnarmorling figured out that it was a problem of my local machine. It said before that the database file could not be found for roundabout 50% of the tests. Disovered also a bug which I'm going to fix with the next commit.

core/src/main/java/org/hibernate/ogm/datastore/document/impl/DotPatternMapHelpers.java
mp911de: that's a c'n'p leftover. I felt more safe to retail the line since I don't get it either.

mp911de:
Commited the changes based on your review. Thanks for reviewing.

gunnarmorling:
Merged; Many thanks, @mp911de!

PR #588, opened 2015-08-31
OGM-145, OGM-914: Exposing type of associations through meta-data; Exposing entity key through association meta-data

gunnarmorling:
@DavideD, @emmanuelbernard: Any thoughts?

@jhalliday: This should give you the meta-data you need for creating the schema properly in C*. Btw. there are convenience methods for getting all association/entity meta-data in BaseSchemaDefiner (I'd like to do some changes to that, but it should not affect you in a general way). Let me know what you think.


gunnarmorling:
Pushed one more commit for improving the way how schema definers can access all the required meta-data for initialising schemas.

@jhalliday With that you should be able to initialise the C* schema as needed.

emmanuelbernard:
I only looked quickly but it looks good. I've added a couple of questions

gunnarmorling:
Pushed one more commit for mapping array type to LIST.

PR #589, opened 2015-09-04
OGM-917 one more test case with EmbeddedId and mongodb

shuraa:
PR contains only test case, no fix for the issue.


gunnarmorling:
Ok to test.

gunnarmorling:
Thanks, @shuraa! Do you have any idea about the actual cause of this?

shuraa:
@gunnarmorling have no idea yet. I'll try to investigate the issue on this week.

shuraa:
Ok, I think, that there is mapping problem. If I have change AuthorId mapping from:

```java
@Embeddable
public class AuthorId implements Serializable {

	private String id;

	public AuthorId(String id) {
		this.id = id;
	}
        
         ....
}
```
to
```java
@Embeddable
public class AuthorId implements Serializable {

	@Column(name = "_id")
	private String id;

	public AuthorId(String id) {
		this.id = id;
	}
        
         ....
}
```
test is passed.

PR #592, opened 2015-10-28
OGM-927 MongoDB CLI logical operator

DavideD:
https://hibernate.atlassian.net/browse/OGM-927

In the end we already support this type of queries, the problem was in the missing quotes.

I've just added some tests and updated the documentation.


gunnarmorling:
Applied; Thanks, @DavideD.

PR #593, opened 2015-11-08
OGM-928 Use Strings in RedisConnection instead of byte array

mp911de:


gunnarmorling:
Merged; Thanks, @mp911de!

PR #594, opened 2015-11-16
OGM-929 Refactor test runners for skip by provider and skip by dialect

mp911de:
* Refactor annotations and testrunner
* Adjust datastore provider lookup within the test to support class names/short names and to deal with type hierarchies
* Add SkipByGridDialect annotation that evaluates the dialect from the configuration

### DIFF
--- a/core/src/test/java/org/hibernate/ogm/utils/SkippableTestRunner.java
+++ b/core/src/test/java/org/hibernate/ogm/utils/SkippableTestRunner.java
@@ -7,6 +7,8 @@
 package org.hibernate.ogm.utils;
 
 import org.hibernate.ogm.datastore.impl.AvailableDatastoreProvider;
+import org.hibernate.ogm.dialect.spi.GridDialect;
+
 import org.junit.runner.Description;
 import org.junit.runner.notification.RunNotifier;
 import org.junit.runners.BlockJUnit4ClassRunner;
@@ -28,7 +30,7 @@ public class SkippableTestRunner extends BlockJUnit4ClassRunner {
 	@Override
 	public void run(RunNotifier notifier) {
 		if ( isTestClassSkipped() || areAllTestMethodsSkipped() ) {
-			notifier.fireTestIgnored( Description.createSuiteDescription( super.getTestClass().getJavaClass() ) );
+			skipTest( notifier );
 		}
 		else {
 			super.run( notifier );
@@ -38,20 +40,32 @@ public class SkippableTestRunner extends BlockJUnit4ClassRunner {
 	@Override
 	protected void runChild(final FrameworkMethod method, RunNotifier notifier) {
 		if ( isTestMethodSkipped( method ) ) {
-			notifier.fireTestIgnored( describeChild( method ) );
+			skipTest( method, notifier );
 		}
 		else {
 			super.runChild( method, notifier );
 		}
 	}
 
-	protected boolean isTestClassSkipped() {
-		return isTestClassSkippedByDialect() || isTestClassSkippedByDatastoreProvider();
+	/**
+	 * Skip a whole test class.
+	 */
+	protected void skipTest(RunNotifier notifier) {
+		notifier.fireTestIgnored( Description.createSuiteDescription( super.getTestClass().getJavaClass() ) );
 	}
 
-	private boolean isTestClassSkippedByDialect() {
-		SkipByGridDialect skipByGridDialect = getTestClass().getJavaClass().getAnnotation( SkipByGridDialect.class );
-		return skipByGridDialect != null ? isSkipped( skipByGridDialect ) : false;
+	/**
+	 * Skip a test method.
+	 */
+	protected void skipTest(FrameworkMethod method, RunNotifier notifier) {
+		notifier.fireTestIgnored( describeChild( method ) );
+	}
+
+	/**
+	 * Whether the test class is skipped by {@link SkipByDatastoreProvider}.
+	 */
+	protected boolean isTestClassSkipped() {
+		return isTestClassSkippedByDatastoreProvider() || isTestClassSkippedByGridDialect();
 	}
 
 	private boolean isTestClassSkippedByDatastoreProvider() {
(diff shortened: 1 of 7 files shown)


core/src/test/java/org/hibernate/ogm/utils/AvailableGridDialect.java
```diff
@@ -0,0 +1,44 @@
...
+package org.hibernate.ogm.utils;
+
+
+/**
+ * The testsuite needs some knowledge on all NoSQL dialects.
+ * We mainly need the name of the GridDialect class.
+ *
+ * @author Mark Paluch
+ */
+public enum AvailableGridDialect {
```
DavideD: Do we really need to define this class?

Can't we just use the Dialect class or class name?

mp911de: The enum is not necessary, but comes handy, you don't have to remember the FQCN all the time. It follows the `GridModule` (former `GridDialectType`) and `AvailableDatastoreProvider` style. 

Referencing the class is not possible because the tests are located in `core` and have no dependencies to the grid modules.

DavideD: > The enum is not necessary, but comes handy, you don't have to remember the FQCN all the time

Right, but since we tend to use <Name provider>Dialect as convetion is not that hard to remember.

>  It follows the GridModule (former GridDialectType) and AvailableDatastoreProvider style.

I'm personally not a big fun of those soution, the fact the we have another place to update in core when we add a datstore or dialect is not a nice thing imo.

> Referencing the class is not possible because the tests are located in core and have no dependencies to the grid modules.

Good point, I always forget about this.

core/src/test/java/org/hibernate/ogm/utils/SkippableTestRunner.java
```diff
@@ -118,4 +174,33 @@ private boolean isSkipped(SkipByDatastoreProvider skipByDatastoreProvider) {
 
 		return false;
 	}
+
+	private boolean isSkipped(SkipByGridDialect skipByGridDialect, SessionFactory sessionFactory) {
+		GridDialect gridDialect = ( (SessionFactoryImplementor) sessionFactory ).getServiceRegistry().getService(
+				GridDialect.class
+		);
+
+		gridDialect = unwrap( gridDialect );
```
DavideD: I think you can use ```GridDialects.getWrappedDialect( gridDialect )``` here

core/src/test/java/org/hibernate/ogm/utils/TestHelper.java
```diff
@@ -124,11 +123,11 @@ public static GridDialect getCurrentGridDialect(DatastoreProvider datastoreProvi
 
 	public static <D extends DatastoreConfiguration<?>> Class<D> getCurrentDatastoreConfiguration() {
 		@SuppressWarnings("unchecked") // relies on the fact that the caller assigns correctly; that's ok for this purpose
-		Class<D> configurationType = (Class<D>) helper.getDatastoreConfigurationType();
+				Class<D> configurationType = (Class<D>) helper.getDatastoreConfigurationType();
```
DavideD: I think this indentation is wrong

core/src/test/java/org/hibernate/ogm/utils/TestHelper.java
```diff
@@ -289,17 +295,37 @@ private static AvailableDatastoreProvider getDatastoreProvider() {
...
+					}
+				}
+
+				Class<?> configuredProviderClass = loadClass( value );
+				if ( configuredProviderClass != null ) {
+
+					for ( AvailableDatastoreProvider provider : AvailableDatastoreProvider.values() ) {
+						Class<?> availableProviderClass = loadClass( provider.getDatastoreProviderClassName() );
+
+						if ( availableProviderClass.isAssignableFrom( configuredProviderClass ) ) {
```
DavideD: How come does it need to be assignable to the available datastore provider? Can't we just instantiate it?

This function is also quite long, I wonder if a one line comment for each block wouldn't make it clearer what's going on.

core/src/test/java/org/hibernate/ogm/utils/AvailableGridDialect.java
DavideD: Just to clarify, no need to change this as it is consistent to what we have.

DavideD:
Thanks @mp911de,

It looks good, just a couple of minor questions. After that I think it can be included.

DavideD:
By the way, we tend to keep the stylistic changes in separated commits from the main issue. This make it easier to review the solution of a problem as there are less changes and potentially less conflicts when merging a pull request.

Non need to change the current pull, though.

core/src/test/java/org/hibernate/ogm/utils/AvailableGridDialect.java
mp911de: This duplications came as well to my mind but wanted to keep that out of scope. I'm easy on whether class name (`String`) or enum.

core/src/test/java/org/hibernate/ogm/utils/SkippableTestRunner.java
mp911de: Thanks, does exactly what I need, just the name is misleading :)

core/src/test/java/org/hibernate/ogm/utils/TestHelper.java
mp911de: I extended the functionality so it's no longer limited to short names but allows exact type matching and to exclude whole hierarchies (subtypes of providers). Will split up the method into parts.

mp911de:
Thanks for reviewing. Fixed the findings and removed the unrelated, reformatted sections from the PR. Also changed `determineGridDialectType()` to `determineGridModule()` to reflect the real purpose. Missed that one in my initial change.

mp911de:
Removed `@SkipByGridModule` annotation and updated the PR.

core/src/test/java/org/hibernate/ogm/backendtck/package-info.java
```diff
@@ -7,7 +7,7 @@
 /**
  * This package contains end-to-end tests which should be executed for every datastore module. If required, single
  * tests can be skipped on specific dialects by annotating them with
- * {@link org.hibernate.ogm.utils.SkipByGridDialect}.
+ * {@link org.hibernate.ogm.utils.SkipByGridModule}.
```
gunnarmorling: This change seems obsolete now. It should mention SkipByGridDialect/SkipByDatastoreProvider instead.

core/src/test/java/org/hibernate/ogm/backendtck/queries/SimpleQueriesTest.java
```diff
@@ -483,7 +485,7 @@ public void testAutoFlushIsAppliedDuringQueryExecution() throws Exception {
 		hypothesis.setPosition( 29 );
 		session.persist( hypothesis );
 
-		if ( EnumSet.of( MONGODB, NEO4J ).contains( TestHelper.getCurrentDialectType() ) ) {
+		if ( EnumSet.of( GridModule.MONGODB, GridModule.NEO4J ).contains( TestHelper.getCurrentModule() ) ) {
```
gunnarmorling: I thought the notion of "module" is not needed really, all tests can either be skipped / controlled by dialect or datastore provider?

core/src/test/java/org/hibernate/ogm/backendtck/associations/collection/manytomany/ManyToManyExtraTest.java
```diff
@@ -22,8 +23,8 @@
 /**
  * @author Emmanuel Bernard &lt;emmanuel@hibernate.org&gt;
  */
-@SkipByGridDialect(
-		value = { GridDialectType.CASSANDRA },
+@SkipByDatastoreProvider(
```
gunnarmorling: Is this change (and others of the same kind) actually needed? There is one datastore provider for Cassandra and one dialect, so it shouldn't matter whether it's skipped by dialect or provider? This should only make a difference where we have several dialects (Redis) or providers (Mongo).

gunnarmorling:
Hey @mp911de, thanks for the update, took another look and added some more comments. Essientally it's not quite clear to me, why some of the changes are actually needed.

As per my understanding it should "only" be required to change the implementation of `@SkipByDialect` to actually consider the current dialect (and not the module where the test is run from) and possibly to sprinkle some `@SkipByDialect` for the Redis backend as per the needs of your new hash dialect.

mp911de:
I will revert the `@SkipByDatastoreProvider` change on the tests and will implement `@SkipByDialect` propertly so it does what it's name says.

mp911de:
I updated the PR by fixing `@SkipByGridDialect` instead of adjusting the tests.

gunnarmorling:
Merged it, @mp911de. Thanks a lot! I've added one more commit to simplify the `GridModule` enum. I am even wondering whether the resolution of the test helper should not go into the dialect enum, but we can leave that for another time.

PR #595, opened 2015-11-16
OGM-930 Mapping of @Lob as String or byte[]

DavideD:
https://hibernate.atlassian.net/browse/OGM-930

### DIFF
--- a/core/src/test/java/org/hibernate/ogm/backendtck/type/Bookmark.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/type/Bookmark.java
@@ -63,9 +63,17 @@ public class Bookmark {
 	@Type(type = "numeric_boolean")
 	private boolean isShared;
 
-	// byte arrays
+	// Lobs
 	@Lob
 	private byte[] lob;
+
+	@Lob
+	private String lobWithString;
+
+	@Lob
+	private Long lobWithLong;
+
+	// byte arrays
 	private byte[] data;
 
 	// enum type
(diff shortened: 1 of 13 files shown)


core/src/test/java/org/hibernate/ogm/backendtck/type/Bookmark.java
```diff
@@ -63,9 +63,14 @@
 	@Type(type = "numeric_boolean")
 	private boolean isShared;
 
-	// byte arrays
+	// Lobs
 	@Lob
 	private byte[] lob;
+
+	@Lob
+	private String lobWithString;
```
gunnarmorling: Could you add a test for other basic types (e.g. long) and an element collection thereof, too?

DavideD:
I've added some additional @Lob mapping and changed the original patch.

I've created a separate issue for List<byte[]> as it seems more complicated that these cases:
https://hibernate.atlassian.net/browse/OGM-932

gunnarmorling:
Could you add this to `BuiltinTypeMappingTest`, too (and potentially its relatives we may have in other backends)?

gunnarmorling:
@DavideD so I am wondering about the mapping of `@Lob` in general. As per the spec, 

> A Lob may be either a binary or character type. The Lob type is inferred from the type of the per- sistent field or property and, except for string and character types, defaults to Blob.

Atm. we e.g. create `NumberLong` field for `@Lob long`, effectively ignoring the specification as LOB. Is that the right approach, as it favors a more natural mapping, or should we actually store it as binary object as requested through usage of the `@Lob` annotation?

DavideD:
Yeah, I guess it make sense to store them as byte[], instead of using the original type

DavideD:
I've updated the patch, now it will store the @Lob in the same way the dialect store the array of byte

gunnarmorling:
Looks good now, thanks, @DavideD! Rebased and applied it. I've opened [OGM-942](https://hibernate.atlassian.net/browse/OGM-942) as a follow-up: `@Lob` doesn't work yet in conjunction with `@ElementCollection` for Cassandra, so we should fix that and tests for it as well.

DavideD:
Thanks @gunnarmorling

PR #596, opened 2015-11-16
[PREVIEW] OGM-366 Neo4j remote

DavideD:

This pull request creates a new Provider and Dialect for Neo4j remote server.
It is possible to test it using:
```
mvn clean install -DdatastoreProvider=neo4j
``` 
The server needs to be configured with username "neo4j" and password "admin" (these can be changed in the hibernate.properties file). 

As discussed, I removed the ThreadLocal and now the dialect can access the TransactionCoordinator from the various Context (TupleContext, AssociationContext and a new QueryContext).

I wonder if we should continue to support the embedded mode since Neo4j is planning to remove it.
Currently, I've kept the previous dialect but it should be possible to reuse one dialect and one provider (for embedded and remote) by creating some interfaces for the execution of queries and getting the results.

This pull request looks quite big but it is not as big as ti looks like, I've split it in several commits, and the first ones are mainly for preparing the road to the new dialect.

So, the first commits will:
* Make sure that the tests in Neo4j read the default configuration from the property file.
* Add tests to check that the sequence increments is thread safe. There is already one but it only tests for the table generation strategy. Eventually we should add this tests in core and make sure that they are executed for the other db as well.
* upgrade to Neo4j 2.6
* Move the DatabaseIdentifier class  in CouchDB to core, so that it can be reused
* Make sure that the AssociationKey default implementation checks the collectionRole in the equals method (forgot the hashcode, but I will update it)

After all those things are in place the patch will:
* Add the classes to map the requests to the server

  All the operation on the server are executed via Cypher. The Neo4j API allows the execution of multiple queries inside the same request and these queries can be executed inside a running transaction or they can be executed immediately (the server will open a new transaction and it will commit it).

* To execute queries inside an open transaction via Rest I need the the transaction Id. I've changed the various context so that they can contain the TransactionCoordinator so that a dialect can get an Id before any operation.

* Add some new GridType converter that are not need for the embedded use case

* Add the classes for the Remote Neo4j datastore and dialect.

The dialect for the embedded Neo4j uses mainly Cypher queries but some parts are still using the programmatic API (property update, embedded association creation, sequences).

For the dialect I had to convert those parts into Cypher queries and it was a bit more complicated than expected. This resulted in two dialects and providers while in theory we don't need that as we can execute Cypher queries  in both backends.

Another difference of the remote dialect is that it creates a certain amount of queries and then executes them with one request to the server while the embedded can execute a query and update a graph element as soon as needed.

In the end, I think the most complicated part is the generation of the Cypher queries used by the dialects, in particular  the one related to the embedded elements.

Cypher is also missing an operator to acquire a lock an a graph element, reading around I've found a way to do it buy changing a property on a node and than updating the same node inside the same transaction. This seems to work.

This is the gist of it. I hope this help as a high level overview and to start asking for question.


gunnarmorling:
Wow, a huge piece of work. Congrats for getting this done! I'll look into it asap.

In the meantime, could you maybe add some high-level descriptions to this PR to things you feel are worth noting? Like the general approach for mapping or something related to the changes in core / the other dialects. Anything what you think may help to get started with reviewing this :)

Looking forward to offer this new dialect to users!

DavideD:
I've added some details to the pull request description. I hope it can help.

DavideD:
I think the `RemoteNeo4jDialect` is probably the best way to start to have on overview of what the new classes are doing

gunnarmorling:
Thanks for adding that description!

Sanne:
@DavideD could you rebase it? there's a micro-conflict in the pom, I could do that myself but then since this is large it will get confusing on were to add comments.

Sanne:
I'm having a deployment failure during the integration tests on WildFly:

     Hibernate OGM Integration Tests for Neo4j .......... FAILURE

Caused by: org.hibernate.service.spi.ServiceException: Unable to create requested service [org.hibernate.ogm.datastore.spi.DatastoreProvider]
    Caused by: org.hibernate.HibernateException: OGM000044: Cannot load class org.hibernate.ogm.datastore.neo4j.impl.Neo4jDatastoreProvider specified via configuration property 'hibernate.ogm.datastore.provider'

DavideD:
@Sanne Rebasing and checking ...

Sanne:
N.B. I had the test failure before rebasing.

Sanne:
No luck?

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/configuration/CassandraConfiguration.java
```diff
@@ -15,7 +15,7 @@
  *
  * @author Jonathan Halliday
  */
-public class CassandraConfiguration extends DocumentStoreConfiguration {
```
Sanne: I'd rather see such changes extracted as a different PR, which we could already apply?

DavideD: Ok, I will send a separate pull requests for this, It should be already a set of separate commit

core/src/main/java/org/hibernate/ogm/dialect/impl/AssociationContextImpl.java
```diff
@@ -25,21 +26,23 @@
...
+	public AssociationContextImpl(AssociationTypeContext associationTypeContext, Tuple entityTuple, TransactionCoordinator transactionCoordinator) {
+		this( associationTypeContext, entityTuple, null, transactionCoordinator );
 	}
 
 	public AssociationContextImpl(AssociationContextImpl original, OperationsQueue operationsQueue) {
-		this( original.associationTypeContext, original.entityTuple, operationsQueue );
+		this( original.associationTypeContext, original.entityTuple, operationsQueue, original.transactionCoordinator );
 	}
 
-	private AssociationContextImpl(AssociationTypeContext associationTypeContext, Tuple entityTuple, OperationsQueue operationsQueue) {
```
Sanne: it seems that the client for getTransactionCoordinator() will throw an NPE is we don't make sure thisnew _transactionCoordinator_ is never null. Add the assertion as the other parameter?

core/src/main/java/org/hibernate/ogm/dialect/query/spi/QueryContext.java
```diff
@@ -0,0 +1,18 @@
...
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.dialect.query.spi;
+
+import org.hibernate.resource.transaction.TransactionCoordinator;
+
+/**
+ * @author Davide D'Alto
+ */
+public interface QueryContext {
```
Sanne: could you add some javadoc here?

core/src/main/java/org/hibernate/ogm/hibernatecore/impl/BackendCustomLoader.java
```diff
@@ -182,8 +184,23 @@ public BackendCustomLoaderContext(QueryableGridDialect<T> gridDialect, BackendCu
...
+			return gridDialect.executeBackendQuery( query, queryParameters, new BackendQueryContext( coordinator ) );
+		}
+	}
+
+	private static class BackendQueryContext implements QueryContext {
+
+		private final TransactionCoordinator coordinator;
+
+		public BackendQueryContext(TransactionCoordinator coordinator) {
+			this.coordinator = coordinator;
```
Sanne: are we sure it's never null at this point?

core/src/main/java/org/hibernate/ogm/dialect/impl/AssociationContextImpl.java
DavideD: THe plan is to remove the TransactionCoordinator form AssociationContext and TupleContext because this objects should contains only the meta information and therefore they shouldn't change.

I will probabaly change these adding TransactionalContext parameter that contains the transactionId

core/src/main/java/org/hibernate/ogm/dialect/impl/ForwardingGridDialect.java
```diff
@@ -181,8 +182,8 @@ public void executeBatch(OperationsQueue queue) {
 	 */
 
 	@Override
-	public ClosableIterator<Tuple> executeBackendQuery(BackendQuery<T> query, QueryParameters queryParameters) {
-		return queryableGridDialect.executeBackendQuery( query, queryParameters );
+	public ClosableIterator<Tuple> executeBackendQuery(BackendQuery<T> query, QueryParameters queryParameters, QueryContext context) {
+		return queryableGridDialect.executeBackendQuery( query, queryParameters, context);
```
Sanne: whitespace?

core/src/main/java/org/hibernate/ogm/dialect/impl/AssociationContextImpl.java
Sanne: Why not simply add an?    Contracts.assertParameterNotNull( ..

core/src/main/java/org/hibernate/ogm/loader/impl/OgmLoader.java
```diff
@@ -617,6 +620,12 @@ private TupleAsMapResultSet getResultSet(Serializable id, QueryParameters qp, Og
 		return resultset;
 	}
 
+	private TupleContext tupleContext(SessionImplementor session, OgmEntityPersister persister) {
+		TupleContext tupleContext = persister.getTupleContext( session );
+		tupleContext = new TupleContextImpl( tupleContext, session.getTransactionCoordinator() );
```
Sanne: This line seems redundant? The method 'persister.getTupleContext( session )' will already do the same, provided the seesion is not null. (and this method will NPE if it is, so I guess we expected it to be not null).

core/src/main/java/org/hibernate/ogm/model/impl/DefaultAssociationKeyMetadata.java
```diff
@@ -289,6 +289,14 @@ public boolean equals(Object o) {
 			return false;
 		}
 
+		if ( collectionRole == null && that.collectionRole != null ) {
+			return false;
+		}
+
+		if ( !collectionRole.equals( that.collectionRole ) ) {
```
Sanne: NPE if collectionRole is null.

core/src/main/java/org/hibernate/ogm/hibernatecore/impl/BackendCustomLoader.java
DavideD: I'm pretty confident :)
I guess it could be null in case a user define a custom transaction coordinator builder.

Anyway, this is not a problem for most of the dialect, but Neo4j Remote needs to check that it is not null and throw a meaningful exception. I will add an assert like you suggested before

core/src/main/java/org/hibernate/ogm/model/impl/DefaultAssociationKeyMetadata.java
DavideD: Good catch

core/src/main/java/org/hibernate/ogm/loader/impl/OgmLoader.java
DavideD: Right, I will remove it

core/src/main/java/org/hibernate/ogm/remote/impl/DatabaseIdentifier.java
```diff
@@ -4,7 +4,7 @@
  * License: GNU Lesser General Public License (LGPL), version 2.1 or later
  * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
  */
-package org.hibernate.ogm.datastore.couchdb.util.impl;
```
Sanne: do we really want this implementation to be shared across all dialects?

If you need something like this, I can see the benefit for an interface but I wouldn't assume that this same implementation is good enough for all other dialects (it's not for Infinispan, eg.).

DavideD: It doesn't have to be shared across all dialect, but it seems to have some methods the could be shared across some dialects.

I think I will avoid this change for now and create different classes. We can eventually share the code later as a different pull request

Sanne: +1

DavideD:
I'll prepare a new pull request

PR #597, opened 2015-11-22
OGM-931 Better exception when association owner is not managed

DavideD:
https://hibernate.atlassian.net/browse/OGM-931

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/persister/impl/EntityAssociationUpdater.java
+++ b/core/src/main/java/org/hibernate/ogm/persister/impl/EntityAssociationUpdater.java
@@ -158,6 +158,10 @@ class EntityAssociationUpdater {
 		}
 		associationPersister.getAssociation().put( rowKey, associationRow );
 
+		if ( associationPersister.getAssociationContext().getEntityTuple() == null ) {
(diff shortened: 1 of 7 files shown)


emmanuelbernard:
I am not sure that is the right way to solve the problem.
It seems that we get a NPE because the Tuple is not cached in the entity extra state. At least that's one of the cause, I don't know if that the actual one here.
If that's the case, we should not fail and rather simply load that needed tuple based ont he metadata.

Maybe I misunderstand the issue and we don't even have the metadata but then I don't get why ORM does not crash.

gunnarmorling:
Ok, I just tested this in ORM, and there this execution suceeds: it will take the id from the transient instance. So the same should be doable in OGM, too, i.e. we should not expect and exception.

ORM's code for obtaining the id is hit with OGM, too, but apparently then we fail to set up the entity tuple with it. Need to dig a bit further, will do so next week.

gunnarmorling:
Ok, going to close this one, I'll send another one.

We need to fetch the entity on the inverse side so e.g. embedded (bi-di) associations can be updated. In ORM, the load naturally is not needed, as associations will only be stored once from the main side.

As a general rule, let's always double check how ORM behaves in certain cases, so we stick to the same behavior as far as possible. Otherwise we'll cause users pain when migrating to OGM.

PR #599, opened 2015-12-03
OGM-894 support cassandra 2.2

jhalliday:
Move to use Cassandra 2.2 driver API and add support for the new 2.2 data types.


jhalliday:
Note CI tests may barf until the cluster is upgraded to run 2.2 server

DavideD:
It looks good to me but I think there two things to do before integrating it:
* Upgrade cassandra on the ci slaves
* Update the cassandra module or the Integration tests will not pass

I'm working on these.

DavideD:
Thanks @jhalliday,
I've included this commit in a pull request that also updates the modules for WildFly.

PR #600, opened 2015-12-04
OGM-873 Redis Hash dialect

mp911de:
* Refactor Redis JSON Strategy to base-class and impl
* Adopt tests
* Implement hash mapping with converting all data types to String


gunnarmorling:
Hey @mp911de, Going to look into this now. Could you rebase it to the latest master in the meantime? Thanks!

core/src/test/java/org/hibernate/ogm/backendtck/optimisticlocking/OptimisticLockingTest.java
```diff
@@ -103,7 +104,7 @@ public void updatingEntityUsingOldVersionCausesException() throws Throwable {
 	 */
 	@Test
 	@SkipByGridDialect(
-			value = { HASHMAP, INFINISPAN, EHCACHE, NEO4J, COUCHDB, CASSANDRA, REDIS },
+			value = { HASHMAP, INFINISPAN, EHCACHE, NEO4J, COUCHDB, CASSANDRA, REDIS, REDIS_HASH },
```
gunnarmorling: Could you add a comment describing why it's excluded for Redis?

integrationtest/redis/src/test/java/org/hibernate/ogm/test/integration/jboss/RedisModuleMemberRegistrationWithTTLConfiguredIT.java
```diff
@@ -33,6 +33,7 @@
 import org.junit.runner.RunWith;
 
 import com.lambdaworks.redis.RedisConnection;
+import com.lambdaworks.redis.codec.Utf8StringCodec;
```
gunnarmorling: Seems not needed?

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisDialect.java
```diff
@@ -6,590 +6,24 @@
...
  * Stores tuples and associations inside Redis.
- * <p>
+ * <p/>
  * Tuples are stored in Redis as a JSON serialization of a {@link Entity} object. Associations are stored in Redis obtained as a
  * JSON serialization of a {@link Association} object either within the entity or external.
  * See {@link org.hibernate.ogm.datastore.document.cfg.DocumentStoreProperties#ASSOCIATIONS_STORE} on how to configure
  * entity or external storage.
  *
  * @author Mark Paluch
+ * @deprecated Use {@link RedisJsonDialect}
```
gunnarmorling: This could be just removed. We'd only need to deprecate it if had been part of a non-preview (alpha, beta etc.) release.

redis/src/main/java/org/hibernate/ogm/datastore/redis/impl/hash/RedisHashEnumType.java
```diff
@@ -0,0 +1,49 @@
...
+import org.hibernate.cfg.NotYetImplementedException;
+import org.hibernate.engine.spi.SessionImplementor;
+import org.hibernate.ogm.model.spi.Tuple;
+import org.hibernate.ogm.type.impl.EnumType;
+import org.hibernate.ogm.util.impl.Log;
+import org.hibernate.ogm.util.impl.LoggerFactory;
+import org.hibernate.type.CustomType;
+
+/**
+ * Enum type that represents the ordinal value as String.
```
gunnarmorling: To be sure, this still supports ordinal *and* string-based mapping of enums, right? It's only that in the ordinal case a string such as "3" would be persisted? If so, I find the comment a bit misleading, as it suggests this *only* writes the ordinal value.

redis/src/test/java/org/hibernate/ogm/datastore/redis/test/associations/MapMappingTest.java
```diff
@@ -32,6 +34,7 @@
  * @author Gunnar Morling
  * @author Mark Paluch
  */
+@SkipByGridDialect(GridDialectType.REDIS_HASH)
```
gunnarmorling: Why is it excluded for HASH?

redis/src/test/java/org/hibernate/ogm/datastore/redis/test/mapping/RedisMappingTest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.junit.Test;
+
+import static org.fest.assertions.Assertions.assertThat;
+
+/**
+ * Test for Redis mapping.
+ *
+ * @author Mark Paluch
+ */
+public class RedisMappingTest extends RedisOgmTestCase {
```
gunnarmorling: Could you add one for asserting the association mappings, too?

gunnarmorling:
When running `ManyToOneTest.testBidirectionalManyToOneRegular()` I see the following in the datastore:

* A `SalesGuy` hash: eric | Eric | sales_force 
* A `SalesGuy` hash: simon | Simon | sales_force 
* An association with the key "Associations:SalesGuy:sales_force", containing the values "eric" and 

So key and values of the association don't seem to match; I'd rather expect the key to contain the name of the table on the inverse side (`SalesForce`) as this is the side which this Redis association represents. But we'd also need to encode the role of the collection, because there could be several associations between the two entities (e.g. `salesGuys` and `outstandingSalesGuys`) and the current structure would not allow to restore to which association an element belongs.

gunnarmorling:
For one-to-one associations (e.g. see `OneToOneTest.testBidirectionalOneToOne()`) it'd be nice if we could avoid the assocation hash and instead store the associated id in the entity hash on the respective other side. E.g. we already do it for MongoDB. The trick is that this association is passed to `insertOrUpdateAssociation()`, but one can find out from the passed meta-data that it is an one-to-one actually.

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisHashDialect.java
```diff
@@ -0,0 +1,282 @@
...
+		Map<String, Object> map = ( (RedisTupleSnapshot) tuple.getSnapshot() ).getMap();
+		Map<String, String> entity = new HashMap<>();
+		List<String> toDelete1 = new ArrayList<>();
+
+		for ( TupleOperation action : tuple.getOperations() ) {
+			switch ( action.getType() ) {
+				case PUT:
+					map.put( action.getColumn(), action.getValue() );
+
+					// TODO: IllegalStateException when value is not String/Character?
```
gunnarmorling: Can it actually happen that it is something different?

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisHashDialect.java
```diff
@@ -0,0 +1,282 @@
...
+					map.remove( action.getColumn() );
+					toDelete1.add( action.getColumn() );
+					break;
+			}
+		}
+
+		String entityId = entityId( key );
+		Long currentTtl = connection.pttl( entityId( key ) );
+
+		if ( !toDelete1.isEmpty() ) {
```
gunnarmorling: The numbered variable names are a bit tough to grok. Maybe you can find something more descritive? Or just `toDelete` and inline the second one.

gunnarmorling:
@mp911de Did a first pass, looks great overall. I think some of the mappings need a bit more fine-tuning (see my comments inline), and we'd also need an update to the documentation.

Thanks a lot, it's much appreciated!

core/src/test/java/org/hibernate/ogm/backendtck/optimisticlocking/OptimisticLockingTest.java
mp911de: Redis/Redis Hash does not implement `OptimisticLockingAwareGridDialect`, there's no `UPDATE ... WHERE ...` command.

mp911de:
About `ManyToOneTest.testBidirectionalManyToOneRegular`: For single-column associations the pattern is `"Associations:" + <tableName> + ":" + keyColumns[0]`. The role is not yet part of the association key. I'm not sure where you're heading, let's have a chat about that topic.

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisHashDialect.java
mp911de: Whoops. Shame on me.

redis/src/main/java/org/hibernate/ogm/datastore/redis/RedisHashDialect.java
mp911de: Yep. The `NumericBoolean` or the `EnumType`  (or any other custom types) can produce types other than `String`. This place would be a good fit to catch those values and raise a better exception than `ClassCastException` in a later stage. WDYT?

gunnarmorling:
> I'm not sure where you're heading, let's have a chat about that topic.

+1 Just ping me if you have some time.

redis/src/test/java/org/hibernate/ogm/datastore/redis/test/associations/MapMappingTest.java
mp911de: Because embedded maps are not supported like it works with JSON. Agree to externalize maps with a key of `Table:id:collectionRole`.

mp911de:
@mp911de Use as table name for associations AKM->getEntityKeyMetadata()->getTable()

gunnarmorling:
@mp911de, on a related note, you might map `Set`-typed associationst to have Redis sets (if that's not the case yet). You can find out about the nature of an association through `AssociationKeyMetadata#getAssociationType()`.

How do you deal with ordered collections btw.?

mp911de:
Rebased and implemented your feedback. The only open point from my side is the bidirectional one-to-one mapping. Currently two data structures are needed. The association is stored on the owner side and within an association list to reflect the `SELECT (owner) FROM (table) WHERE FK = FK_ID`. A structured search over hashes in a `WHERE`/filter-style is not possible.

gunnarmorling:
Hey @mp911de, is it good to go now from your side?

core/src/test/java/org/hibernate/ogm/backendtck/associations/onetoone/Offender.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.persistence.FetchType;
+import javax.persistence.JoinColumn;
+import javax.persistence.JoinColumns;
+import javax.persistence.OneToOne;
+
+/**
+ * @author Mark Paluch
+ */
+@Entity
+public class Offender {
```
gunnarmorling: Gee, what sort of example ;)

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
```diff
@@ -36,6 +38,8 @@ The following properties are available to configure Redis support in Hibernate O
 .Redis datastore configuration properties
 hibernate.ogm.datastore.provider::
 To use Redis as a datastore provider, this property must be set to `redis_experimental`
+hibernate.ogm.datastore.grid_dialect::
+The dialect type how Hibernate OGM Redis maps entities and associations within Redis. The following two dialects exist: `org.hibernate.ogm.datastore.redis.RedisJsonDialect` (store entities as JSON documents within Redis keys. Associations are mapped to Redis lists/sets) and `org.hibernate.ogm.datastore.redis.RedisHashDialect` (store entities as key-value pairs within Redis hashes. Associations are mapped as JSON documents to Redis lists/sets)
```
gunnarmorling: > Associations are mapped as JSON documents to Redis lists/sets

Is that true actually? Isn't this only the case for composite keys?

Btw. I recommend to just have one sentence per line. This keeps lines shorter and easier to diff down the road.

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
```diff
@@ -382,236 +393,356 @@ so you can name the entity id like you want.
...
+    private String owner;
+
+    // getters, setters ...
+}
+----
 
 [source, JSON]
 ----
-  News:42
+Key: News:939c892d-1129-4aff-abf8-e6c26e59dcb
```
gunnarmorling: Maybe also add id and title from the outer entity to avoid confusion?

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
```diff
@@ -382,236 +393,356 @@ so you can name the entity id like you want.
...
+
+[NOTE]
 ====
-[source, JAVA]
+Hibernate OGM doesn't store null values in Redis,
+setting a value to null will be the same as removing the field
+in the corresponding object in the db.
+====
+
+Hibernate OGM support by default the following types:
```
gunnarmorling: supports

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
```diff
@@ -717,10 +845,9 @@ public class GrandChild {
 
 [source, JSON]
 ----
-Key: GrandMother:86ada718-f2a2-4299-b6ac-3d90b1ef2331
-{
-    "grandChildren" : [ "Luke", "Leia" ]
-}
+Key: Associations:GrandMother_grandChildren:86ada718-f2a2-4299-b6ac-3d90b1ef2331:grandChildren
```
gunnarmorling: Is it indeed the association table name ("GrandMother_grandChildren") or actually the entity table name ("GrandMother")? IIRC, we meant to go for the latter?

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
```diff
@@ -1900,6 +2027,254 @@ Key: Association:AccountOwner:BankAccount:owners:account_1
...
+
+Two main strategies are supported:
+
+1. <<redis-table-id-generation-strategy, TABLE>>
+2. <<redis-sequence-id-generation-strategy, SEQUENCE>>
+
+Both strategy will operate in the keys starting with `Identifiers` containing the last value of the id. The difference
+between the two strategies is the name of the key containing the values.
+
+Hibernate OGM goes not support the `IDENTITY` strategy and an exception is thrown at startup
```
gunnarmorling: That's not generally true; E.g. it is supported for MongoDB (maps to server-generated object ids in that case).

redis/pom.xml
```diff
@@ -154,18 +154,43 @@
...
                         <REDIS_PORT>${env.REDIS_PORT}</REDIS_PORT>
                     </environmentVariables>
                     <forkMode>once</forkMode>
-
                     <dependenciesToScan>
                         <dependency>org.hibernate.ogm:hibernate-ogm-core</dependency>
                     </dependenciesToScan>
-
                 </configuration>
+                <executions>
```
gunnarmorling: Nice!

redis/src/test/java/org/hibernate/ogm/datastore/redis/test/options/ttl/RedisTTLTest.java
```diff
@@ -78,7 +78,7 @@ public void ttlOnMappings() {
 
 		// when
 		Long bandTTL = getConnection().pttl( "Band:1" );
-		Long songTTL = getConnection().pttl( "Associations:Band:1" );
+		Long songTTL = getConnection().pttl( "Associations:Band_Song:1:songs" );
```
gunnarmorling: Hm, so apparently it is the association table name.

gunnarmorling:
@mp911de Added some more comments. I think we are getting there :)

On a general note I am getting doubts about the way associations are mapped for the hash dialect. Maybe we should go just for the table approach as we do for ISPN/Ehcache. But the native Redis lists felt temping. Anyways, I think we can go with the current approach (minus my one question about the used table name) and wait for feedback. It's marked as experimental after all.

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
mp911de: We agreed on using the association table name.

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
mp911de: Adjusted this section.

redis/src/test/java/org/hibernate/ogm/datastore/redis/test/options/ttl/RedisTTLTest.java
mp911de: Yup.

gunnarmorling:
Hey @mp911de, just rebased and applied this. Thanks a lot for this work! I'll open a follow-up issue for looking into the association mapping again, but I think what we have is a good starter. It's marked experimental, so we could e.g. revisit when getting some actual feedback.

mp911de:
Woohoo! Thanks!

gunnarmorling:
Yeah, miracles happen :)

Sorry again for the long delay, I was heads-down into some other stuff. So
thanks for your patience, too. Any idea of what you'd like to tackle next?

PR #601, opened 2015-12-11
Upgrades

Sanne:
- https://hibernate.atlassian.net/browse/OGM-935
 - https://hibernate.atlassian.net/browse/OGM-936
 - https://hibernate.atlassian.net/browse/OGM-937
 - https://hibernate.atlassian.net/browse/OGM-938
 - https://hibernate.atlassian.net/browse/OGM-939

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -44,6 +44,7 @@
         <hibernateParserVersion>1.2.0.Final</hibernateParserVersion>
         <hibernateCommonsAnnotationsVersion>5.0.1.Final</hibernateCommonsAnnotationsVersion>
         <jbossjtaVersion>4.16.4.Final</jbossjtaVersion>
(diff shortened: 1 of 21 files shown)


gunnarmorling:
Nice, thanks! But you won't get extra points for opening the most micro JIRA issues ;)

Sanne:
I understand, but the point of changelogs and JIRA is to track changes so a single "updated it all" doesn't cut it ;)

Thanks for the quick merge!
Now with Infinispan 8 I have no more excuses to delay my real task...

gunnarmorling:
It's not that the issue title needs to convey it all, there is a description field, too ;)

But anyways, thanks for updating all. I always find it a bit tedious to find the right versions, would be nice to get warned in a way when building with different lib versions than provided by the target WF version.

PR #602, opened 2015-12-14
OGM-894 Support cassandra 22

DavideD:
Follows up for: https://github.com/hibernate/hibernate-ogm/pull/599

I've just updated the modules for wildfly, I've aligned the netty libraries version to the in Redis. It seems to work, otherwise I'm not sure how in the modules project we can inlcude two different versions of the same libraries.

I've also updated the version of cassandra on CI, so this pull request should pass all the tests.


gunnarmorling:
Merged; Thanks, @DavideD.

It'd be good if we could re-use the Netty modules provided by the container for the sake of download size but also avoiding redundant classes in the container environment. There is [OGM-890](https://hibernate.atlassian.net/browse/OGM-890) for this (affects Redis and Cassandra), and the versions seem compatible.

PR #603, opened 2015-12-15
OGM-890 Use WildFly Netty module

DavideD:
https://hibernate.atlassian.net/browse/OGM-890

The second commit align the version of Netty for Cassandra and Redis to the one in WildFly.
It is not required but I think it's good to run the unit tests using the same version.


gunnarmorling:
Rebased and applied. Thanks!

PR #604, opened 2015-12-15
OGM-907 Remove test for duplicated keys

DavideD:
Follows up for https://github.com/hibernate/hibernate-ogm/pull/585

### DIFF
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
@@ -191,6 +191,8 @@ public class MongoDBDialect extends BaseGridDialect implements QueryableGridDial
 			for ( int i = 0; i < searchObjects.length; i++ ) {
 				if ( dbObject.get( ID_FIELDNAME ).equals( searchObjects[i] ) ) {
 					tuples.set( i, createTuple( keys[i], tupleContext, dbObject ) );
+					// We assume there are no duplicated keys
+					break;
 				}
 			}
 		}
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
@@ -12,6 +12,7 @@ import static org.hibernate.ogm.datastore.mongodb.dialect.impl.MongoDBTupleSnaps
 import static org.hibernate.ogm.datastore.mongodb.dialect.impl.MongoHelpers.hasField;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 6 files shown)


mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -191,6 +191,8 @@ public Tuple getTuple(EntityKey key, TupleContext tupleContext) {
 			for ( int i = 0; i < searchObjects.length; i++ ) {
 				if ( dbObject.get( ID_FIELDNAME ).equals( searchObjects[i] ) ) {
 					tuples.set( i, createTuple( keys[i], tupleContext, dbObject ) );
```
gunnarmorling: Thinking about it, should we better consume the cursor into a map and then drive result set creating by iterating the requested keys and looking up the corresponding result.

In the current approach we'll have n * n/2 loop iterations on average (it used to be n * n, so it's better already), an n can go up to 1000 as per @emmanuelbernard. Using a temporary map, it'd be n + n (once iterating the cursor, once the incoming keys), at the cost of the map allocation. Of course we should not optimize prematurely, but the n * n growth is scary. @Sanne, any thoughts?

Sanne: Do we know for sure it's limited to "up to a 1000" or is that just the expected use case?

I like your idea. Hibernate Search does use maps and a two-phase iteration approach, as an answer to user reported (real street) issues.

What I find more amusing is the unneccessary N loop introduced by using 

     CollectionHelper.initializeSizedList( searchObjects.length, null );

that's not premature optimisation, that's just pointless overhead.

emmanuelbernard: 1000 is a reasonable guess. BTW you could decide to have two code branch depending on n.

Sanne: I think rather than using a Map I'd see if one can have the objects returned in the order we need them?

gunnarmorling: > I think rather than using a Map I'd see if one can have the objects returned in the order we need them?

Yes, sure, that'd be much better and was in fact my original idea as well. It seems not possible with the current datastore versions, though. E.g. MongoDB does not obey the requested order for a $IN predicate.

Sanne: > BTW you could decide to have two code branch depending on n.

+1

gunnarmorling: > that's not premature optimisation, that's just pointless overhead.

Yes, that could go away with the map approach. To be fair it was needed in the current approach, driven by the actual results. But yet another reason to change it...

DavideD: > that's not premature optimisation, that's just pointless overhead.

It's needed for this approach. Out of curiosity, would it be better to use an array instead of list in this case? Something like:
```
Tuple[] tuples = new Tuple[ keys.length];
...
return Arrays.asList( tuples );
```

Sanne: yes, use an array. no need to set it all to null as it's being done already.

DavideD:
I added a commit that uses an array instead of a List

gunnarmorling:
Ok, but why not the map approach? For 1000 ids, ~500.000 iterations vs. (2000 iterations + 1000 map look ups) is a huge difference.

Sanne:
+1 for the map approach, if the composite cardinality is beyond some trivial size (to be defined with some tests or a lazy rule of thumb?)

Although, don't underestimate the cost of ~1000 map lookups and writes. Maps are bloody slow, especially compared to arrays.

DavideD:
> +1 for the map approach, if the composite cardinality is beyond some trivial size (to be defined with some tests or a lazy rule of thumb?)

I would feel more confident with a seprate tests. Couldn't we treat this performance optimization as a separate issue? The reason I've sent this pull request was to remove the wrong assumption (and thus the test) about the duplicated keys.

Sanne:
+1 for a separate issue. Let's merge this, it's already faster than the original code ;)

Sanne:
merged

PR #606, opened 2015-12-15
OGM-931 Allowing to persist association to unmanaged entity on inverse side

gunnarmorling:
Supersedes #597: No exception anymore, instead the associated entity is loaded to make the same flow of execution work as in ORM.


Sanne:
Shouldn't we squash things together when there are serious changes in direction, like the entityname typo being fixed and the actual goal of the test changing?

That was tricky to follow, but looks good. Merged.

gunnarmorling:
Good point, I probably would have squashed it if it was completely my work, but I am always a bit reluctant of squashing commits by others into mine; Maybe I should have done so and committed Davide's id ;)

PR #607, opened 2015-12-16
OGM-917 Supporting embeddable key with single column

gunnarmorling:


Sanne:
boooo booooo ! it's failing checkstyle:D

gunnarmorling:
Wot?! Can't believe it ;)

2015-12-16 14:33 GMT+01:00 Sanne Grinovero <notifications@github.com>:

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/mapping/EmbeddableIdMappingTest.java
```diff
@@ -0,0 +1,117 @@
...
+	@Before
+	public void setUpTestData() {
+		OgmSession session = openSession();
+		Transaction transaction = session.beginTransaction();
+
+		session.persist( new SingleBoardComputer( new SingleBoardComputerPk( "sbc-1" ), "Raspberry Pi" ) );
+
+		session.persist(
+				new News(
+						new NewsID( "seven-projects", "Jerry" ),
```
Sanne: Why is this test which also uses a dual-property embedded key specific to MongoDB only?

gunnarmorling: It's a test for the specific persistent mapping, here we are asserting the
MongoDB-specific JSON representation.

Sanne:
merged after biting of some whitespace

PR #609, opened 2015-12-23
OGM-948 Update Redis driver library to 4.0.2.Final

mp911de:


gunnarmorling:
Merged; thanks, @mp911de! Added one more commit to make sure we add the right Guava version for the Lettuce driver, too.

PR #610, opened 2016-01-04
OGM-897 cache cassandra PreparedStatements

jhalliday:


gunnarmorling:
Jenkins, retest this, please.

gunnarmorling:
Jenkins, retest this, please.

gunnarmorling:
@jhalliday There seems to be a classpath issue

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
```diff
@@ -62,6 +63,9 @@
 import com.datastax.driver.core.querybuilder.Insert;
 import com.datastax.driver.core.querybuilder.QueryBuilder;
 import com.datastax.driver.core.querybuilder.Select;
+import com.google.common.cache.CacheBuilder;
```
gunnarmorling: Interesting, so far we refrained from using Guava in (non-test) code to avoid conflicts with versions of that super-popular library used by users themselves. But given that the C* driver itself uses Guava, I suppose that argument is moot in this case.

@emmanuelbernard, any thoughts?

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
```diff
@@ -96,8 +112,17 @@ public LockingStrategy getLockingStrategy(Lockable lockable, LockMode lockMode)
 
 	private ResultSet bindAndExecute(Object[] columnValues, RegularStatement statement) {
 
+		PreparedStatement preparedStatement;
+		try {
+			preparedStatement = preparedStatementCache.get( statement.getQueryString() );
+			session.prepare( statement );
+		}
+		catch (ExecutionException e) {
+			log.failToPrepareCQL( statement.getQueryString(), e.getCause() );
```
gunnarmorling: I'd prefer to either log *or* re-throw. Best IMO would be to throw a `HibernateException`, wrapping the original one and containing the query, thrown via our common `Log` class pattern (so it gets a unique code, too).

gunnarmorling: Ah, just seeing you already declare the required log method, only the exception returned by it is ignored. It should be 

    throw log.failToPrepareCQL( statement.getQueryString(), e.getCause() );

gunnarmorling:
Logs good to me, apart from the two questions/remarks inline.

gunnarmorling:
Rebased and applied. Thanks, @jhalliday.

PR #611, opened 2016-01-06
OGM-952 implement JPA AttributeConverter type

emmanuelbernard:


emmanuelbernard:
Note that this code fails for Neo4J and couchDb due to tuple already exists exception of sort. I suspect this is a side effect of testing UUID generation and not the feature per se.

Someone could change the `Printer.id` mapping to make it more classic and check it out. That mean we have some bug somewhere though.

core/src/main/java/org/hibernate/ogm/type/impl/TypeTranslatorImpl.java
```diff
@@ -121,4 +135,40 @@ else if ( type instanceof org.hibernate.type.CollectionType ) {
 		}
 		throw log.unableToFindGridType( type.getClass().getName() );
 	}
+
+	/*
+	   We directly pass the converter "database" type as value to the Tuple
```
gunnarmorling: I'll open up a follow-up JIRA for this one.

Would we even need that registry you describe or couldn't we "just" take the target type from the converter (`UUID` in your example), find the right ORM `Type` for this and pass it to `GridDialect#overrideType()`? We could then feed the converted value through that grid type before writing it.

gunnarmorling:
@emmanuelbernard This does not yet take care of converter application for query parameters, right?

gunnarmorling:
> Note that this code fails for Neo4J and couchDb due to tuple already exists exception of sort. I suspect this is a side effect of testing UUID generation and not the feature per se.

Actually it is related to the feature. But it also reveals an issue in the existing code :) 

The problem is that ORM issues an INSERT *and* an UPDATE for this persist. This happens because during dirty checking upon flush the `MyString` typed property is considered as dirty. That's because `MyString` doesn't implement `hashCode()`/`equals()` but also because the uper-case/lower-case cycle isn't necessarily symmetric ("FooBar" will be "foobar" after upper-casing + lower-casing).

So the test passes if I add the `equals()`/`hashCode()` methods and change the original value to "foobar"; Then ORM will not consider the property as dirty and only the INSERT will be executed.

That all being said, the test should also pass with the INSERT+UPDATE sequence, that indeed seems to be an error within these dialects. Looking into it.

core/src/main/java/org/hibernate/ogm/type/impl/TypeTranslatorImpl.java
emmanuelbernard: Ah, right we could indeed reuse ORM logic to do Java type -> ORM Type and then do our regular Type -> GridType.
I don't even think we need anything specific, we could do it today.

(I'm glad I named it GridType and not Type, that would be massively confusing ;) )

emmanuelbernard: Actually, is this really the behavior we want ?
Don't we rather want the user to have final say on how the type is transformed?

I guess it is always fine to pass the guessed Type into the `TypeTranslatorImpl.getType()` again. @gunnarmorling can you see downsides where it would be inappropriate?

gunnarmorling: To me it'd feel right if the target type of the converter is used as input for finding the right grid type.

> Don't we rather want the user to have final say on how the type is transformed?

Yes, but that's subject to custom (grid) types, right?

So I think I see attribute converters more as "logical" converters which e.g. change the representation or do some "formatting", whereas grid types do the actual physical conversion into the column type(s) as understood by individual dialects. E.g. one couldn't write several columns via an attribute converter.

Or, in other words, attribute converters are specific to the (business) attribute types of an application. So attribute converters would always emit one of the basic attribute types defined by JPA, whereas grid types make sure that all these potential basic attribute types are mapped into corresponding column types supported by a store.

Or am I actually having a wrong interpretation of the `AttributeConverter` contract?

emmanuelbernard:
@gunnarmorling I've added 3 commits related to the support for AttributeConverter and the use of the intermediary grid type as we expect. But it does not work for Neo4JDialect as the move to TypeTranslator to a SessionFactory Service will make them fail.

I'm at the end of the rope time wise, if you could take a look @gunnarmorling, that would be cool.

gunnarmorling:
Yes, I can take over. I already fixed the `TupleAlreadyExistsException` for Neo4j locally, for CouchDB it's a bit tougher. Not sure yet what to do regarding the service mess...

gunnarmorling:
Hum, so Neo4j uses the `TypeTranslator` to convert query parameter values into grid types. That should be done on the engine side anyways to have all dialects benefit from it, but it requires a change to the `GridDialect` contract. Good that we are up for a new major release :)

emmanuelbernard:
@gunnarmorling about the service, I think replacing `getType(Type)` by `getType(Type, TypeResolver)` is viable.
Looking at all the call sites, they all can access `SessionFactory.getTypeResolver()` as they have `Session`, `Lockable`, `PersistenceCallerContext`, etc.
The trickiest one is `ComponentType` but it's created by `TypeTranslator` so we are good.

That's a bit fugly to have to pass it all the time but at least that's a viable solution.

core/src/test/java/org/hibernate/ogm/backendtck/type/custom/JpaConvertCustomTypeTest.java
```diff
@@ -0,0 +1,171 @@
...
+import org.hibernate.ogm.OgmSessionFactory;
+import org.hibernate.ogm.cfg.OgmConfiguration;
+import org.hibernate.ogm.utils.OgmTestCase;
+import org.hibernate.type.descriptor.java.UrlTypeDescriptor;
+
+import static org.fest.assertions.Assertions.assertThat;
+
+/**
+ * Test the JPA @Convert logic in OGM
+ * TODO: should it be in the TCK, that's core code but it's nice to see it tested across all backends
```
gunnarmorling: +1 for having it in the TCK. That way we can e.g. make sure that the two-step conversion works on Redis' hash dialect, which *only* knows of Strings.

core/src/test/java/org/hibernate/ogm/backendtck/type/custom/JpaConvertCustomTypeTest.java
```diff
@@ -0,0 +1,171 @@
...
+		assertThat( printer ).isNotNull();
+		assertThat( printer.name.toString() ).isEqualTo( "somefoo" );
+		session.delete( printer );
+		session.getTransaction().commit();
+
+		session.close();
+	}
+
+	@Test
+	public void testGridTypeForIntermediaryTypeIsUsed() throws Exception {
```
gunnarmorling: So how does that test make sure that the intermediary type is actually used?

gunnarmorling:
Closing; Superseded by #612.

core/src/test/java/org/hibernate/ogm/backendtck/type/custom/JpaConvertCustomTypeTest.java
emmanuelbernard: The next one with an unsupported type does. But this test was the first one I wrote especially to verify that the mechanic worked. I am not certain if its intrinsic value anymore.

PR #613, opened 2016-01-08
OGM-909 support cassandra native CQL queries

jhalliday:


cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/test/query/nativequery/CassandraEntityManagerNativeQueryTest.java
```diff
@@ -0,0 +1,252 @@
...
+		delete( portia, athanasia, critic );
+		commit();
+		close( em );
+	}
+
+	@Test
+	public void testSingleResultQuery() throws Exception {
+		begin();
+
+		String nativeQuery = "SELECT * FROM \"WILDE_POEM\" WHERE name='Portia'";
```
gunnarmorling: Why is it again that you apply the quoting by default? It's the only dialect doing it, and I feel we need a more general solution, based on the keyword handling and (auto-)quoting mechanisms provided by ORM core (we could override the list of reserved keywords through grid dialects).

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/query/impl/CassandraParameterMetadataBuilder.java
```diff
@@ -0,0 +1,69 @@
...
+import org.hibernate.engine.query.spi.OrdinalParameterDescriptor;
+import org.hibernate.engine.query.spi.ParameterMetadata;
+import org.hibernate.mapping.Column;
+import org.hibernate.mapping.Table;
+import org.hibernate.ogm.dialect.query.spi.ParameterMetadataBuilder;
+import org.hibernate.type.Type;
+
+/**
+ * {@link ParameterMetadataBuilder} for native Cassandra CQL queries.
+ * This implementation delegates parsing to the server cluster, so no string positional information.
```
gunnarmorling: The [C* docs](https://datastax.github.io/java-driver/2.2.0-rc3/features/statements/prepared/) mention named parameter support in "native protocol v2"; So does this mean our driver is not using this protocol?

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
```diff
@@ -468,4 +475,38 @@ public boolean supportsSequences() {
...
+	@Override
+	public ParameterMetadataBuilder getParameterMetadataBuilder() {
+		return new CassandraParameterMetadataBuilder( session, provider.getMetaDataCache() );
+	}
+
+	@Override
+	public String parseNativeQuery(String nativeQuery) {
+		// we defer the work, since at this point the table may not yet exist in the db and without
+		// the db server supplied metadata or parsing assistance we can't do much meaningful validation.
+		return nativeQuery;
```
gunnarmorling: That's interesting, it indicates that something is wrong with the timing of schema initialization.

Simply deferring renders start-up validation of named queries moot, rather I think we should fix that ordering issue. I've filed [OGM-955](https://hibernate.atlassian.net/browse/OGM-955) and will look into it.

cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/test/query/nativequery/CassandraEntityManagerNativeQueryTest.java
jhalliday: IIRC CASSANDRA-8393 was the straw that broke the camel's back, but in general c* automatically lower cases anything not quoted, whilst the driver's internals don't. Because we're having to play funny games with concatenating identifiers, it's also possible that something which starts out looking innocuous then winds up containing metacharacters and thus requiring quoting. It's basically a lot simpler to quote everything for safety than to derive a foolproof set of rules for when quoting is required.  Maybe open a JIRA for it and we'll revisit if it gets enough votes?

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/query/impl/CassandraParameterMetadataBuilder.java
jhalliday: It is, but how is that relevant?

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
jhalliday: Agreed.  Merge as is, modify after OGM-955? Or wait?

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/query/impl/CassandraParameterMetadataBuilder.java
gunnarmorling: The comment says "no string positional information", but it seems we actually could expose named parameters in addition to index based parameters?

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
gunnarmorling: I'd say let's revisit after OGM-955 is done.

cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/test/query/nativequery/CassandraEntityManagerNativeQueryTest.java
gunnarmorling: Ok, I'll open one.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/query/impl/CassandraParameterMetadataBuilder.java
```diff
@@ -0,0 +1,69 @@
...
+		if ( columnDefinitions.size() > 0 ) {
+
+			// the cassandra metadata will give us the CQL type, but the type conversion system only goes
+			// in hibernate->cassandra direction, so we can't turn it back into the required hibernate type.
+			// instead we rely on the cached hibernate metdata from schema creation time
+
+			String tableName = columnDefinitions.getTable( 0 );
+			Table table = metaDataCache.get( tableName );
+
+			for ( ColumnDefinitions.Definition definition : columnDefinitions ) {
```
gunnarmorling: Just fully getting this, it's very neat. It should spare the user from passing the Hibernate type via `setParameter()`.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraSchemaDefiner.java
```diff
@@ -101,18 +102,36 @@ private void processIndexes(CassandraDatastoreProvider datastoreProvider, Table
...
+
+
+		Iterator<Index> indexIterator = table.getIndexIterator();
+		while ( indexIterator.hasNext() ) {
+			Index index = indexIterator.next();
+
+			List<String> columnNames = extractColumnNames( index.getColumnIterator() );
+
+			// cassandra won't allow single index on multiple cols, so index first col only.
+			if ( !primaryKeys.contains( columnNames.get( 0 ) ) ) {
```
gunnarmorling: We should raise a warning if a multi-column index is configured which cannot be mapped.

gunnarmorling:
Rebased and applied. Thanks, @jhalliday! I'll send a follow-up for the index warning I mentioned before.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/query/impl/CassandraParameterMetadataBuilder.java
jhalliday: confusing two things here maybe?. string positional info means 'character offset withing string', used for parse error reporting, which isn't available because the parse is server side rather than client side. That's different to 'ordinal position of parameter within array' for bind purposes, which is a whole other problem.

PR #614, opened 2016-01-08
OGM-910 added Single Field Indexes support for MongoDB

francoisledroff:
* based on a new ogm `OgmIndex` meta-annotated Mongo `Indexed` field annotation
* the `OgmEntityPersister` will parse these new annotations and trigger `GridDialect` and `BaseSchemaDefiner` index related activities
* added index related method to `GridDialect` : `createIndex` and `getIndexSpec`
* the index are validated and created at bootstrap through `MongoDBEntityMappingValidator` `BaseSchemaDefiner` (hence only once)

This is still in progress
* lots of `TODO`s (among them proper index spec validation)
* lots of missing unit-tests
* check-style validation and missing headers


francoisledroff:
This is just a first draft

If you agree with the path, I'll polish it: add unit tests and finish all the `TODO`s.

FYI: This implementation only target Single Field Indexes

It will not support:
* Compound Indexes
* Multikey Indexes
* Geospatial Indexes and Queries
* Text Indexes
* Hashed Index

we could then discuss what it would take to support the above too.

gunnarmorling:
Hey @francoisledroff, thanks a lot!

What is the reasoning behind that meta-annotation approach and the introduction of the MongoDB-specific annotation? Did you consider to base this on the existing JPA 2.1 construct `@Table#indexes()`?

Generally, we try to map/re-use existing JPA concepts as far as possible, so here `@Table#indexes()` would be the natural candidate. Stuff which isn't mappable is flagged during bootstrap (so e.g. here we'd have to flag multi-column indexes if not supported by a certain backend). The rationale being, that existing ORM apps can be migrated more easily and users have one way to achieve a certain thing across the board.

Then we can amend this by store-specific options/settings, e.g. for the time-to-life stuff. There is a pluggable option system in OGM which allows the user to configure such settings via annotations and/or in a fluent API style. This is e.g. used for the TTL handling in the Redis dialect (look for `@TTL`, `RedisEntityContext.` etc.). The same could be done for MongoDB, e.g. to add specifics such as hash-based indexes and what else may be needed.

We could use this to amend the index configuration done in the standard JPA way for exposing MongoDB-specific settings as needed:

    @Table(indexes={@Index(name="idx_1", columnList="col1 ASC, col2 DESC", unique=true)})
    @MongoDBIndex(forIndex="idx_1", hashed=true)
    @Entity
    public class MyIndexedEntity {}

This resembles a bit the approach for field configuration in HSEARCH via `@Field` + `@NumericField`. For the programmatic style it could be done even nicer, as the store-specific configuration context interfaces may simply expose additional settings, usable with inherited standard settings in a fluent manner.

WDYT?

core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
```diff
@@ -526,6 +543,47 @@ private TupleContext createTupleContext() {
...
+	{
+		gridDialect.createIndex(indexSpec);
+	}
+
+	public List<OgmIndexSpec> getIndexSpec()
+	{
+		return gridDialect.getIndexSpec(tableName,indexAnnotations);
+	}
+
+	private Map<String, Annotation> getOgmIndexAnnotations(Class cls) {
```
gunnarmorling: The persister is not the right place to deal with annotation retrieval. It works on meta-data retrieved and validated in previous bootstrap phases.

Assuming we go the path I suggested and re-use `@Table#indexes()`, schema definers would be able to retrieve the configured indexes from the passed `Database` objects and its table mappings. That way also XML-based mappings will be applied and it's abstracted wether fields/or getters have been annotated.

For store-specific settings (should we need them), there is `OptionsService` which exposes these settings (again in a unified manner, so it's the same handling for annotation and API configuration).

francoisledroff:
Hey Gunnar,

Thanks for the feedback.

I actually tried at first to embrace the JPA annotations.
But as far as I recall the annotation parsing made by `hibernate-core` lost some information on the way.
Especially when I used `unique=true` on a single field/column index.

As an end-user, I also preferred the field annotation, but that is a question of taste I guess.

I will give it another try, and have a look at the `Redis` code pointers you gave me.

core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
francoisledroff: I actually looked at the `option` path too.
But sadly, I cant recall why I dismissed using that ..., I will try again 
and this time. I'll take some note, if I get stuck again.

francoisledroff:
@gunnarmorling 

I gave JPA indexes another try, here is the issue I faced las time

I annotated my test entity as follows:

    @Entity
    @Table(name="POEM", indexes = @Index(columnList = "author ASC", name = "author_idx", unique=true))
     public class Poem {
	private String id;
	private String name;
       private String author;
    }

In the `OmgEntityPersister` I tried to get hold of the above index meta-data:

* when `unique` is `false` I can use `persistentClass.getTable().getIndexIterator()` this works , but when `unique` is `true` the `persistentClass.table.indexes` Map is empty.
* when `unique`, it looks I can get hold of the index info by fetching `persistentClass.getTable().getUniqueKeyIterator()`   

This is due to [hibernate-core processing](https://github.com/hibernate/hibernate-orm/blob/5.0/hibernate-core/src/main/java/org/hibernate/boot/internal/InFlightMetadataCollectorImpl.java#L1997)

That's the extra complexity that lead me to dismiss using the JPA annotation in my first commits.
But if this looks legitimate to you I will proceed and start using the above strategy and my next commits.

gunnarmorling:
Hey @francoisledroff, I don't think `OgmEntityPersister` is the right place to access the index information. Instead, the `SchemaDefiner` implementation should evaluate it (should be retrievable from the table mappings via `database`) and create the right indexes in the database.

francoisledroff:
@gunnarmorling  Thanks for your quick feedback

Sure I can move that to `MongoDBEntityMappingValidator` (which extends `BaseSchemaDefiner`) 
But the complexity remains the same: whether you `true` or `false` the index `unique` attribute you will have to fetch

* either `context.getDatabase().getDefaultNamespace().getTables()` and iterate to get `UniqueKeyIterator`
* or `context.getDatabase().getDefaultNamespace().getTables()` and iterate to get the `IndexIterator`

Does this extra (hibernate-core) complexity looks legitimate to you? if so I will proceed and start using the above strategy and my next commits.

gunnarmorling:
Yes, I think that's the way forward. I.e. use the meta-data as provided by ORM core and do the right thing based on this in the schema definer. I am not sure what you mean by "extra (hibernate-core) complexity", do you mean the complexity of *using* this functionality from core? I think that's alright, better than re-building it entirely on our side.

Some additional complexity may arise from combining that meta-data with meta-data fed in through the OGM option system (let's say to mark an index as MongoDB hash index), but I'd say let's cross that bridge when we get there.

francoisledroff:
@gunnarmorling

This is still in progress

* lots of TODOs (among them proper index spec validation)
* lots of missing unit-tests 
* missing code clean up
* missing javadocs 
* check-style validation and missing headers

and later support (other Jira entries TODO)
* Compound Indexes
* Multikey Indexes
* Geospatial Indexes and Queries
* Text Indexes
* Hashed Index

core/src/main/java/org/hibernate/ogm/dialect/spi/GridDialect.java
```diff
@@ -183,4 +188,17 @@
...
 	 */
 	DuplicateInsertPreventionStrategy getDuplicateInsertPreventionStrategy(EntityKeyMetadata entityKeyMetadata);
+
+	/**
+	 * //TODO doc
+	 * @param tableName
+	 * @param indexAnnotations
+     * @return
+     */
+	List<OgmIndexSpec> getIndexSpec(String tableName, Map<String, Annotation> indexAnnotations);
```
gunnarmorling: @francoisledroff Is this still needed as per your latest changes?

gunnarmorling:
Hey @francoisledroff, it seems to be in a somewhat intermediary state?

Generally speaking, I think you are on the right track by using the schema definer for this instead of going through persisters. One open question (would be interesting to see what the other guys think is), whether

* schema definers themselves should navigate the `Mapping` and do with it whatever they have
* or whether there should be a contract with specific "instructions" such as `createTable()`, `createIndex()` etc. which would be invoked by core to "drive" the schema initialization. Might be useful to prepare the schema data but I am not sure whether it's really worth it and how much we'd gain

So personally I think I'd lean towards the first option, see how many commonalities we have and then reconsder the second option down the road if it appears sensible. WDYT everyone?

core/src/main/java/org/hibernate/ogm/index/OgmIndex.java
```diff
@@ -0,0 +1,25 @@
...
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * MetaAnnotation for various Index related annotation
+ *
+ * @author Francois Le Droff
+ */
+@Target({ ANNOTATION_TYPE })
+@Retention(RUNTIME)
+public @interface OgmIndex {
```
DavideD: What's this for?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBEntityMappingValidator.java
```diff
@@ -33,8 +48,76 @@ public void validateMapping(SchemaDefinitionContext context) {
...
+		indexSpecs= new ArrayList<>();
+		Database database = context.getDatabase();
+		for(Table table : database.getDefaultNamespace().getTables())
+		{
+			Iterator<UniqueKey> keys = table.getUniqueKeyIterator();
+			while(keys.hasNext())
+			{
+				IndexSpec indexSpec = new IndexSpec( keys.next() );
+				validateIndexSpec(indexSpec);
+				indexSpecs.add(indexSpec);
```
DavideD: I wonder if we should add a `preProcessing(...)` method to initialize some variables. I don't like too much that the validation method is also responsible for the initialization of the fields

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/index/IndexSpec.java
```diff
@@ -0,0 +1,152 @@
...
+
+
+	public String getCollection() {
+		return collection;
+	}
+
+	private void addIndexKeys(Map<Column,String> columnOrderMap) {
+		for(Column column : columnOrderMap.keySet())
+		{
+			indexKeys.put(column.getName(),(columnOrderMap.get(column).equals("asc")) ? 1 : -1);
```
DavideD: Couldn't you use the values in IndexOrder instead of these numbers ?

DavideD:
> schema definers themselves should navigate the Mapping and do with it whatever they have

I prefer this option as well. I don't see any benefit in the creation of a more detailed contract at the moment.

DavideD:
I've added a few comments, overall the approach of using a SchemaDefiner seems the way to go.
Waiting for additional changes as the pull request seems in an intermediary state (as @gunnarmorling  pointed out)

core/src/main/java/org/hibernate/ogm/dialect/spi/GridDialect.java
francoisledroff: no It will be deleted

core/src/main/java/org/hibernate/ogm/index/OgmIndex.java
francoisledroff: dead code from my first iteration, see the first comments and the first commit.
I will clean that up.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/index/IndexSpec.java
francoisledroff: sure I will.

francoisledroff:
Thanks for the feedback.
I will proceed and clean that mess up and work on the `TODO`s mentioned above.

gunnarmorling:
@francoisledroff I saw you pushed some stuff. Do you think it's good to go now? Or are more changes coming up? Btw. can you use the email address from your GH account for authoring commits, otherwise they will not be attributed to your account properly.

francoisledroff:
@gunnarmorling  no there are still a few missing pieces.... for it to be usable
plus I now have some conflict to resolve ;-(

gunnarmorling:
Hey @francoisledroff, howdy? Did you ever put any further effort into this? I/we can take over, just thought I'd touch base... :)

vladmihalcea:
Hi @francoisledroff,

We are looking for someone to share their story about the Hibernate OGM in our newsletter. I think this is a good opportunity for your team to advertise your product too. Let me know if you are interested.

Thanks, Vlad

gsmet:
Closing this PR as I'm working on another one. It will integrate Francois's work.

PR #615, opened 2016-01-12
OGM-941 Add mapping tests for Cassandra

nibbio84:
I converted some MongoDB tests into Cassandra tests, in order to verify the database mapping using a builder-style syntax for assertions.
I found out that Yes/No and True/False types were treated incorrectly, as they were stored as "true" - "false" strings, instead of using "Y" - "N", or "T" - "F".
I also fixed the issue.


gunnarmorling:
ok to test

gunnarmorling:
> I found out that Yes/No and True/False types were treated incorrectly, as they were stored as "true" - "false" strings, instead of using "Y" - "N", or "T" - "F".

Very cool, awesome that you found those! Could you just create a separate JIRA issue, split the commit into two, one for adding the mapping tests and one for fixing this bug? Thanks!

cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/test/mapping/BuiltinTypeMappingTest.java
```diff
@@ -0,0 +1,117 @@
...
+		Transaction transaction = session.beginTransaction();
+
+		session.delete( session.get( Bookmark.class, bookmarkId ) );
+
+		transaction.commit();
+		session.close();
+	}
+
+	@Test
+	public void booleanMapping() {
```
gunnarmorling: How about this: Let's add interfaces in core which defines the test methods for the different mapping tests and have specific test classes implement these interfaces. We cannot share the actual tests as the assertions are different, but that way we'd at least know we don't forget any dialect when adding a mapping test (method) in core. WDYT?

gunnarmorling: Or, by making it an abstract class instead of an interface, we even could reuse some set-up / tear-down code.

cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/test/mapping/ElementCollectionListWithIndexTest.java
```diff
@@ -0,0 +1,78 @@
...
+	public void testMapping() throws Exception {
+		OgmSession session = openSession();
+		Transaction transaction = session.beginTransaction();
+
+		rowAssertion( session.getSessionFactory(), "GrandMother" )
+				.id( "id", granny.getId() )
+				.assertNoOtherColumnPresent()
+				.execute();
+
+		rowAssertion( session.getSessionFactory(), "GrandMother_grandChildren" )
```
gunnarmorling: I like that approach, it might even allow to implement it identically for other dialects.

I'd only rename `id()` into `keyColumn`, because that's what it is: one id/key comprising of several columns, not several ids.

cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/test/id/GuitarPlayer.java
```diff
@@ -0,0 +1,49 @@
...
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.TableGenerator;
+
+/**
+ * @author Nicola Ferraro
+ */
+@Entity
+public class GuitarPlayer {
```
gunnarmorling: Could you think about whether we can re-use these types across different dialects?

cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/test/mapping/BuiltinTypeMappingTest.java
nibbio84: Yes, I think it is a good idea, at least for type mapping. I can also extend the test to other types, not only the boolean specific case.

cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/test/mapping/ElementCollectionListWithIndexTest.java
nibbio84: Right!

cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/test/id/GuitarPlayer.java
nibbio84: Yes, I think it is better to move them to the core.

nibbio84:
I will open a separate ticket for the boolean mapping issue and another pull request.

cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/type/impl/CassandraYesNoType.java
```diff
@@ -21,7 +20,7 @@
 	public static final CassandraYesNoType INSTANCE = new CassandraYesNoType();
 
 	public CassandraYesNoType() {
-		super( StringMappedGridTypeDescriptor.INSTANCE, org.hibernate.type.YesNoType.INSTANCE.getJavaTypeDescriptor() );
+		super( CharacterStringGridTypeDescriptor.INSTANCE, org.hibernate.type.YesNoType.INSTANCE.getJavaTypeDescriptor() );
```
gunnarmorling: Could you check whether this one works alternatively:

    super( new TranslatingGridTypeDescriptor( Character.class ), org.hibernate.type.YesNoType.INSTANCE.getJavaTypeDescriptor() );

Would save that new grid descriptor type. The whole type business is far from perfect btw., it's very high on our agenda to rework that. So many thanks for fighting through it :)

cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/test/id/GuitarPlayer.java
gunnarmorling: Cool; Some are possibly in Mongo/Redis/CouchDB, too, I hope they don't differ :)

cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/test/mapping/BuiltinTypeMappingTest.java
gunnarmorling: > Yes, I think it is a good idea, at least for type mapping.

Cool.

> I can also extend the test to other types, not only the boolean specific case.

That'd be awesome, only I'd say let's do it via a separate issue + pull request, too. Generally, we try to limit one issue and PR to one specific feature or improvement, to keep reviews, history and changelog manageable.

gunnarmorling:
@nibbio84 Good stuff :)

Could you sign the CLA for Hibernate (ORM), too? It's needed since the Bean Validation one is based on another license (ASF2 vs. LGPL). Good news is that should you contribute to HSEARCH at some point, you don't need to sign yet another one as that's LGPL, too :)

nibbio84:
I separated the two issue.

cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/test/embeddable/EmbeddableMappingTest.java
```diff
@@ -0,0 +1,202 @@
...
+		transaction = session.beginTransaction();
+		loadedAccount = session.get( Account.class, account.getLogin() );
+		session.delete( loadedAccount );
+		transaction.commit();
+
+		session.close();
+	}
+
+	@Test
+	@Ignore("@ElementCollection seems not to work in Cassandra if no @OrderColumn is specified")
```
gunnarmorling: Indeed that's not supported, as we cannot determine a primary key for the table in such case. I'll remove the test method altogether. Once we have interfaces for the mapping tests in core (see [OGM-965](https://hibernate.atlassian.net/browse/OGM-965), we'd have to skip the test on Cassandra via `@SkipByGridDialect`.

cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/test/mapping/BuiltinTypeMappingTest.java
```diff
@@ -0,0 +1,120 @@
...
+				.keyColumn( "id", bookmarkId )
+				.assertColumn( "favourite", true )
+				.execute();
+
+		transaction.commit();
+		session.close();
+	}
+
+	@Test
+	@Ignore("OGM-962 True/False mapping needs to be fixed")
```
gunnarmorling: I'll enable these now, as your fix is already merged.

gunnarmorling:
Rebased and applied. Good stuff, @nibbio84!

PR #619, opened 2016-01-17
OGM-962 Fix Yes/No and True/False type mapping in Cassandra

nibbio84:
Boolean types mapped to Yes/No or True/False should be mapped to 'Y'/'N' or 'T'/'F' characters. Since Cassandra does not currently support the character data type, they should be single character strings. 
There is evidence in the code that this was the expected mapping, anyway, that fields are incorrectly mapped to "true" or "false" (strings).

I've implemented the change to fix the behavior and written unit tests. I added mockito to the pom test dependencies.

As suggested by @gunnarmorling, I've tried using the TranslatingGridTypeDescriptor to fix the issue, without success.

### DIFF
--- a/cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/test/mapping/CassandraBooleanCharacterTypeTest.java
+++ b/cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/test/mapping/CassandraBooleanCharacterTypeTest.java
@@ -1,0 +1,74 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.datastore.cassandra.test.mapping;
+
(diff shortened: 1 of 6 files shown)


cassandra/src/test/java/org/hibernate/ogm/datastore/cassandra/test/mapping/CassandraBooleanCharacterTypeTest.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Tests that 'yes_no' and 'true_false' types are correctly mapped to single-character strings.
+ *
+ * @author Nicola Ferraro
+ */
+public class CassandraBooleanCharacterTypeTest {
+
+	@Test
+	public void testYesNoTypeMapping() {
+
```
gunnarmorling: Interesting, I never thought of testing it this way :)

Usually we use more integrative means of testing (like the mapping format tests you've seen), but I suppose you did it that way because the required infra for that is part of your other PR? That's cool, if you like you can change it later on, but nothing critical to me.

gunnarmorling:
Merged; Thanks, @nibbio84! Note that I've reworked it a bit to actually make use of `TranslatingGridTypeDescriptor`. That avoids the new descriptor.

PR #620, opened 2016-01-20
OGM-966 Neo4j test sequences

DavideD:
https://hibernate.atlassian.net/browse/OGM-966

In prepartion of the new Neo4j Remote dialect, this pull request cleans up the test in Neo4j that checks the thread saftiness of the sequence generation.

I guess we could move this tests in core and execute them on all the datatastores (as a separate issue).


gunnarmorling:
Rebased and applied. Thanks, @DavideD.

Sanne:
sweet, it was bothering me that we didn't have such a test.

PR #621, opened 2016-01-21
OGM-969 DefaultAssociationKeyMetada#equals

DavideD:
https://hibernate.atlassian.net/browse/OGM-969

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/model/impl/DefaultAssociationKeyMetadata.java
+++ b/core/src/main/java/org/hibernate/ogm/model/impl/DefaultAssociationKeyMetadata.java
@@ -289,6 +289,14 @@ public class DefaultAssociationKeyMetadata implements AssociationKeyMetadata {
 			return false;
 		}
 
+		if ( collectionRole == null && that.collectionRole != null ) {
+			return false;
+		}
+
+		if ( !collectionRole.equals( that.collectionRole ) ) {
+			return false;
+		}
+
 		return true;
 	}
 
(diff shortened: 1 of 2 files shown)


gunnarmorling:
@DavideD, did you spot an actual case where this is a problem? Or is it more a theoretical/aesthetic issue?

DavideD:
> @DavideD, did you spot an actual case where this is a problem? Or is it more a theoretical/aesthetic issue?

You cannot use it as key in a Map (when you want to pre-compute some association specific operations, like Cypher queries).

gunnarmorling:
Understood about the inability of using them in maps or sets, just wasn't sure where we'd actually be doing this :)

gunnarmorling:
I think i'd rather add `columnName` to the hash code determination. Seems more in line with how we handle these things in general.

DavideD:
> I think i'd rather add columnName to the hash code determination. Seems more in line with how we handle these things in general.

The two associations have the same column and table names, I'm not sure how adding the columns to the hashcode can solve the issue.

gunnarmorling:
@DavideD The thing is that you need two distinct join tables for this mapping, one per association (In ORM you'd run into an exception with your original mapping).

And for the case we we use one table to map both navigations of a bi-di association (the bank account / account owner example), adding the column names should do the trick to make it distinguishable.

gunnarmorling:
@DavideD, did you ever follow up on this one?

DavideD:
> adding the column names should do the trick to make it distinguishable.

If you are talking about AssociationKeyMetadata#getColumnNames, that does not work.
Maybe I don't understand what you are suggesting, if you show me how you would change the equals I will test it on the Neo4j remote branch.

Sanne:
so this is a requirement for #623 ? Is something preventing this to be merged first then?

DavideD:
> so this is a requirement for #623 ? Is something preventing this to be merged first then?

The thing is that this mapping is not supported by ORM so I will close the pull request and change the tests.

PR #622, opened 2016-01-22
OGM-945, OGM-970

gunnarmorling:
OGM-945: Using multi-id-get also if no batch size has been given explicitly; This avoids n+1 SELECTs and I think is a useful default behavior as we can't do fetch joins or get associations by a real query.
OGM-970: Allows to use mass indexer and queries also on entities which have a batch size set (which now will be the default for multi-id-get enabled dialects as per OGM-945). This was an existing bug, it's just now much more apparent due to OGM-945.


core/src/main/java/org/hibernate/ogm/loader/impl/TupleLoader.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * "Loads" entities from given tuple representations.
+ * <p>
+ * TODO: This is used for the mass indexer and also to convert query results (tuple collections) into objects. At least
+ * the latter shouldn't work that way, instead the loader itself should obtain the tuple result set from the datastore,
+ * so we have query execution time statistics correct etc.
+ *
+ * @author Gunnar Morling
+ */
+public interface TupleLoader extends UniqueEntityLoader {
```
DavideD: Shouldn't this be called something like `EntitiesLoader`?

DavideD:
I need to run it but it looks good.
I'm just not sure about the name of `TupleLoader`.

core/src/main/java/org/hibernate/ogm/loader/impl/TupleLoader.java
gunnarmorling: But they all load entities in the end, no? I meant to express the fact that it obtains objects from tuples.

DavideD: `TupleLoader` sounds like something that load tuples not entities to me.

> I meant to express the fact that it obtains objects from tuples.

Shouldn't be the method to highlight it to something like `loadFromTuples`

gunnarmorling:
Renamed it to `TupleBasedEntityLoader`/ `loadEntitiesFromTuples()` and force-updated this PR.

DavideD:
Rebased and Merged https://github.com/hibernate/hibernate-ogm/pull/622

PR #623, opened 2016-01-22
OGM-366 Remote Neo4j 

DavideD:

I've tried to split in sensible commits but there are still a lot of classes and changes.
In commits https://github.com/hibernate/hibernate-ogm/commit/95183dc3669c61ab45e9cc7aa4df19854f0e9f8d and https://github.com/hibernate/hibernate-ogm/commit/b599ba4ee82f202dcf360a64a7c00c7c8c1457b6 I've added a new `TransactionContext` class that allows the dialect to access the TransactionDriver and then get the tx id from it.

An alternative solution could be to add some method in DataStoreProvider that returns a Transaction identifier given the TransactionCoordinator.

What it is missing:
* Integration tests with the remote server
* 1 to 1 embedded are not loaded eagerly
* Documentation

Follow up of: https://github.com/hibernate/hibernate-ogm/pull/596

UPDATE: Built on top of https://github.com/hibernate/hibernate-ogm/pull/621


core/src/main/java/org/hibernate/ogm/dialect/spi/GridDialect.java
```diff
@@ -47,16 +47,17 @@
 	 *
 	 * @param key The tuple identifier
 	 * @param tupleContext Contains additional information that might be used to create the tuple
+	 * @param transactionContext Contains information related to the running transaction, it might be null when the db does not require transacitons
 	 * @return the tuple identified by the key
 	 */
-	Tuple getTuple(EntityKey key, TupleContext tupleContext);
+	Tuple getTuple(EntityKey key, TupleContext tupleContext, TransactionContext transactionContext);
```
gunnarmorling: Can you make it part of `TupleContext`? That way `TupleContext` (and similarly, `AssociationContext`) are the entry point into all the information:

* "static" one, as exposed through `AssociationContext#getAssociationTypeContext()`, probably should be done in the same way for the tuple context)
* "runtime" one: operations queue and transaction context

DavideD: ok

gunnarmorling:
Can we validate that a user name + PW is given? Otherwise Neo4j complains a bit cryptically, better to catch it on our side.

core/src/main/java/org/hibernate/ogm/type/impl/ByteArrayMappedType.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.hibernate.MappingException;
+import org.hibernate.engine.spi.Mapping;
+import org.hibernate.ogm.type.descriptor.impl.ByteArrayMappedGridTypeDescriptor;
+import org.hibernate.type.descriptor.java.ByteArrayTypeDescriptor;
+
+/**
+ * @author Davide D'Alto
+ */
+public class ByteArrayMappedType extends AbstractGenericBasicType<Byte[]> {
```
gunnarmorling: Where is the difference between this one and `BinaryMappedType`?

core/src/main/java/org/hibernate/ogm/util/impl/ArrayHelper.java
```diff
@@ -86,4 +87,18 @@ public static boolean contains(Object[] array, Object element) {
 
 		return result;
 	}
+
+	public static <T> T[] concat(List<T[]> arrays) {
+		T[] result = arrays.get( 0 );
```
gunnarmorling: Hum, I think I'd allocate the result once in the final size and then populate it array by array.

neo4j/pom.xml
```diff
@@ -130,4 +156,25 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <profiles>
```
gunnarmorling: Can you check what @mp911de does in the pending Redis PR? There two Surefire executions are configured. Would be nice to do the same here.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/EntityQueries.java
```diff
@@ -0,0 +1,667 @@
...
+						appendRelationshipType( queryBuilder, path[i] );
+						queryBuilder.append( "]->" );
+						if ( i == path.length - 2 ) {
+							queryBuilder.append( "(e:EMBEDDED) " );
+						}
+						else {
+							queryBuilder.append( "(:EMBEDDED) " );
+						}
+					}
+					queryBuilder.append( "REMOVE e." );
```
gunnarmorling: A while back we discussed some DSL for query creation? Is that still not available in Maven Central?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/Neo4jConfiguration.java
```diff
@@ -0,0 +1,101 @@
...
+import org.hibernate.ogm.cfg.spi.DocumentStoreConfiguration;
+import org.hibernate.ogm.cfg.spi.Hosts;
+import org.hibernate.ogm.util.configurationreader.impl.Validators;
+import org.hibernate.ogm.util.configurationreader.spi.ConfigurationPropertyReader;
+
+/**
+ * @see DocumentStoreConfiguration
+ * @author Davide D'Alto
+ */
+public class Neo4jConfiguration {
```
gunnarmorling: TX timeout would be a valuable addition.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/transaction/impl/RemoteLocalTransactionCoordinator.java
```diff
@@ -0,0 +1,105 @@
...
+import org.hibernate.ogm.datastore.neo4j.remote.impl.RemoteNeo4jDatastoreProvider;
+import org.hibernate.ogm.transaction.impl.ForwardingTransactionCoordinator;
+import org.hibernate.ogm.transaction.impl.ForwardingTransactionDriver;
+import org.hibernate.resource.transaction.TransactionCoordinator;
+
+/**
+ * A {@link TransactionCoordinator} for Neo4j.
+ *
+ * Note that during a JTA transaction Neo4j {@link Transaction} are
+ * synchronized using the {@link Synchronization} interface. A commit to the Neo4j transaction will happen before the
```
gunnarmorling: How / by whom is that commit triggered?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/transaction/impl/RemoteLocalTransactionCoordinator.java
```diff
@@ -0,0 +1,105 @@
...
+		if ( tx != null ) {
+			tx.rollback();
+			tx = null;
+		}
+	}
+
+	private class RemoteTransactionDriver extends ForwardingTransactionDriver implements Neo4jTransactionDriver {
+
+		public RemoteTransactionDriver(TransactionDriver delegate) {
+			super( delegate );
```
gunnarmorling: Could you make sure that the delegate works the `NoopJdbcResourceTransaction` (requires to make the latter a non-private class). The default one works, but the less JDBC-related pieces are in there, the better. IIRC, you only delegate to benefit from the flush triggering upon commit, right? Would be good to document the reasoning behind the design.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/transaction/impl/RemoteLocalTransactionCoordinator.java
```diff
@@ -0,0 +1,105 @@
...
+import org.hibernate.ogm.datastore.neo4j.remote.impl.Neo4jClient;
+import org.hibernate.ogm.datastore.neo4j.remote.impl.RemoteNeo4jDatastoreProvider;
+import org.hibernate.ogm.transaction.impl.ForwardingTransactionCoordinator;
+import org.hibernate.ogm.transaction.impl.ForwardingTransactionDriver;
+import org.hibernate.resource.transaction.TransactionCoordinator;
+
+/**
+ * A {@link TransactionCoordinator} for Neo4j.
+ *
+ * Note that during a JTA transaction Neo4j {@link Transaction} are
```
gunnarmorling: How does JTA play into this, given the class is named "...Local..."?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/transaction/impl/RemoteTransactionCoordinatorBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+import org.hibernate.resource.transaction.TransactionCoordinator;
+import org.hibernate.resource.transaction.TransactionCoordinatorBuilder;
+import org.hibernate.resource.transaction.spi.TransactionCoordinatorOwner;
+
+/**
+ * Builder for a Neo4j remote {@link TransactionCoordinator}.
+ *
+ * @author Davide D'Alto
+ */
+public class RemoteTransactionCoordinatorBuilder implements TransactionCoordinatorBuilder {
```
gunnarmorling: Could be derived from `ForwardingTransactionCoordinatorBuilder`.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/transaction/impl/RemoteJtaTransactionCoordinator.java
```diff
@@ -0,0 +1,146 @@
...
+import org.hibernate.ogm.datastore.neo4j.remote.impl.RemoteNeo4jDatastoreProvider;
+import org.hibernate.ogm.transaction.impl.ForwardingTransactionCoordinator;
+import org.hibernate.ogm.transaction.impl.ForwardingTransactionDriver;
+import org.hibernate.resource.transaction.TransactionCoordinator;
+import org.hibernate.resource.transaction.spi.TransactionStatus;
+
+/**
+ * @author Davide D'Alto
+ */
+public class RemoteJtaTransactionCoordinator extends ForwardingTransactionCoordinator {
```
gunnarmorling: Is a custom coordinator even needed? Couldn't you just add a `Synchronization` to the default coordinator in the coordinator builder and use it then as is, i.e. without all the delegation business?

core/src/main/java/org/hibernate/ogm/hibernatecore/impl/BackendCustomLoader.java
```diff
@@ -97,6 +101,14 @@ private boolean isEntityQuery() {
 		}
 	}
 
+	private TransactionContext transactionContext(SessionImplementor session) {
```
gunnarmorling: This seems to exist in several places. How about making it a static method on `TransactionContext` itself.

core/src/main/java/org/hibernate/ogm/dialect/spi/TransactionContext.java
```diff
@@ -0,0 +1,17 @@
...
+package org.hibernate.ogm.dialect.spi;
+
+import org.hibernate.resource.transaction.TransactionCoordinator.TransactionDriver;
+
+/**
+ * @author Davide D'Alto
+ */
+public interface TransactionContext {
+
+	TransactionDriver getTransactionDriver();
```
gunnarmorling: @emmanuelbernard How do you feel about exposing `TransactionDriver` here? It's the right "scope" and it works, only I am wondering whether we should go for something abstract.

I.e. should we define a callback which allows to populate the context with the "business" data that the grid dialect operations are actually interested in (such as the TX id in Neo's case). Then `TransactionContext` would have a generic API a la `get(TransactionId.class)` or `get(Long.class)`, isolating grid dialects from the TX driver contract.

It's not strictly needed, but it seems a bit more extensible than fetching the state from the plain TX driver. Also it'd allow us to feed in generic TX state from within core down the road and dialects could access it (no idea whether we'll need it, but it'd be nice to have that option).

WDYT?

gunnarmorling:
@DavideD, done with the first round. I focused on TX handling mainly. Did not look into the mapping yet, but I suppose it's identical to what you've done for embedded? Or is there any difference.

core/src/main/java/org/hibernate/ogm/dialect/spi/TransactionContext.java
DavideD: Sounds good, one thing that I don't like about the TransactionDriver is that it exposes the begin-commit-rollback methods.

DavideD:
Thanks @gunnarmorling,

No, there is shouldn't be any difference in the mapping.

gurkerl83:
Wow, this is a monster pull request! I have no idea if the compatibility with our approach works ones this gets merged. So far we are using the latest hibernate-ogm-neo4j version and everything works fine. However our application has to evolve from the embedded datastore to a remote datastore in a mid term. One question is about the upcoming Neo4j 3.0 release and the bolt protocol they have introduced in version 3. Is this pull request capable of adapting the connection approach to https://github.com/neo4j/neo4j-java-driver ones it is stable, instead of using the rest connection to communicate cypher queries?

@DavideD it`s always a pleasure how you constantly evolve hibernate-ogm-neo4j, please keep up this great work!

Thanks,

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/EntityQueries.java
DavideD: The problem with it was that it wasn't aligned with the feature in Cypher.
but they might have improve it. I didn't check

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/transaction/impl/RemoteLocalTransactionCoordinator.java
DavideD: I guess it was a copy and paste error

DavideD:
> Wow, this is a monster pull request!
Yes, sorry. It usually happens when a new backend integration get built.

> One question is about the upcoming Neo4j 3.0 release and the bolt protocol they have introduced in version 3. Is this pull request capable of adapting the connection approach to https://github.com/neo4j/neo4j-java-driver ones it is stable, instead of using the rest connection to communicate cypher queries?

I still need to read more about Neo4j 3, but OGM as several layers of abstraction and since we already generate the cypher queries that we need, my guess is that it shouldn't be too hard to use a different client to run them.

> @DavideD it`s always a pleasure how you constantly evolve hibernate-ogm-neo4j, please keep up this great work!

Thank you very much, it's great to hear from someone using it :)
Feel free to leave some feedback, it helps us a lot.

Sanne:
This needs to be rebased: there are several conflicts.

DavideD:
> This needs to be rebased: there are several conflicts.

I'm working on it

core/src/main/java/org/hibernate/ogm/dialect/spi/GridDialect.java
DavideD: @gunnarmorling  QueryableGridDialect#execute* does not have a context. Should I add the TransactionCOntext as parameter or should I create a new QueryContext for future extensions?

gunnarmorling: The latter seems a bit more flexible.

DavideD:
I'm going to close this and  send separate smaller pull requests

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/transaction/impl/RemoteJtaTransactionCoordinator.java
DavideD: As far as I know `Synchronization` does not tell me when the transaction starts, it only tells me when the before and after commit event. The reason I have all this delegation business is to provide a `TransactionDriver` bound to the remote neo4j transaction.
Let me know if I'm missing something.

Sanne: that's right, a Transaction might have been started long before OGM got involved.
But why does that matter? Our "transaction" with Neo4J starts when we first read/write to it.

DavideD: > Our "transaction" with Neo4J starts when we first read/write to it.

Right, I have to admit I haven't thought about it. I thought it should be started when the begin method is called. I'll try to change it in that direction.

PR #628, opened 2016-02-19
OGM Dialect for Apache Ignite and GridGain

Z-z-z-z:
Hi!
Here is dialect for Apache Ignite (http://ignite.apache.org/features/datagrid.html) and it's commertial version GridGain (http://www.gridgain.com/products/in-memory-data-fabric/overview/)

Victor


Sanne:
@Hibernate-CI ok to test

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/dialect/criteria/spi/CriteriaGridDialectInitiator.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.Map;
+
+import org.hibernate.boot.registry.StandardServiceInitiator;
+import org.hibernate.ogm.dialect.impl.GridDialects;
+import org.hibernate.ogm.dialect.spi.GridDialect;
+import org.hibernate.service.spi.ServiceRegistryImplementor;
+
+/**
+ * vk: 09.02.16. убрал его использование. удалить
```
Sanne: I don't understand this! sorry :)

Z-z-z-z: O! I'm sorry!
I'll change comments.

Victor

2016-02-19 17:15 GMT+03:00 Sanne Grinovero <notifications@github.com>:

gunnarmorling:
@DavideD, are you going to look into this one?

gunnarmorling:
Thanks for this contribution, @Z-z-z-z. Could you give high-level overview about this implementation, i.e. how are typical JPA idioms mapped to GridGain? My knowledge of this store is very limited and some help to getting into it would be much appreciated to get a feeling for this pull request. Thanks!

DavideD:
@gunnarmorling yes, I'll take this

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/impl/IgniteSessionImpl.java
```diff
@@ -0,0 +1,88 @@
...
+import org.hibernate.internal.CriteriaImpl;
+import org.hibernate.ogm.OgmSessionFactory;
+import org.hibernate.ogm.datastore.ignite.loader.criteria.OgmCriteriaLoader;
+import org.hibernate.ogm.datastore.ignite.loader.criteria.impl.CriteriaCustomQuery;
+import org.hibernate.ogm.exception.NotSupportedException;
+import org.hibernate.ogm.hibernatecore.impl.OgmSessionImpl;
+import org.hibernate.resource.transaction.TransactionCoordinator;
+import org.hibernate.resource.transaction.backend.jta.internal.JtaTransactionCoordinatorImpl;
+
+public class IgniteSessionImpl extends OgmSessionImpl {
```
DavideD: What's the reason for creating a specific session for Ignite?

ignite/src/test/resources/org/hibernate/ogm/datastore/ignite/config.xml
```diff
@@ -0,0 +1,499 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+    Copyright (C) GridGain Systems. All Rights Reserved.
```
DavideD: I don't know if we can include this copyright, @emmanuelbernard ?

ignite/src/test/java/org/hibernate/ogm/datastore/ignite/utils/IgniteTestHelper.java
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	@Override
+	public void dropSchemaAndDatabase(SessionFactory sessionFactory) {
+		// TODO че тут делать???
+	}
+
+	@Override
+	public Map<String, String> getEnvironmentProperties() {
+		// TODO а тут?
```
DavideD: This is used by some dialect to set some properties programmatically, see for example Neo4jTestHelper
It's ok to not implemented if you don't need it

ignite/src/test/java/org/hibernate/ogm/datastore/ignite/utils/IgniteTestHelper.java
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	@Override
+	public boolean backendSupportsTransactions() {
+		return true;
+	}
+
+	@Override
+	public void dropSchemaAndDatabase(SessionFactory sessionFactory) {
+		// TODO че тут делать???
```
DavideD: This will be called by the test methods extending OgmTestCase, and it's executed before closing the session factory.
An in memory db I don't think has to do anything (when dealing with remote dbs we want to drop the db, for example)

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/impl/IgniteSessionImpl.java
Z-z-z-z: Hi!

We have to do this because we implements Criteria API, but
in OgmSessionImpl we have only

 @Override public Criteria createCriteria(Class persistentClass, String
alias) { //TODO plug the Lucene engine throw new NotSupportedException(
"OGM-23", "Criteria queries are not supported yet" ); }
And I've always wanted to ask why don't you support criteria queries?

Victor


2016-03-08 2:34 GMT+03:00 DavideD <notifications@github.com>:

DavideD: > And I've always wanted to ask why don't you support criteria queries?

Lack of time and resources I think, we wanted to work on that but we decided to work on other features/dialects first.

This is also the reason it might take some time to review big PR like this one

emmanuelbernard: Actually the reason is that a future version of Hibernate ORM will merge the generation tree of JP-QL and Criteria queries and generate a common AST. OGM could then get criteria support for free. It is a longer time coming but than expected but the idea of implementing criteria to ditch the implementation later is not super appealing compared to the usefulness of the feature compared to the other feature ideas we have.

ignite/src/test/java/org/hibernate/ogm/datastore/ignite/test/IgniteTest.java
```diff
@@ -0,0 +1,130 @@
...
+			e.printStackTrace();
+		}
+
+		session.close();
+
+	}
+
+	@Test
+	@Ignore
+	public void testUpdateClient() throws Exception {
```
DavideD: How come all these tests are ignored?

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/impl/IgniteSessionImpl.java
DavideD: I think it would be best to separate the Criteria part from the implementation of the dialect.
It would be a new feature for Hibernate OGM and it requires some additional discussion.

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/PortableBuilderDelegate.java
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.datastore.ignite;
+
+import org.apache.ignite.binary.BinaryObjectException;
+import org.jetbrains.annotations.Nullable;
+
+public interface PortableBuilderDelegate {
```
DavideD: THis is not use now, Is it a replacement for the BinaryObjectBuilder?

DavideD:
As far as I understand you are using the same approach used in Infinispan, looks good overall.
I think it would be nice if you could send a new pull regquest with only ignite, without the Critera implementation, the commented  unusued code and the ignored tests.

We can also add later the integration with GridGain, but we wiill need of a way to test it on our CI.

You could send a second PR with the Criteria APi , it would also be nice to discuss about it in the mailing list, as I don't think that the creation of a new Session implementation is the right approach.

Last but not least, sorry it took me so long to review this PR next time I will be faster.

Thanks a lot for the contribution

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/PortableBuilderDelegate.java
Z-z-z-z: We don't need it now. I'll remove it.

2016-03-17 3:01 GMT+03:00 DavideD <notifications@github.com>:

Z-z-z-z: Hello!
I think I've done all you asked.
I removed criteria implementation, unused classes and GridGain support.

Victor


2016-03-17 12:43 GMT+03:00 Victor Z <vikad111@gmail.com>:

Z-z-z-z:
I think I've done all you asked.
I removed criteria implementation, unused classes and commercial version support.
Is it ok?

DavideD:
Thanks @Z-z-z-z , I'll have a look today

Sanne:
Hi @Z-z-z-z 

> I removed criteria implementation, unused classes and commercial version support.

Just to be clear: we're not against supporting commercial versions. That's totally cool, we can add that back in later.
The only requirement is that we can find a reasonable way to test it automatically, especially to prevent future regressions introduced by other people.

DavideD:
@Z-z-z-z It seems that the maven build is failing, could you add "ignite" to the [list of maven modules](https://github.com/hibernate/hibernate-ogm/blob/master/pom.xml#L24) to build. This way the build with jenkins will fail if something is wrong.

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/IgniteDialect.java
```diff
@@ -0,0 +1,533 @@
...
+				Map<BinaryObject, BinaryObject> associationMap = binaryObject.field( "ASSOCIATION" );
+				if (associationMap != null) {
+					for (BinaryObject portableKey : associationMap.keySet()) {
+						BinaryObject portableValue = associationMap.get( portableKey );
+						RowKey rowKey_tmp = portableKey.<RowKey>deserialize();
+						// sort arrays in RowKey for correct comparing of keys
+						String[] columnNames = rowKey_tmp.getColumnNames();
+						Object[] columnValues = rowKey_tmp.getColumnValues();
+						Arrays.sort( rowKey_tmp.getColumnNames() );
+						Arrays.sort( rowKey_tmp.getColumnValues() );
```
DavideD: This does not seem right, could you explain it a bit? 
I can see two problems here:
1) You changing the position of the values, this means that you don't know which value correspond to a column;
2) getColumnValues can contains different types, you cannot order them without having a classcast exception

ignite/pom.xml
```diff
@@ -0,0 +1,264 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>org.hibernate.ogm</groupId>
+            <artifactId>hibernate-ogm-core</artifactId>
+			<exclusions>
+				<exclusion>
+					<artifactId>xml-apis</artifactId>
+					<groupId>xml-apis</groupId>
+				</exclusion>
+			</exclusions>
```
DavideD: How come you need to exclude this?

ignite/pom.xml
```diff
@@ -0,0 +1,264 @@
...
+		</dependency>
+-->		
+        <!-- Ignite 1.5.0 -->
+<!--  		
+-->		
+        <dependency>
+			<groupId>org.apache.ignite</groupId>
+			<artifactId>ignite-core</artifactId>
+			<version>1.5.0.final</version>
```
DavideD: It would be nicer if the version could be a property in the BOM. like we did for the other datastore dependencies

ignite/pom.xml
```diff
@@ -0,0 +1,264 @@
...
+		</dependency>
+        <dependency>
+            <groupId>org.apache.ignite</groupId>
+            <artifactId>ignite-spring</artifactId>
+            <version>1.5.0.final</version>
```
DavideD: Is this dependency needed?
Spring is not a requirement to work with OGM

ignite/pom.xml
```diff
@@ -0,0 +1,264 @@
...
+			<version>1.1.3</version>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.ignite</groupId>
+			<artifactId>ignite-jta</artifactId>
+			<version>1.5.0.final</version>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.ignite</groupId>
+			<artifactId>ignite-indexing</artifactId>
```
DavideD: Is this dependency needed?

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/IgniteDialect.java
```diff
@@ -0,0 +1,533 @@
...
+			Map<String, Object> map = new HashMap<>();
+			for (int i = 0; i < value.size(); i++) {
+				ScalarReturn ret = (ScalarReturn) queryReturns.get( i );
+				map.put( ret.getColumnAlias(), value.get( i ) );
+			}
+			return new MapTupleSnapshot(map);
+		}
+	}
+	
+	private class IgnitePortableResultCursor extends BaseResultCursor<Entry<String, BinaryObject>> {
```
DavideD: This class is never used

ignite/pom.xml
```diff
@@ -0,0 +1,264 @@
...
+<!--
+						<exclude>org/hibernate/ogm/backendtck/queries/QueriesWithEmbeddedCollectionTest.java</exclude>
+						<exclude>org/hibernate/ogm/backendtck/queries/CompositeIdQueriesTest.java</exclude>
+						<exclude>org/hibernate/ogm/backendtck/queries/QueriesWithEmbeddedTest.java</exclude>
+						<exclude>org/hibernate/ogm/backendtck/queries/QueriesWithEmbeddedTest.java</exclude>
+						<exclude>org/hibernate/ogm/backendtck/queries/SimpleQueriesTest.java</exclude>
+						<exclude>org/hibernate/ogm/backendtck/queries/JpaQueriesTest.java</exclude> 
+-->
+					</excludes>
+					<argLine> -Dtest=IgniteBackendTckHelper</argLine>
```
DavideD: Do you need to pass the test variable and use the exclude, couldn't you just avoid using the excludes clause?

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/impl/IgniteKeyProvider.java
```diff
@@ -0,0 +1,149 @@
...
+
+	public String getEntityCache(String entity) {
+		if (entity.indexOf( "." ) >= 0) {
+			String[] arr = entity.split( "\\." );
+			if (arr.length != 2) {
+				throw new IgniteHibernateException("Invalid entity name " + entity);
+			}
+			return arr[0];
+		}
+		return entity;
```
DavideD: How come do you split the entity name for the cache? Why limit the length to 2? Could you add some comments?

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/loader/impl/IgniteLoader.java
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public Object load(Serializable id, Object optionalObject, SessionImplementor session, LockOptions lockOptions) {
+		Object result = super.load( id, optionalObject, session, lockOptions );
+		GridDialect gridDialect = entityPersisters[0].getFactory().getServiceRegistry().getService( GridDialect.class );
+		if (lockOptions.getLockMode() == LockMode.PESSIMISTIC_READ && session.isTransactionInProgress() && entityPersisters.length > 0) {
```
DavideD: This part doe not seem Ignite specific, couldn't it be in core? Is it also needed?

ignite/src/test/java/org/hibernate/ogm/datastore/ignite/utils/IgniteTestHelper.java
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	@Override
+	public boolean backendSupportsTransactions() {
+		return true;
+	}
+
+	@Override
+	public void dropSchemaAndDatabase(SessionFactory sessionFactory) {
+		// TODO what to do here???
```
DavideD: This method is supposed to delete the dabase after tests are executed. It's OK to leave it empty if you don't need it

ignite/src/test/java/org/hibernate/ogm/datastore/ignite/utils/IgniteTestHelper.java
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	@Override
+	public Map<String, String> getEnvironmentProperties() {
+		// TODO and here?
```
DavideD: Unless you need to set some properties before running the tests, you can leave this method empty.
Maybe return an empty map.

ignite/src/test/resources/hibernate.properties
```diff
@@ -0,0 +1,11 @@
...
+# Hibernate OGM, Domain model persistence for NoSQL datastores
+#
+# License: GNU Lesser General Public License (LGPL), version 2.1 or later
+# See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+#
+
+hibernate.search.lucene_version=LUCENE_CURRENT
+hibernate.ogm.datastore.provider=org.hibernate.ogm.datastore.ignite.impl.IgniteDatastoreProvider
+hibernate.show_sql=true
+#hibernate.format_sql=true
```
DavideD: We usually keep an empty line at the end of a file.

ignite/src/test/resources/hibernate.properties
```diff
@@ -0,0 +1,11 @@
+#
+# Hibernate OGM, Domain model persistence for NoSQL datastores
+#
+# License: GNU Lesser General Public License (LGPL), version 2.1 or later
+# See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+#
+
+hibernate.search.lucene_version=LUCENE_CURRENT
```
DavideD: I think you can  remove it

ignite/src/test/resources/hibernate.properties
```diff
@@ -0,0 +1,11 @@
...
+hibernate.ogm.datastore.provider=org.hibernate.ogm.datastore.ignite.impl.IgniteDatastoreProvider
```
DavideD: I would set this to IGNITE_EXPERIMENTAL, this way if you change the path or the name of the class you don't need to touch this file. It also make sure that the constant is correct

ignite/src/test/resources/org/hibernate/ogm/datastore/ignite/config.xml
```diff
@@ -0,0 +1,527 @@
...
+        
+        <property name="publicThreadPoolSize" value="2"/>
+        
+		<property name="cacheConfiguration">
+			<list>
+<!-- SequenceIdGeneratorTest  -->
+				<bean id="Song" parent="cache-template">
+					<property name="name" value="Song" />
+				</bean>
+				<bean id="Actor" parent="cache-template">
```
DavideD: What are you defining here? Do we need all this configuration or caches are created programmatically?

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/impl/IgniteDatastoreProvider.java
```diff
@@ -0,0 +1,195 @@
...
+		if (cacheManager != null && !localNode) {
+			Ignition.stop( cacheManager.name(), true );
+		}
+	}
+
+	private String createGridName() {
+		String result = null;
+		if (config.getUrl() != null) {
+			result = config.getUrl().getPath();
+			result = result.replaceAll( "[\\,\\\",:,\\*,\\/,\\\\]", "_" );
```
DavideD: Wouldn't you risk to have collisions between grid names? A comment or javadoc would be nice

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/impl/IgniteDatastoreProvider.java
```diff
@@ -0,0 +1,195 @@
...
+				gridName = createGridName();
+				try {
+					if (gridName != null) {
+						cacheManager = (IgniteEx) Ignition.ignite( gridName );
+					}
+					else {
+						cacheManager = (IgniteEx) Ignition.ignite();
+					}
+				}
+				catch (IgniteIllegalStateException iise) {
```
DavideD: When is this exception thrown?

DavideD:
@Z-z-z-z I think I left enough comments for now, an overview about how you are mapping the results would be nice. As far as I unerstand you are creating a cache per table, am I right?

ignite/pom.xml
Z-z-z-z: I don't know. You are right.

ignite/pom.xml
Z-z-z-z: Spring is necessary to load test configuration. I changed dependency scope to "test"

ignite/pom.xml
Z-z-z-z: done

ignite/pom.xml
Z-z-z-z: done

ignite/pom.xml
Z-z-z-z: yes. it's necessary
done

ignite/pom.xml
Z-z-z-z: done

ignite/src/test/resources/org/hibernate/ogm/datastore/ignite/config.xml
Z-z-z-z: It is optimization for tests. The default cache configuration requires too much memory.

ignite/src/test/resources/hibernate.properties
Z-z-z-z: done

ignite/src/test/resources/hibernate.properties
Z-z-z-z: done

ignite/src/test/resources/hibernate.properties
Z-z-z-z: done

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/impl/IgniteDatastoreProvider.java
Z-z-z-z: When grid has't started yet.

				catch (IgniteIllegalStateException iise) {
					//not found, then start

Z-z-z-z:
Hi!
Could you take another code review. 
I think I fixed issues you wrote.

DavideD:
@Z-z-z-z  Thanks a lot!

I'll try to have a look at it today

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/impl/IgniteDatastoreProvider.java
DavideD: Wouldn't be better to have this as configuration option?

DavideD:
I went through the PR @Z-z-z-z, well done!

Some notes:
* Exception should be defined using `org.hibernate.ogm.datastore.ignite.logging.impl.Log`
* I would wait to work on the queries until after the core funcionality are ready
* Please, try toavoid including classess that are not used in the PR, it slows down the review process
* Could you add some Javadoc/comments explaining a bit what's going on with the different new classes and in the dialect? Many classes don't have any comment and maybe you could at least add the author tag.

I've created a new branch on my repository here: https://github.com/DavideD/hibernate-ogm/tree/ApacheIgnite

It starts with your PR but it removes all the classes that should not be needed, move the exception in the Log following the approach we use in all our projects and disable some tests that are not suppose to work at the moment.

Could you continue your work on top of that branch?
Ideally you could run the maven build and go through the tests that are still failing, I think the majority are failing because the dialect does not support composite keys. Any particular reason you haven't implemented that yet?

I've tried to remove only the things that I think aren't needed, there might be some mistakes, let me know.

DavideD:
I've decided to remove the parsing of the queries for now, the approach you used is not the one that we use in OGM, I would prefer to work on it after the other tests are fixed. Note that queries are working at the moment because it uses Hibernate Search.

DavideD:
By the way, I don't think you have to re-implement the JPA interfaces, you should be able to add the missing pieces using the dialect and other OGM specific interfaces.

DavideD:
By the way:
* would it make sense to pass the GridName as a configuration property? Instead of obtaining it from the url?
* Is there some tool to check the content of Ignite?

Thanks a lot

Z-z-z-z:
Hi David!

I'm using Ignite Dialect in my projects, but without queries it completely useless.
And I can't support two branches simultaneously. That's why I'm still working in my fork.

Tell me what's wrong with queries implementation? I'll try to improve code.

I almost fixed your comments about using Log for exceptions, unused classes and merge code with latest OGM verson.

If you have time take another look at the code.

Thank you

DavideD:
I'll have a look this week

DavideD:
> And I can't support two branches simultaneously. That's why I'm still working in my fork.

When working with git is quite common to have several branches, what I'm asking you to do is to rebase your branch on top of mine. This way you can continue from the changes that I've made to clean the code and rebase it to the latest branch. We also not merge branches but rebase them making sure that the history appears linear. And we don't merge PR if they are not rebased properly.

If you create a branch starting from my branch `ApacheIgnite` on my repository you should be good to continue to work.

> Tell me what's wrong with queries implementation? I'll try to improve code.

We have several classes to implement to convert HQL to queries that Ignite can understand. You can have a look at the classes under the package `org.hibernate.ogm.datastore.mongodb.query.parsing` or `org.hibernate.ogm.datastore.neo4j.query` for inspiration. But the point is that it would be better to first fix the CRUD operations and then work on the queries. Even if you don't implement the query parser you can still execute queries using the hibernate search integration and using @Index and @Field annotation (although it might not fit your use case).

Z-z-z-z:
Hi David!
I believe I made a rebase to your branch.
Could you please take a look at this because I'm not sure that it is what you want from me.

Thank you

DavideD:
Thanks Victor,
I'll have a look soon

DavideD:
> I believe I made a rebase to your branch.

@Z-z-z-z No, you didn't :)

So, what I'm asking you to do is this:

1. Create a local branch on your machine of my branch `ApacheIgnite`.
   You will have to do something like this:
   ```
   git remote add DavideD git@github.com:DavideD/hibernate-ogm.git
   git checkout DavideD/ApacheIgnite
   git checkout -b DavideD-ApacheIgnite
   ```

   If you follow this steps you will have a local copy of my branch on your machine called `DavideD-ApacheIgnite` 

2. Now you can rebase your branch on top of mine:
   ```
   git checkout master
   git rebase DavideD-ApacheIgnite
   ```

   You will have some conflicts that you have to solve, once you have done that, your branch will be correctly rebased


I hope this help and that now it is clearer what I'm asking to do.
If you are not familiar with the concept of rebase you can find a good explanation on this page:
https://git-scm.com/docs/git-rebase

Z-z-z-z:
I read documentation about git rebase and I understand the idea.
The only problem is to make it.
Thank you David. I'll try again :)


2016-10-27 18:12 GMT+03:00 DavideD <notifications@github.com>:

Z-z-z-z:
Hi David, 
I made rebase again. Is it ok now? :)

DavideD:
Mmmh...I can see you haven't done a rebase because the history is not linear
And when I try to rebase it I have many conflicts.
Which commands did you run exactly?

I can create a new branch and rebase everything for you this time, but then you need to continue working on the new branch I create or won't be able to integrate the PR

DavideD:
So, I've create a new branch on my repository called [OGM-983-ApacheIgnite-integration] (https://github.com/DavideD/hibernate-ogm/tree/OGM-983-ApacheIgnite-integration).

I've squashed together all your commits plus some changes that I needed to apply to fix the checkstyle errors and to make everything compile.

Could you have look at it and make sure that your next changes are applied on top of that branch?

Tomorrow I will have a better look at the dialect and see how it works after the latest changes.

Thanks

Z-z-z-z:
Ok. Let's try to work with another branch from scratch
By the way for rebase I run exactly those commands you wrote. There were a
lot of conflicts and so on...

2016-11-04 1:37 GMT+03:00 DavideD <notifications@github.com>:

DavideD:
Nice, would be also possible to use the latest version?

Z-z-z-z:
Hi, David

Some time ago I made new branch started from your's
Here it is
https://github.com/Z-z-z-z/hibernate-ogm/tree/OGM-983-ApacheIgnite
Now all basic tests passed successfully
"org.hibernate.ogm.backendtck.id.*",
"org.hibernate.ogm.backendtck.simpleentity.*",
"org.hibernate.ogm.backendtck.associations.*" (95%)

Review code please if you have time.

Thank you
Victor


2016-11-07 18:05 GMT+03:00 DavideD <notifications@github.com>:

> Nice, would be also possible to use the latest version?
>

DavideD:
Thanks Victor,
I'll try to have a look at it before the end of the week.

DavideD:
Sorry Victor,
I haven't yet, I'll look at it today

Davide

DavideD:
Please Victor, send a new Pull request so that I can add comments to it

Z-z-z-z:
Done
https://github.com/hibernate/hibernate-ogm/pull/804

2016-12-13 3:18 GMT+03:00 DavideD <notifications@github.com>:

PR #629, opened 2016-02-24
OGM-947 Implement Redis Cluster support

mp911de:
Add support for Redis Cluster. Upgrade lettuce driver version. Document how to test OGM Redis with Redis Standalone and Redis Cluster. Create Makefile for Redis setup. Create a specific forEachTuple test since the cluster scan requires a scan over all master nodes.

The cluster routing is handled by the lettuce driver library, so mainly the interface changed to a smaller type which is implemented by the Cluster and the Standalone API. The lettuce driver cannot handle a cluster-wide SCAN command that is required to iterate over all keys in the Redis keyspace. This is done within the Redis dialect code and requires to keep track of the iteration state.

Open issues:
- [ ] Review
- [x] Documentation


Sanne:
@Hibernate-CI add to whitelist

gunnarmorling:
Hey @mp911de, great stuff! How is the data going to be distributed in the cluster; Will all nodes be replicas essentially? In a way I'd expect some means of configuration on our side which allows to configure stuff like the number of replicas and/or shards to store a given entity on etc. This also can happen as a follow-up, I'd just like to better understand what a user gains from the clustering support in the proposed form.

redis/src/main/java/org/hibernate/ogm/datastore/redis/AbstractRedisDialect.java
```diff
@@ -316,4 +358,126 @@ protected void storeAssociation(
...
+
+			if ( nodeIds.isEmpty() ) {
+				return scan( connection, cursor, scanArgs );
+			}
+
+			currentNodeId = nodeIds.get( 0 );
+		}
+		else {
+			if ( !( cursor instanceof ClusterwideKeyScanCursor ) ) {
+				throw new IllegalArgumentException(
```
gunnarmorling: When would that happen? Could you use the `LOG` facade for raising the exception?

redis/src/main/java/org/hibernate/ogm/datastore/redis/AbstractRedisDialect.java
```diff
@@ -316,4 +358,126 @@ protected void storeAssociation(
...
+				nodeKeyScanCursor
+		);
+	}
+
+	private String getNodeIdForNextScanIteration(
+			List<String> nodeIds,
+			ClusterwideKeyScanCursor<String> clusterKeyScanCursor) {
+		if ( clusterKeyScanCursor.isScanOnCurrentNodeFinished() ) {
+			if ( clusterKeyScanCursor.isFinished() ) {
+				throw new IllegalStateException( "Cluster scan is finished" );
```
gunnarmorling: Same here.

redis/Makefile
```diff
@@ -0,0 +1,98 @@
+
+REDIS_VERSION:=3.0.7
```
gunnarmorling: Wow, tough piece :) Did you write it from scratch? I suppose at least the ports should be pulled into variables/parameters?

All in all I feel we really need to look into using Docker for setting up stuff like that...

mp911de:
Hi @gunnarmorling,

Redis Cluster is a bunch of Redis nodes with a configuration about sharding. The key is `CRC16 hashed % 16384` (simplified version) and the result is called a *slot*. Nodes are either a master or a slave. Master nodes are serving read/write for one or more slots, slaves have exactly one master but you're not required to run Redis cluster with slaves.

You don't have control over replica count. You can use specific prefixes to group data on one particular node but that's part of the key. Guess adding config infrastructure to configure the key/key pattern would be interesting, but not only in terms of Redis Cluster.

redis/Makefile
mp911de: Nah, I had some templates :)

redis/src/main/java/org/hibernate/ogm/datastore/redis/AbstractRedisDialect.java
mp911de: That's only an additional guard. It _should_ not happen but you know how it goes with _should_.
I'll change the Exception handling. The dialect API is not intended for outside use the guard is not necessary.

mp911de:
Updated the PR, added docs and some tests to illustrate hash key calculation. Maybe we should rethink about composite key representation. Switching from JSON to the Couchbase-like type of plain string concatenation (or similar) helps to avoid clashes with [Redis Cluster hash tagging](http://redis.io/topics/cluster-spec#keys-distribution-model)

gunnarmorling:
Hey @mp911de, when running `make test`, I'm getting this error:

    testScan(org.hibernate.ogm.datastore.redis.RedisDialectClusterForEachTest)  Time elapsed: 0.01 sec  <<< ERROR!
        java.lang.IllegalArgumentException: resource redis-cluster-slothashes.txt not found.
	    at com.google.common.base.Preconditions.checkArgument(Preconditions.java:145)
    	at com.google.common.io.Resources.getResource(Resources.java:197)
	    at org.hibernate.ogm.datastore.redis.RedisDialectClusterForEachTest.testScan(RedisDialectClusterForEachTest.java:46)

Have you seen that one before?

redis/Makefile
gunnarmorling: Ok, I see. Was/is there a copyright notice on the original file we should apply here, too?

More generally, introducing the make file doesn't make me super-happy: it adds another thing to know and be aware of, and it won't easily work on Windows (usage of curl etc.) I'm ok with adding it temporarily for now, but we should replace this with some Maven build steps subsequently. Looking at the performed steps, it shouldn't be too difficult to do.

gunnarmorling:
> Maybe we should rethink about composite key representation. Switching from JSON to the Couchbase-like type of plain string concatenation (or similar) helps to avoid clashes with Redis Cluster hash tagging.

Yes, that's a good point. Do you think it should be part of this PR? Or do it separately? Related to this, there should be a way to explicitly define the hash tag (that can be handled separately for sure).

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
```diff
@@ -69,6 +70,11 @@ Defines the timeout used by the driver
 when the connection to the Redis instance is initiated.
 This configuration is expressed in milliseconds.
 The default value is `5000`.
+hibernate.ogm.redis.cluster::
+Boolean flag to enable Redis Cluster mode.
+It's strongly recommended to use simple primary and association keys
```
gunnarmorling: Can you add a new JIRA issue and point to it here should you decide to do this not as part of this PR?

redis/Makefile
```diff
@@ -0,0 +1,98 @@
...
+test-standalone:
+	$(MAKE) redis-start
+	sleep 1
+	REDIS_HOSTNAME=localhost:6479 REDIS_CLUSTER=false mvn clean install
+	$(MAKE) redis-stop
+
+test-cluster:
+	$(MAKE) cluster-init
+	sleep 1
+	REDIS_HOSTNAME=localhost:7379 REDIS_CLUSTER=true mvn clean install
```
gunnarmorling: How about passing all the nodes here, to ensure that this actually works?

redis/src/main/java/org/hibernate/ogm/datastore/redis/AbstractRedisDialect.java
```diff
@@ -42,13 +53,49 @@
...
+	protected final RedisClusterCommands<String, String> connection;
+	protected final JsonSerializationStrategy strategy = new JsonSerializationStrategy();
+	protected final boolean clusterMode;
 
-	public AbstractRedisDialect(RedisConnection<String, String> connection) {
+	/**
+	 * Creates a new {@link AbstractRedisDialect}.
+	 *
+	 * @param connection a Redis connection (A regular Redis connection implements also {@link RedisClusterCommands})
+	 * @param configuredForCluster {@literal true} if the connection is configured for cluster operations.
```
gunnarmorling: Please make it "code" instead of "literal", as that's what we ususally use.

redis/src/main/java/org/hibernate/ogm/datastore/redis/AbstractRedisDialect.java
```diff
@@ -316,4 +366,119 @@ protected void storeAssociation(
 			}
 		}
 	}
+
+	/**
+	 * Scan over keys. This method is aware whether the the client is connected to a Redis Cluster.
```
gunnarmorling: "the the" = "the"

redis/src/main/java/org/hibernate/ogm/datastore/redis/AbstractRedisDialect.java
```diff
@@ -316,4 +366,119 @@ protected void storeAssociation(
...
+			KeyScanCursor<String> cursor,
+			ScanArgs scanArgs) {
+		if ( cursor != null ) {
+			return commands.scan( cursor, scanArgs );
+		}
+		return commands.scan( scanArgs );
+	}
+
+	@SuppressWarnings("unchecked")
+	private KeyScanCursor<String> clusterScan(KeyScanCursor<String> cursor, ScanArgs scanArgs) {
```
gunnarmorling: So scanning across all the masters sounds awefully inperformant :) Maybe we can find a way to limit the number of master nodes one entity type is sharded across, so we can limit the scan to those? Another consideration would be parallelization of this scan (e.g. different master nodes could be queried in parallel).

redis/src/main/java/org/hibernate/ogm/datastore/redis/AbstractRedisDialect.java
```diff
@@ -316,4 +366,119 @@ protected void storeAssociation(
...
+	}
+
+	/**
+	 * @return {@literal true} if the connected Redis node (default connection) is running in Redis Cluster mode.
+	 */
+	public boolean isClusterMode() {
+		return clusterMode;
+	}
+
+	static class ClusterwideKeyScanCursor<K> extends KeyScanCursor<K> {
```
gunnarmorling: I have to admit I'm a bit confused by the design of these cursors ;)

I would somehow have expected the cursor type(s) to have methods such as hasNext(), next() instead of getKeys(). Also couldn't the `ClusterwideKeyScanCursor` implementation keep all the required state and logic (e.g. `getNodeIdForNextScanIteration()`) internally, so instead of repeatedly invoking `scan()` which returns a new cursor instance each time, there'd be a single invocation returning a single cursor, which then would allow to iterate over all the rows.

WDYT?

redis/src/test/java/org/hibernate/ogm/datastore/redis/RedisDialectClusterForEachTest.java
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @author Mark Paluch
+ */
+public class RedisDialectClusterForEachTest extends RedisOgmTestCase {
+
+	@Test
+	public void testScan() throws Exception {
+
+		AbstractRedisDialect dialect = RedisTestHelper.getDialect( getProvider() );
+		assumeTrue( dialect.isClusterMode() );
```
gunnarmorling: Nice one, wasn't aware of that :)

redis/src/test/java/org/hibernate/ogm/datastore/redis/impl/TestRedisDatastoreProvider.java
```diff
@@ -37,8 +39,26 @@ public void run() {
...
+		return (RedisClient) staticInstance;
+	}
+
+	@Override
+	protected RedisClusterClient createClusterClient(Hosts hosts) {
+		if ( staticInstance == null ) {
+			final RedisClusterClient redisClient = super.createClusterClient( hosts );
+
+			Runtime.getRuntime().addShutdownHook(
+					new Thread( "RedisClient shutdown hook" ) {
```
gunnarmorling: That's interesting, shouldn't this happen in the `DatastoreProvider#stop()` method?

gunnarmorling:
Hi @mp911de, did another pass, see my inline comments. Some of the things can happen separately, could you open JIRA issues for them?

Thanks a lot for this, it's a great move towards proper cluster support. Very excited to see that!

mp911de:
I'll do so. Thanks for reviewing, will take a look and fix the findings within the next days.

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
mp911de: Done, https://hibernate.atlassian.net/browse/OGM-998

redis/src/main/java/org/hibernate/ogm/datastore/redis/AbstractRedisDialect.java
mp911de: At 25k Ops/sec slow gets relative. I didn't want to come up with async in the first place but it's possible. Lettuce is async under the hood so we could use `Future`s to query run queries on all masters. To control where a particular entity lives, we would need hash tagging but hash tagging is done by changing the key. Not sure whether I like the idea because it may lead to some surprise when looking directly into Redis.

redis/src/main/java/org/hibernate/ogm/datastore/redis/AbstractRedisDialect.java
mp911de: The current code is aligned to the Redis API. Drawing a Cursor API would encapsulate the Redis API and expose an `Iterator` (or similar). Taking async into account this might be the better solution since the Cursor hides the traversing and from a user perspective (well, Dialect developer to be precise), it's nicer to use.

redis/src/test/java/org/hibernate/ogm/datastore/redis/impl/TestRedisDatastoreProvider.java
mp911de: The `DatastoreProvider#stop()` is called after each test and we want to keep the `RedisClient` instances to prevent long test durations caused by the lengthy client shutdown.

mp911de:
About `java.lang.IllegalArgumentException: resource redis-cluster-slothashes.txt not found`: I missed a file while committing *d'oh*

mp911de:
I created https://hibernate.atlassian.net/browse/OGM-998 to revisit the key pattern. We might still have an issue if someone uses curly braces within the primary key but you might want to do this intentionally so data is stored on a particular node. The other side is, hash tagging then still applies even if you didn't use curly braces intentionally. There's currently no way how to disable hash tagging when curly braces are part of a key.

I'll add async to the key scanning and rework the cursor to expose an `Iterator`. The async part just requires to use the async API of lettuce, we don't need any thread-pools inside the Redis dialect.

redis/Makefile
mp911de: It's taken from my code which is MIT licensed. Do I need to add the OGM license header to this file?

gunnarmorling: Maybe you just can re-license the file under LPGL? If it's yours, that should be no problem.

redis/src/main/java/org/hibernate/ogm/datastore/redis/AbstractRedisDialect.java
gunnarmorling: Yes, that's essentially what I had in mind. It'd require some changes in your driver.

redis/src/test/java/org/hibernate/ogm/datastore/redis/impl/TestRedisDatastoreProvider.java
gunnarmorling: Got it.

gunnarmorling:
Rebased and merged. Thanks, @mp911de!

gunnarmorling:
Hey @DavideD, with this one in we're good for CR1 IMHO, if you want to do one today.

mp911de:
I will send you the makefile license header PR today. Wasn't aware you want
to merge this one today, sorry.
Gunnar Morling <notifications@github.com> schrieb am Do., 24. März 2016 um
13:29:

PR #631, opened 2016-03-13
OGM-988 Change log message id range from 16xx to 17xx

mp911de:


gunnarmorling:
Merged; Thanks, @mp911de!

PR #632, opened 2016-03-15
OGM-989 Align Narayana and JTA API versions to those of WildFly 10 an…

Sanne:
…d Infinispan 8

https://hibernate.atlassian.net/browse/OGM-989

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -41,7 +41,7 @@
         <hibernateSearchVersion>5.5.1.Final</hibernateSearchVersion>
         <hibernateParserVersion>1.2.0.Final</hibernateParserVersion>
         <hibernateCommonsAnnotationsVersion>5.0.1.Final</hibernateCommonsAnnotationsVersion>
-        <jbossjtaVersion>4.16.4.Final</jbossjtaVersion>
+        <narayanaVersion>5.2.14.Final</narayanaVersion>
         <jbossJtaAPIVersion>1.0.0.Final</jbossJtaAPIVersion>
         <jbossLoggingVersion>3.2.1.Final</jbossLoggingVersion>
         <jbossLoggingProcessorVersion>1.2.0.Final</jbossLoggingProcessorVersion>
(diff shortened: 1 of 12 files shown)


Sanne:
Adding also the Infinsipan upgrade as it depends on this:
 - https://hibernate.atlassian.net/browse/OGM-992

bom/pom.xml
```diff
@@ -543,135 +543,9 @@
...
-                    <exclusion>
-                        <artifactId>jboss-logging-processor</artifactId>
-                        <groupId>org.jboss.logging</groupId>
-                    </exclusion>
-                    <exclusion>
-                        <artifactId>jboss-logging-generator</artifactId>
-                        <groupId>org.jboss.logging</groupId>
-                    </exclusion>
-                </exclusions>
+                <groupId>org.jboss.narayana.jta</groupId>
```
gunnarmorling: Nice, seems that POM finally is sane now :)

bom/pom.xml
```diff
@@ -24,7 +24,7 @@
         <!-- compile/runtime dependency versions, managed in the BOM -->
         <classmateVersion>1.0.0</classmateVersion>
         <ehcacheVersion>2.6.9</ehcacheVersion>
-        <infinispanVersion>8.0.2.Final</infinispanVersion>
+        <infinispanVersion>8.2.0.Final</infinispanVersion>
```
gunnarmorling: How did you pick these versions? As per the [WF 10 POM](https://github.com/wildfly/wildfly/blob/10.0.0.Final/pom.xml#L150..L166), it's 5.2.12.Final for Narayana and 8.1.0.Final for ISPN.

gunnarmorling:
Looks good, only let's reassure we are using the correct versions.

bom/pom.xml
Sanne: N.B. my goal of "align with WildFly 10" applies to the JTA API only.

Narayana 5.2.12 is indeed what was in WF 10.0.0.Final, but the 5.2.14.Final is now going to replace it.. micro bugfix release so it shouldbn't matter for us.
I can make it 5.2.12 if you prefer ..

Regarding Infinispan: in this case we do not follow WildFly versions but track the Infinispan community, so that upgrade is a different JIRA; I included it in this PR as it requires the JTA upgrade.

gunnarmorling: Yes, the micro difference for Narayana should be fine.

Regarding ISPN, let's have a chat.

Sanne:
forced to swich to Infinispan 8.1 instead

gunnarmorling:
Merged; Thanks, @Sanne!

PR #633, opened 2016-03-16
OGM-969 MapTest fix

DavideD:
https://hibernate.atlassian.net/browse/OGM-969

### DIFF
--- a/core/src/test/java/org/hibernate/ogm/backendtck/associations/collection/types/User.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/associations/collection/types/User.java
@@ -70,6 +70,7 @@ public class User {
 	@OneToMany
 	@MapStorage(MapStorageType.AS_LIST)
 	@MapKeyColumn(name = "phoneType")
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/hibernate/ogm/backendtck/associations/collection/types/MapTest.java
```diff
@@ -34,6 +37,23 @@
 public class MapTest extends OgmTestCase {
 
 	@Test
+	@TestForIssue(jiraKey = "OGM-969")
+	// ORM does not support mapping different associations of the same entity using the same table.
+	// Each association must have a different table. For this reason we decided to update the tests and rename the tables
+	// I'll leave this here just in case someone decides to change the test.
+	public void testDefaultAssociationKeyMetadataEquals() throws Exception {
```
gunnarmorling: Can we move that out of the TCK? I don't think it needs to run for every dialect.

gunnarmorling:
Merged; Thanks, @DavideD!

PR #634, opened 2016-03-16
OGM-993 Modules for Infinispan should use the Infinispan community modules rather than the WildFly ones

Sanne:
https://hibernate.atlassian.net/browse/OGM-993

### DIFF
--- a/documentation/manual/src/main/asciidoc/en-US/modules/configuration.asciidoc
+++ b/documentation/manual/src/main/asciidoc/en-US/modules/configuration.asciidoc
@@ -373,7 +373,7 @@ Add a `WEB-INF/jboss-deployment-structure.xml` in your archive with the followin
 ----
 
 More information about the descriptor can be found in the
-https://docs.jboss.org/author/display/WFLY9/Class+Loading+in+WildFly[WildFly documentation].
+https://docs.jboss.org/author/display/WFLY10/Class+Loading+in+WildFly[WildFly documentation].
 
 ==== Enabling both the Hibernate Search and Hibernate OGM modules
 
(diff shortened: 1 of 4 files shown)


gunnarmorling:
Nice; Should we add a note how users would update the slot OGM depends on? Or just say "OGM uses version x.y.z, please add the right module ZIP for that version"?

Sanne:
Were would you have that note? BTW not sure if you noticed but I'm having it link directly to the right zip.

gunnarmorling:
> Were would you have that note?

In the ref guide, within the section you added.

> I'm having it link directly to the right zip.

Yes, what I mean is, say ISPN 8.3 is out, and people want to use that. So
they'd have to update the referenced slot in the hibernate-ogm-infinispan
module.xml, too, IIUC. Or, we'd just say that we only support that single
version mentioned in the ref guide and that's it.

Sanne:
> Or, we'd just say that we only support that single
version mentioned in the ref guide and that's it.

I'd avoid to mention "support" but I'll try to clarify :)

Sanne:
pushed a new version with more verbose docs

documentation/manual/src/main/asciidoc/en-US/modules/configuration.asciidoc
```diff
@@ -403,3 +403,22 @@ The Hibernate Search documentation explains the details of downloading and deplo
...
+so a micro version update should be safe but is untested.
+
+In case you want to experiment with a more significant version upgrade, you will need to edit the modules of Hibernate OGM:
+the module identifiers are hardcoded in the XML files representing the module.
+
+Download the Infinispan modules pack for WildFly 10 from here:
+
+ * https://repo1.maven.org/maven2/org/infinispan/infinispan-as-embedded-modules/{infinispan-version}/infinispan-as-embedded-modules-{infinispan-version}.zip[Infinispan WildFly modules version {infinispan-version} from the Maven repository]
+
+Then similarly to what you did with the Hibernate OGM mozules zip, unpack this one too in your `modules` directory within the application server.
```
DavideD: mo**z**ules

Sanne: thanks! fixed

Sanne:
added one more commit, for the depending Infinispan upgrade to 8.2 :
https://hibernate.atlassian.net/browse/OGM-994

gunnarmorling:
Nice change! Merge yourself if you like? Otherwise I'll do it tomorrow. Thanks!

Sanne:
great, thanks merged.

PR #636, opened 2016-03-17
[PREVIEW] OGM-366 Transaction context 

DavideD:
Previous pull request: https://github.com/hibernate/hibernate-ogm/pull/623

This is a first step towards the *Neo4j Remote*, I need to get the transaction id and use it with the Rest API on each call.
Following the previous remarks I've added the `TransactionContext` to the `TupleContext` and `AssociationContext` and I've added a `QueryContext` for the dialect methods to execute queries.

@gunnarmorling For now I prefer to pass the `TransactionDirver` with the TransactionContext
I'm not completely sold on the idea of using the `get(TransactionId.class)` as it seems to add complexity that we don't need at the moment. Also I'm not ure how to pass the transctionId to the context since the TransactionDriver does not have a method to obtain it.
@emmanuelbernard Any opinion about passing the TransactionDriver around? You can see the starting of the conversation  here: https://github.com/hibernate/hibernate-ogm/pull/623#discussion-diff-50811891R16

An example of how this will be used is:
```
Long txId = transactionId( tupleContext.getTransactionContext());
        ...
private Long transactionId(TransactionContext context) {
    return ( (Neo4jTransactionDriver) context.getTransactionDriver() ).getTransactionId();
}
	
```

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/dialect/spi/TransactionContext.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/spi/TransactionContext.java
@@ -1,0 +1,17 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.dialect.spi;
+
+import org.hibernate.resource.transaction.TransactionCoordinator.TransactionDriver;
+
+/**
+ * @author Davide D'Alto
+ */
+public interface TransactionContext {
+
+	TransactionDriver getTransactionDriver();
+}
--- a/cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
+++ b/cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/CassandraDialect.java
@@ -44,6 +44,7 @@ import org.hibernate.ogm.dialect.spi.BaseGridDialect;
 import org.hibernate.ogm.dialect.spi.GridDialect;
 import org.hibernate.ogm.dialect.spi.ModelConsumer;
 import org.hibernate.ogm.dialect.spi.NextValueRequest;
+import org.hibernate.ogm.dialect.spi.QueryContext;
 import org.hibernate.ogm.dialect.spi.TupleAlreadyExistsException;
 import org.hibernate.ogm.dialect.spi.TupleContext;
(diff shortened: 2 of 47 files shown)


DavideD:
Rebased

gunnarmorling:
The issue I see with exposing `TransactionDriver` is that enables `commit()`, `rollback()` and the like.

DavideD:
I've added a commit that should remove The `TransactionDriver`

DavideD:
Jenkins, rebuild this please

core/src/main/java/org/hibernate/ogm/dialect/spi/TransactionContext.java
```diff
@@ -0,0 +1,15 @@
...
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.dialect.spi;
+
+/**
+ * @author Davide D'Alto
+ */
+public interface TransactionContext {
+
+	Long getTransactionId();
```
gunnarmorling: It's guess-work at this point, but I feel limiting it to `Long` might be too tight. I'd make it `String` at least, better yet `Object` to be prepared for other implmentations.

gunnarmorling:
@DavideD, can you add the implementation for Neo4j, too? That way you will be sure that the mechanism actually works and you can even add some tests based on Neo4j.

Apart from that and my comment above it looks good. Yeah :)

core/src/main/java/org/hibernate/ogm/dialect/spi/TransactionContext.java
DavideD: Ok, let's go with Object for now and see how it works

DavideD:
> @DavideD, can you add the implementation for Neo4j, too? 

You mean Neo4j Remote? I've already tested this solution. I'll send the rest of the PR as soon as I applied the previous remarks. I'm also trying to split it more to make the review process less of a pain.

> Apart from that and my comment above it looks good. Yeah :)

Yeah :)

DavideD:
By the way, @Sanne suggested to start the transaction when we execute the first operation on the dialect. I was wondering how to do that, should I use something similar to what we do for the creation of the `OpeariontQueue`? Create a `TransactionContext` for datastore providers who support transactions  and use the `Synchronization` interface to rollback and commit that transaction?

DavideD:
I guess it can also be an improvement for later as we will still need the same interfaces.

DavideD:
It seems we found an agreement on the interface, I'm closing this for now

gunnarmorling:
Shouldn't we integrate it? First time we agree on something and still not apply it :)

PR #638, opened 2016-03-23
OGM-366 Neo4j remote

DavideD:
Follows the discussion here:
https://github.com/hibernate/hibernate-ogm/pull/623#discussion-diff-50811891R16

The pull request is quite big as before but it is possible check it one commit at a time, the first commits are related to the transaction pull requests: I've sent before: https://github.com/hibernate/hibernate-ogm/pull/636

I'll send these as separate PRs, I need to check if they are feasible:

> Can you check what @mp911de does in the pending Redis PR? There two Surefire executions are configured. Would be nice to do the same here.

> Can we validate that a user name + PW is given? Otherwise Neo4j complains a bit cryptically, better to catch it on our side.


gunnarmorling:
@Sanne Are you up for the review challenge again?

Sanne:
> I'll send these as separate PRs, I need to check if they are feasible

@DavideD what is "these" you are referring to?

> @Sanne Are you up for the review challenge again?

@gunnarmorling Sure, but I hope you'll all stop making massive changes :)

DavideD:
@Sanne The follow up quoted questions:

>     Can you check what @mp911de does in the pending Redis PR? There two Surefire executions are configured. Would be nice to do the same here.
> 
>     Can we validate that a user name + PW is given? Otherwise Neo4j complains a bit cryptically, better to catch it on our side.
>

core/src/main/java/org/hibernate/ogm/dialect/impl/IdentifiableDriver.java
```diff
@@ -0,0 +1,17 @@
...
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.dialect.impl;
+
+import org.hibernate.resource.transaction.TransactionCoordinator.TransactionDriver;
+
+/**
+ * @author Davide D'Alto
+ */
+public interface IdentifiableDriver extends TransactionDriver {
```
Sanne: The name and/or javadoc could use some improvement, as I only get some hint of what "IdentifiableDriver" means when looking at what it extends.

core/src/main/java/org/hibernate/ogm/dialect/impl/TransactionContextImpl.java
```diff
@@ -0,0 +1,28 @@
...
+ * Default implementation of the {@link TransactionContext}.
+ *
+ * @author Davide D'Alto
+ */
+public class TransactionContextImpl implements TransactionContext {
+
+	private final Object transactionId;
+
+	public TransactionContextImpl(IdentifiableDriver driver) {
+		this.transactionId = driver.getTransactionId();
```
Sanne: The Neo4J RemoteJtaTransactionCoordinator seems to use a lazy initialization for getTransactionId().
Is it then a good idea to initialize it in a final field eagerly in this case?

core/src/main/java/org/hibernate/ogm/util/impl/TransactionContextHelper.java
```diff
@@ -0,0 +1,39 @@
...
+public final class TransactionContextHelper {
+
+	private TransactionContextHelper() {
+	}
+
+	public static TransactionContext transactionContext(Session session) {
+		return transactionContext( (SessionImplementor) session );
+	}
+
+	public static TransactionContext transactionContext(SessionImplementor session) {
```
Sanne: Could we document that this method could return `null`, and in which cases?
Should we create a "null TX" representation object rather than play with `null` having a valid meaning here?

core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
```diff
@@ -1741,8 +1744,17 @@ public String getSubclassForDiscriminatorValue(Object value) {
...
 
-	public TupleContext getTupleContext() {
-		return tupleContext;
+	/**
+	 * Returns the {@link TupleContext}.
+	 *
+	 * @param session the current session, if null the {@link TupleContext#getTransactionContext()} will be null.
+	 * @return the tupleContext for the session
+	 */
+	public TupleContext getTupleContext( SessionImplementor session ) {
```
Sanne: I'm wondering why you opted to require the `SeesionImplementor` in all those code paths (the ones invoking this method) when you're needing the transaction context.

Woudln't it have been possible to invoke that TransactionContextHelper less frequently?

core/src/main/java/org/hibernate/ogm/dialect/impl/IdentifiableDriver.java
DavideD: ok

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/transaction/impl/RemoteTransactionCoordinatorBuilder.java
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	@Override
+	public TransactionCoordinator buildTransactionCoordinator(TransactionCoordinatorOwner owner, TransactionCoordinatorOptions options) {
+		if ( isJta() ) {
+			TransactionCoordinator coordinator = super.buildTransactionCoordinator( owner, options );
+			return new RemoteJtaTransactionCoordinator( coordinator, datastoreProvider );
+		}
+		else {
+			TransactionCoordinatorOwner noopedOwner = new NoopJdbcResourceTransactionCoordinatorOwner( owner );
```
Sanne: Could you add some comments here to explain why a NoopJdbcResourceTransactionCoordinatorOwner is used as a real transaction owner?

core/src/main/java/org/hibernate/ogm/util/impl/TransactionContextHelper.java
DavideD: Yes, I think I can create an obkject that throws an exception in case the transaction id is null

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/transaction/impl/RemoteTransactionCoordinatorBuilder.java
Sanne: I mean: both "noop" and "jdbc" in the name seem smelly.

Also this implementation was a private class before this PR, so I'm wondering if it was really a good idea to make it public.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/transaction/impl/Neo4jLocalTransactionCoordinator.java
```diff
@@ -20,7 +20,7 @@
  *
  * Note that during a JTA transaction Neo4j {@link Transaction} are
  * synchronized using the {@link Synchronization} interface. A commit to the Neo4j transaction will happen before the
- * end of the JTA transaction, meaning that it won't be possible to rollback if an error happen after succesful commit
```
Sanne: if you're into fixing language, then:
 - "if an error happen" -> "if an error happens"
 - "after successful" -> "after a successful"

core/src/main/java/org/hibernate/ogm/type/impl/ByteStringType.java
```diff
@@ -28,7 +28,7 @@ public ByteStringType() {
 
 	@Override
 	public String getName() {
-		return "byte_integer";
+		return "byte_as_string";
 	}
```
Sanne: almost deserves a separate JIRA / PR ;)

I'm reviewing the 7th commit and no change so far seems related to OGM-366, you could have pulled all these out in smaller changes & quicker integrations.

core/src/main/java/org/hibernate/ogm/dialect/impl/TransactionContextImpl.java
DavideD: Mmmh....I will double check, but this class should be called when a transaction is already started and I'm expecting this to be already initialized. I should probably validate the identifier and make sure that it is not null.

core/src/main/java/org/hibernate/ogm/util/impl/ArrayHelper.java
```diff
@@ -86,4 +87,20 @@ public static boolean contains(Object[] array, Object element) {
 
 		return result;
 	}
+
+	public static <T> T[] concat(List<T[]> arrays) {
```
Sanne: no need for a unit test?
At least improve the commit message

core/src/main/java/org/hibernate/ogm/util/impl/ArrayHelper.java
```diff
@@ -86,4 +87,20 @@ public static boolean contains(Object[] array, Object element) {
...
+		int totSize = 0;
+		for ( int i = 0; i < arrays.size(); i++ ) {
+			totSize += arrays.get( i ).length;
+		}
+		@SuppressWarnings("unchecked")
+		T[] joined = (T[]) Array.newInstance( arrays.get( 0 ).getClass().getComponentType(), totSize );
+		int currentLength = 0;
+		for ( int i = 0; i < arrays.size(); i++ ) {
+			T[] entry = arrays.get( i );
+			System.arraycopy( arrays.get( i ), 0, joined, currentLength, entry.length );
```
Sanne: System.arraycopy( entry, 0, joined, currentLength, entry.length );

core/src/main/java/org/hibernate/ogm/util/impl/ArrayHelper.java
Sanne: Javadoc?

BTW the ideal polishing in my opinion would be to not need such things :) Are you sure we can't avoid all this copying around?

core/src/main/java/org/hibernate/ogm/dialect/impl/TransactionContextImpl.java
Sanne: Right, if you could add a check that would be great: I think it would be safer to have it throw an exception here in case the assumptions aren't holding.

core/src/main/java/org/hibernate/ogm/type/impl/BinaryMappedType.java
```diff
@@ -0,0 +1,38 @@
...
+import org.hibernate.ogm.type.descriptor.impl.BinaryMappedGridTypeDescriptor;
+import org.hibernate.type.descriptor.java.PrimitiveByteArrayTypeDescriptor;
+
+/**
+ * @author Davide D'Alto
+ */
+public class BinaryMappedType extends AbstractGenericBasicType<byte[]> {
+	public static final BinaryMappedType INSTANCE = new BinaryMappedType();
+
+	public BinaryMappedType() {
```
Sanne: I'd prefer to see all these constructors private when there's `INSTANCE`, but I see all types are having this.

core/src/main/java/org/hibernate/ogm/type/impl/ByteArrayStringType.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.hibernate.MappingException;
+import org.hibernate.engine.spi.Mapping;
+import org.hibernate.ogm.type.descriptor.impl.StringMappedGridTypeDescriptor;
+import org.hibernate.type.descriptor.java.ByteArrayTypeDescriptor;
+
+/**
+ * @author Davide D'Alto
+ */
+public class ByteArrayStringType extends AbstractGenericBasicType<Byte[]>  {
```
Sanne: s/  {/ {/

core/src/main/java/org/hibernate/ogm/type/impl/ByteMappedType.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.hibernate.MappingException;
+import org.hibernate.engine.spi.Mapping;
+import org.hibernate.ogm.type.descriptor.impl.ByteMappedGridTypeDescriptor;
+import org.hibernate.type.descriptor.java.ByteTypeDescriptor;
+
+/**
+ * @author Davide D'Alto
+ */
+public class ByteMappedType extends AbstractGenericBasicType<Byte> {
```
Sanne: `org.hibernate.ogm.type.impl.ByteType` vs `org.hibernate.ogm.type.impl.ByteMappedType` ?

DavideD: Right, wrong name choice.

`ByteType` uses `PassThroughGridTypeDescriptor.INSTANCE` and it passes the object wihout any intermediate step.

`ByteMappedType` uses `ByteMappedGridTypeDescriptor.INSTANCE` and it tries to convert the object to a Byte.

Sanne: Is it a problem that they both return the same string for `getName()` ?

DavideD: I'll change it

neo4j/pom.xml
```diff
@@ -72,6 +72,32 @@
...
+            <artifactId>jaxrs-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
```
Sanne: I've opened https://hibernate.atlassian.net/browse/OGM-1005 as a follow-up to this, as I believe we'll need these in the WildFly modules too.

DavideD: Yes, thanks

core/src/main/java/org/hibernate/ogm/type/impl/ByteMappedType.java
Sanne: Ok, but what about my question. What kind of consequences will we have if we have multiple types registering the same name? Should we add some protection against that?

gunnarmorling: I don't think the method is ever used. It has been inherited when initially copying from ORM's `Type` but OGM doesn't invoke it on `GridType`. Some clean-up of these contracts is overdue, many of the methods can go away.

core/src/main/java/org/hibernate/ogm/dialect/query/spi/BackendQuery.java
```diff
@@ -10,7 +10,7 @@
 
 /**
  * Represents a NoSQL query as to be executed via
- * {@link QueryableGridDialect#executeBackendQuery(BackendQuery, QueryParameters)}.
+ * {@link QueryableGridDialect#executeBackendQuery(BackendQuery, QueryParameters, org.hibernate.ogm.dialect.spi.QueryContext)}
```
gunnarmorling: No need for the FQN.

core/src/test/java/org/hibernate/ogm/backendtck/associations/manytoone/ManyToOneTest.java
```diff
@@ -42,8 +42,8 @@ public void testUnidirectionalManyToOne() throws Exception {
 		session.persist( emmanuel );
 		session.persist( jerome );
 		session.flush();
-		assertThat( getNumberOfEntities( sessions ) ).isEqualTo( 3 );
-		assertThat( getNumberOfAssociations( sessions ) ).isEqualTo( expectedAssociations() );
+		assertThat( getNumberOfEntities( session ) ).isEqualTo( 3 );
```
gunnarmorling: How comes this has changed now? It seems to actually use only the SF eventually anyways?

core/src/test/java/org/hibernate/ogm/backendtck/compensation/CompensationSpiTest.java
```diff
@@ -234,7 +234,7 @@ public void onRollbackPresentsAppliedUpdateOperations() throws Exception {
 
 	@Test
 	@SkipByGridDialect(
-			value = { GridDialectType.NEO4J, GridDialectType.INFINISPAN, GridDialectType.EHCACHE },
+			value = { GridDialectType.NEO4J, GridDialectType.NEO4J_REMOTE, GridDialectType.INFINISPAN, GridDialectType.EHCACHE },
```
gunnarmorling: I'm confused by the comment :) I thought Neo4j isn't using JTA?

core/src/test/java/org/hibernate/ogm/utils/TestableGridDialect.java
```diff
@@ -27,6 +28,13 @@
 	/**
 	 * Returns the number of entities in the datastore
 	 *
+	 * @param session
+	 */
+	long getNumberOfEntities(Session session);
```
gunnarmorling: Couldn't a new transaction be used for all these assertions with Neo4j Remote? That'd allow sticking to passing the SF as done before.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/transaction/impl/RemoteTransactionCoordinatorBuilder.java
```diff
@@ -0,0 +1,42 @@
...
+import org.hibernate.resource.transaction.TransactionCoordinator;
+import org.hibernate.resource.transaction.TransactionCoordinatorBuilder;
+import org.hibernate.resource.transaction.spi.TransactionCoordinatorOwner;
+
+/**
+ * Builder for a Neo4j remote {@link TransactionCoordinator}.
+ *
+ * @author Davide D'Alto
+ */
+public class RemoteTransactionCoordinatorBuilder extends ForwardingTransactionCoordinatorBuilder {
```
gunnarmorling: I'm wondering why this implementation is forwarding. What's the rationale behind it?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/RemoteNeo4jSchemaDefiner.java
```diff
@@ -0,0 +1,200 @@
...
+
+		Statements constraintStatements = new Statements();
+		provider.getSequenceGenerator().createSequencesConstraints( constraintStatements, sequences );
+		provider.getSequenceGenerator().createUniqueConstraintsForTableSequences( constraintStatements, idSourceKeyMetadata );
+
+		StatementsResponse response = provider.getDataStore().executeQueriesInNewTransaction( constraintStatements );
+		List<StatementResult> results = response.getResults();
+		// TODO: Check response for errors
+
+		// We create the sequencese in a separate transaction because
```
gunnarmorling: sequence

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/facade/impl/Neo4jTransactionFacade.java
```diff
@@ -0,0 +1,49 @@
...
+	@Path("/transaction/{transactionId}/commit")
+	Response commit(@PathParam("transactionId") long transactionId);
+
+	@DELETE
+	@Path("/transaction/{transactionId}")
+	Response rollback(@PathParam("transactionId") long transactionId);
+
+	@POST
+	@Path("/transaction/commit")
+	Response executeQuery(Statements statements);
```
gunnarmorling: The method name and path seem not to match?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/transaction/impl/Neo4jLocalTransactionCoordinator.java
gunnarmorling: I haven't figured yet how these pieces fit together. How does JTA play in here, given this is a "local transaction coordinator"?

gunnarmorling:
Hey @DavideD, how is it going with this one? Seems you still need to get back to some remarks from @Sanne? I added some comments as well, but it's too large to dive into all the details. I'm wondering whether you could extract some parts out of it, so it gets better digestable. E.g. the introduction of TX context could be its own PR (as you had it already) which can be merged (into a 5.1 branch), also if it's not yet used.

core/src/main/java/org/hibernate/ogm/util/impl/ArrayHelper.java
DavideD: ok

core/src/main/java/org/hibernate/ogm/type/impl/ByteStringType.java
DavideD: They are in separate commits, Isn't that enough?

Also, this is the only change that it's not required for OGM-366, but it is so small that I decided not to open a separate JIRA for it. I guess I will send a separate pull request or remove the commit

core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
DavideD: I need to add the transaction id to the TupleContext, I get the transaction id from the session.

I can pass the TransactionContext, instead of the session I guess, but I don't think the amount of time I need to call TransactionContextHelper will change.

I guess I could add an interface to identify which dialects need a transaction context, would that be better?

core/src/main/java/org/hibernate/ogm/type/impl/BinaryMappedType.java
DavideD: I suppose we can create a separate JIRA for this

core/src/test/java/org/hibernate/ogm/backendtck/associations/manytoone/ManyToOneTest.java
DavideD: For Neo4j remote I need the session to extract the transaction id

core/src/test/java/org/hibernate/ogm/utils/TestableGridDialect.java
DavideD: Some tests call this method before and after a rollback to make sure that count is correct.
It seems reasonable to execute this inside the same transaction.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/transaction/impl/RemoteTransactionCoordinatorBuilder.java
DavideD: We didn't need this anywhere else before.
This implementation make sure that there are not JDBC calles underneath but still update the status accordingly.
I will check if I can avoid it.

core/src/main/java/org/hibernate/ogm/util/impl/ArrayHelper.java
DavideD: For some queries I need to pass as parameters an array of columns the are not always in one key, Somtimes the values are in the AssociationKey, EntityKey and a third array. This method join these arrays in a unique one that I can pass as parameter. I will add the javadoc.

Not sure if there is a way to avoid it.
I will also open a JIRA to add unit tests for this class since they are missing for all the methods.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/facade/impl/Neo4jTransactionFacade.java
DavideD: This is the Ne4j Rest transaction API, if you want to execute some statements in one request inside a new transaction, you need to call the the /transaction/commit and pass the queries.

The path looks strange if you compare it to the method name, but the method name is telling you what's happening.

Maybe I will change the name to something like  `executeQueriesInNewTransaction`

gunnarmorling:
@DavideD, how should we proceed with this one? Do you plan to replace it with a new one? Or several smaller ones?

DavideD:
Closing this one, I will send new ones https://github.com/hibernate/hibernate-ogm/pull/638

PR #639, opened 2016-03-24
OGM-1003 Add license header to make file in Redis backend

mp911de:


mp911de:
@DavideD here's the missing license header PR

PR #642, opened 2016-03-24
OGM-860 Move DocumentHelpers to more sensible place

DavideD:
https://hibernate.atlassian.net/browse/OGM-860


DavideD:
Shouldn't we also move AssociationRow and AssociationRowFactory

gunnarmorling:
Rebased and applied; Thanks!

PR #644, opened 2016-03-29
OGM-711 Avoid exporting module org.jboss.jts to user classpath

Sanne:
The JTS module is no longer required, provided the correct JTA platform is configured in the integration tests.

It turns out the documentation already pointed out the need for this JTA platform to be set, so that was done already.

### DIFF
--- a/integrationtest/couchdb/src/test/java/org/hibernate/ogm/test/integration/jboss/CouchDBModuleMemberRegistrationIT.java
+++ b/integrationtest/couchdb/src/test/java/org/hibernate/ogm/test/integration/jboss/CouchDBModuleMemberRegistrationIT.java
@@ -77,6 +77,7 @@ public class CouchDBModuleMemberRegistrationIT extends ModuleMemberRegistrationS
 							.createProperty().name( "hibernate.ogm.datastore.database" ).value( "ogm_test_database" ).up()
 							.createProperty().name( "hibernate.ogm.datastore.create_database" ).value( "true" ).up()
 							.createProperty().name( "hibernate.search.default.directory_provider" ).value( "ram" ).up()
+							.createProperty().name( "hibernate.transaction.jta.platform" ).value( "JBossAS" ).up()
 					.up().up();
 	}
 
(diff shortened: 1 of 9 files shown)


Sanne:
pushed a new version as I had forgotten about the Redis tests

DavideD:
I think cassandra is  missing as well, I will update it don't change this anymore

DavideD:
I've added the change for cassandra squashed and rebased.
https://github.com/hibernate/hibernate-ogm/pull/644

integrationtest/couchdb/src/test/java/org/hibernate/ogm/test/integration/jboss/CouchDBModuleMemberRegistrationIT.java
```diff
@@ -77,6 +77,7 @@ private static PersistenceDescriptor persistenceXml() {
 							.createProperty().name( "hibernate.ogm.datastore.database" ).value( "ogm_test_database" ).up()
 							.createProperty().name( "hibernate.ogm.datastore.create_database" ).value( "true" ).up()
 							.createProperty().name( "hibernate.search.default.directory_provider" ).value( "ram" ).up()
+							.createProperty().name( "hibernate.transaction.jta.platform" ).value( "JBossAS" ).up()
```
gunnarmorling: Interesting, how comes that's needed now? Shouldn't Hibernate ORM figure it out by itself within WF?

Sanne: you're anticipating a draft I was writing to hibernate-dev :)

This was needed already, our documentation even mentions this requirement when running on WildFly.
So technically all I did is to align our integration tests to our docs, and doing so the JTS module was no longer needed.

I'm afraid the right platform lookup is being injected by jipijapa so not something we control, will check and then share the details ..

gunnarmorling: Ah, we should work with Scott then to get this sorted out. +1 for bringing it on par with the docs, but it shouldn't be needed eventually, which might require some enablement on the JipiJapa side.

PR #646, opened 2016-03-29
OGM-973 Do not ship an 'org.parboiled' WildFly module

Sanne:
- https://hibernate.atlassian.net/browse/OGM-973


gunnarmorling:
Rebased and applied; Thanks, @Sanne!

PR #650, opened 2016-03-31
OGM-1002 Move integration tests in the same project

DavideD:
https://hibernate.atlassian.net/browse/OGM-1002


Sanne:
oh man the first commit is super confusing :)
But the end result looks very nice, thanks!

PR #651, opened 2016-04-11
OGM-1017 Store BigDecimal as decimal in Cassandra

DavideD:
https://hibernate.atlassian.net/browse/OGM-1017

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/type/impl/BigDecimalPassThroughType.java
+++ b/core/src/main/java/org/hibernate/ogm/type/impl/BigDecimalPassThroughType.java
@@ -1,0 +1,62 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.type.impl;
+
+import java.math.BigDecimal;
+
+import org.hibernate.HibernateException;
+import org.hibernate.MappingException;
+import org.hibernate.engine.spi.Mapping;
+import org.hibernate.ogm.type.descriptor.impl.PassThroughGridTypeDescriptor;
+import org.hibernate.type.descriptor.java.BigDecimalTypeDescriptor;
+
+/**
+ * Type descriptor that will pass the {@link BigDecimal} to the dialect wihtout conversion.
+ *
+ * @see java.math.BigDecimal
+ * @see BigDecimalType
+ *
+ * @author Davide D'Alto
+ */
+public class BigDecimalPassThroughType extends AbstractGenericBasicType<BigDecimal> {
+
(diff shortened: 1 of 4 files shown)


documentation/manual/src/main/asciidoc/en-US/modules/cassandra.asciidoc
```diff
@@ -105,6 +105,7 @@ The type and field names can be overridden by annotations, but concatenation tok
 http://docs.datastax.com/en/cql/3.1/cql/cql_reference/cql_data_types_c.html[CQL3 types]
 
 CQL3 supports a smaller selection of numeric data types than Hibernate, so the missing types are autotmatically converted. byte and short values are promoted to integer.
```
gunnarmorling: autotmatically=automatically

gunnarmorling:
There should be a test for making sure that the mapped value is the one we expect. See `BuiltinTypeMappingTest` for some infra to use.

DavideD:
Ah nice, I thought we didn't have those kind of tests in cassandra yet. I'll add it

DavideD:
Am I wrong or the tests in that class are not really checking the type of the column? They are checking that the value returned is the expected one. how does it differ from the test in core?

DavideD:
Ah no, right, I get it. What I was expecting is a test to check that a particular column is of type "decimal" in cassandra.

DavideD:
Added two commits to fix the remarks

core/src/main/java/org/hibernate/ogm/type/impl/BigDecimalPassThroughType.java
```diff
@@ -0,0 +1,62 @@
...
+import java.math.BigDecimal;
+
+import org.hibernate.HibernateException;
+import org.hibernate.MappingException;
+import org.hibernate.engine.spi.Mapping;
+import org.hibernate.ogm.type.descriptor.impl.PassThroughGridTypeDescriptor;
+import org.hibernate.type.descriptor.java.BigDecimalTypeDescriptor;
+
+/**
+ * Type descriptor that will pass the {@link BigDecimal} to the dialect wihtout conversion.
```
gsmet: s/wihtout/without/

documentation/manual/src/main/asciidoc/en-US/modules/cassandra.asciidoc
```diff
@@ -104,7 +104,8 @@ The type and field names can be overridden by annotations, but concatenation tok
 
 http://docs.datastax.com/en/cql/3.1/cql/cql_reference/cql_data_types_c.html[CQL3 types]
 
-CQL3 supports a smaller selection of numeric data types than Hibernate, so the missing types are autotmatically converted. byte and short values are promoted to integer.
+CQL3 supports a smaller selection of numeric data types than Hibernate, so the missing types are automatically converted. byte and short values are promoted to integer.
```
gsmet: The capitalization/style here is weird. I would have expected either Byte or `byte`.

gunnarmorling:
Added a comment to address @gsmet's comments, rebased and applied. Thanks, @DavideD.

PR #653, opened 2016-04-14
OGM 975 Allow MongoDBDatastoreProvider subclasses to modify the configuration

aplopez:
This is only the core part. Users must provide a subclass of MongoDBDatastoreProvider modifying the configuration as they need. For instance, they could override retrieveClientOptions() to modify the configuration to use their own socket factory.

Following commits will update the tests and the documentation.


aplopez:
Pushed the alternative method, including two tests that additionally show how it works.

gunnarmorling:
@aplopez, could you sign the contributor license agreement, please? It's a web app living at https://cla.jboss.org/. That's needed in order to merge your work. Thanks!

aplopez:
It's done.

PR #657, opened 2016-05-03
OGM-1028 Use a proper logger instead of a Contracts assert

gsmet:
Include the entity toString in the message.


gunnarmorling:
Rebased and applied. Thanks, @gsmet!

gunnarmorling:
Ah, and congrats to first commit merged into the project!

PR #660, opened 2016-05-04
OGM-1030 Upgrading to Hibernate ORM 5.0.9

gunnarmorling:


Sanne:
Is there a specific reason to upgrade?

Just wondering if we need special care to keep WF10 compatibility..

testing

gunnarmorling:
Essentially I meant to go to the latest and greatest, but there is a test
failure. Need to investigate.

WF 10 has 5.0.7, so I was contemplating to align with that instead.

Sanne:
Failed tests: 
      ListTest.testOrderedList:62->OgmTestCase.checkCleanCache:104 [Entity cache should be empty] expected:<[0]> but was:<[5]>
      ListTest.testOrderedListAndCompositeId:172->OgmTestCase.checkCleanCache:104 [Entity cache should be empty] expected:<[0]> but was:<[2]>
      ListTest.testRemovalOfElementFromOrderedListIsApplied:131 expected:<['L[eia]']> but was:<['L[uke]']>
      ListTest.testUpdateToElementOfOrderedListIsApplied:94 expected:<['L[isa]', 'Leia']> but was:<['L[uke]', 'Leia']>

Same test failure?

gunnarmorling:
Yup.

gunnarmorling:
That error shows up as of 5.0.7. Investigating.

gunnarmorling:
Turns out that the update revealed one issue around updates to indexed element collections which used to be recreated prior to 5.0.7 and now not anymore: OGM-1034. Pushed one more commit for fixing that.

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/660

PR #661, opened 2016-05-04
Using slot "5.0" and alias to slot main

Sanne:
I haven't updated the documentation yet.

To verify that the aliases work, I've used a branch in which I revert all changes to the integration tests (as they were originally testing the "master" slot).

 - https://hibernate.atlassian.net/browse/OGM-1015

### DIFF
--- a/integrationtest/src/test/java/org/hibernate/ogm/test/integration/cassandra/ModulesMagicDeckIT.java
+++ b/integrationtest/src/test/java/org/hibernate/ogm/test/integration/cassandra/ModulesMagicDeckIT.java
@@ -44,7 +44,7 @@ public class ModulesMagicDeckIT extends MagiccardsDatabaseScenario {
 				.addClasses( MagicCard.class, MagicCardsCollectionBean.class, ModulesMagicDeckIT.class, MagiccardsDatabaseScenario.class );
 		String persistenceXml = persistenceXml().exportAsString();
 		webArchive.addAsResource( new StringAsset( persistenceXml ), "META-INF/persistence.xml" );
-		ModulesHelper.addModulesDependencyDeclaration( webArchive, "org.hibernate.ogm services, org.hibernate.ogm.cassandra services" );
+		ModulesHelper.addModulesDependencyDeclaration( webArchive, "org.hibernate.ogm:${hibernate-ogm.module.slot} services, org.hibernate.ogm.cassandra:${hibernate-ogm.module.slot} services" );
 		return webArchive;
 	}
 
--- a/integrationtest/src/test/java/org/hibernate/ogm/test/integration/cassandra/ModulesMagicDeckIT.java
+++ b/integrationtest/src/test/java/org/hibernate/ogm/test/integration/cassandra/ModulesMagicDeckIT.java
@@ -57,6 +57,7 @@ public class ModulesMagicDeckIT extends MagiccardsDatabaseScenario {
 				.provider( "org.hibernate.ogm.jpa.HibernateOgmPersistence" )
 				.clazz( MagicCard.class.getName() )
 				.getOrCreateProperties()
+					.createProperty().name( "jboss.as.jpa.providerModule" ).value( "application" ).up()
 					.createProperty().name( "hibernate.search.default.directory_provider" ).value( "ram" ).up()
 					.createProperty().name( "hibernate.ogm.datastore.database" ).value( "ogm_test_database" ).up()
 					.createProperty().name( "hibernate.ogm.datastore.provider" ).value( "cassandra_experimental" ).up()
--- a/integrationtest/src/test/java/org/hibernate/ogm/test/integration/infinispan/InfinispanModuleMemberRegistrationIT.java
+++ b/integrationtest/src/test/java/org/hibernate/ogm/test/integration/infinispan/InfinispanModuleMemberRegistrationIT.java
@@ -41,6 +41,7 @@ public class InfinispanModuleMemberRegistrationIT extends ModuleMemberRegistrati
 				.name( "primary" )
 				.provider( "org.hibernate.ogm.jpa.HibernateOgmPersistence" )
 				.clazz( Member.class.getName() ).getOrCreateProperties()
+				.createProperty().name( "jboss.as.jpa.providerModule" ).value( "application" ).up()
 				.createProperty().name( "hibernate.ogm.datastore.provider" ).value( "infinispan" ).up()
 				.createProperty().name( "hibernate.ogm.infinispan.configuration_resourcename" ).value( "infinispan.xml" ).up()
 				.createProperty().name( "hibernate.search.default.directory_provider" ).value( "ram" ).up()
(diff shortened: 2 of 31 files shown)


integrationtest/src/test/java/org/hibernate/ogm/test/integration/testcase/util/ModulesHelper.java
```diff
@@ -25,6 +26,27 @@
  */
 public class ModulesHelper {
 
+	private static String hibernateOgmVersion = null;
+	private static String hibernateOgmModuleSlot = null;
+
+	private static synchronized String getModuleSlotString() {
+		if ( hibernateOgmModuleSlot == null ) {
+			//This variable is computer from the project version, as the Maven build plugin helper
```
DavideD: s/computer/computed/  ?

integrationtest/src/test/java/org/hibernate/ogm/test/integration/cassandra/ModulesMagicDeckIT.java
```diff
@@ -44,7 +44,7 @@
 				.addClasses( MagicCard.class, MagicCardsCollectionBean.class, ModulesMagicDeckIT.class, MagiccardsDatabaseScenario.class );
 		String persistenceXml = persistenceXml().exportAsString();
 		webArchive.addAsResource( new StringAsset( persistenceXml ), "META-INF/persistence.xml" );
-		ModulesHelper.addModulesDependencyDeclaration( webArchive, "org.hibernate.ogm services, org.hibernate.ogm.cassandra services" );
+		ModulesHelper.addModulesDependencyDeclaration( webArchive, "org.hibernate.ogm:${hibernate-ogm.module.slot} services, org.hibernate.ogm.cassandra:${hibernate-ogm.module.slot} services" );
```
gunnarmorling: I suppose that's going to be replaced via Maven resource filtering? Let's give it a go, though my experiences with it in IDEs are mixed (every now and then, the unfiltered value would show up in the classes dir). But it might be more practicable nowadays, so +1.

modules/pom.xml
```diff
@@ -21,17 +21,17 @@
...
-        <hibernate.ogm.couchdb.module.slot>main</hibernate.ogm.couchdb.module.slot>
-        <hibernate.ogm.redis.module.slot>main</hibernate.ogm.redis.module.slot>
-        <org.parboiled.module.slot>main</org.parboiled.module.slot>
+        <hibernate.ogm.module.slot>${parsedVersion.majorVersion}.${parsedVersion.minorVersion}</hibernate.ogm.module.slot>
+        <hibernate.ogm.ehcache.module.slot>${hibernate.ogm.module.slot}</hibernate.ogm.ehcache.module.slot>
+        <hibernate.ogm.infinispan.module.slot>${hibernate.ogm.module.slot}</hibernate.ogm.infinispan.module.slot>
+        <hibernate.ogm.cassandra.module.slot>${hibernate.ogm.module.slot}</hibernate.ogm.cassandra.module.slot>
+        <hibernate.ogm.mongodb.module.slot>${hibernate.ogm.module.slot}</hibernate.ogm.mongodb.module.slot>
+        <hibernate.ogm.neo4j.module.slot>${hibernate.ogm.module.slot}</hibernate.ogm.neo4j.module.slot>
+        <hibernate.ogm.neo4j-internal.module.slot>${project.version}</hibernate.ogm.neo4j-internal.module.slot>
```
DavideD: Is it correct `${project.version}` here? Shouldn't it be `${hibernate.ogm.module.slot}`?

Sanne: It's intentional: for "private" modules we prefer to use the full-format project version, so to be able to change them in micro releases too.

integrationtest/src/test/java/org/hibernate/ogm/test/integration/infinispan/InfinispanModuleMemberRegistrationIT.java
```diff
@@ -41,6 +41,7 @@ private static PersistenceDescriptor persistenceXml() {
 				.name( "primary" )
 				.provider( "org.hibernate.ogm.jpa.HibernateOgmPersistence" )
 				.clazz( Member.class.getName() ).getOrCreateProperties()
+				.createProperty().name( "jboss.as.jpa.providerModule" ).value( "application" ).up()
```
gunnarmorling: What's that one needed for? Specifically I can't seem to find any docs on the "application" value. I suppose your docs will clarify it :)

integrationtest/src/test/java/org/hibernate/ogm/test/integration/testcase/util/ModulesHelper.java
```diff
@@ -37,6 +59,12 @@ private static Asset manifest(String dependencies) {
 	}
 
 	private static String injectVariables(String dependencies) {
+		String variablesFromProperties = injectVariablesFromProperties( dependencies );
+		//The OGM module slot is "hardcoded" as a special case:
+		return applyPropertyReplacement( variablesFromProperties, "hibernate-ogm.module.slot", getModuleSlotString() );
```
gunnarmorling: Ah, I see. That should clarify my first question about resource filtering.

integrationtest/src/test/java/org/hibernate/ogm/test/integration/cassandra/ModulesMagicDeckIT.java
Sanne: Indeed, I did all those changes to make it work in the IDE.

The maven build-helper plugin doesn't work in the IDE, so while we use that to define the major.minor properties, the Integration tests re-compute the values from a different source (and inject these) using explicit Java code, so to keep the ability to run integration tests easily from the IDE.

pom.xml
```diff
@@ -258,6 +258,21 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <!-- This defines the properties 'parsedVersion.majorVersion'
+                    and 'parsedVersion.minorVersion' to be in synch with the current project version -->
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
+                <version>1.9.1</version>
```
gunnarmorling: Could you put the version into `pluginManagement`? That's how we did it for all other plug-ins.

Sanne: Good point, I copied this from Search so there's a defect there too :)

integrationtest/src/test/java/org/hibernate/ogm/test/integration/testcase/util/ModulesHelper.java
Sanne: +1 thanks

integrationtest/src/test/java/org/hibernate/ogm/test/integration/infinispan/InfinispanModuleMemberRegistrationIT.java
Sanne: Technically it _would_ still work as we provided the "main" alias. But I want to make sure that the tests are properly testing our slotted modules.

The property is needed to prevent Jipijapa from injecting a dependency to "org.hibernate.ogm:main" .

The value "application" means that the user is bundling its own provider and Jipijapa should not add any:
https://docs.jboss.org/author/display/WFLY10/JPA+Reference+Guide

Sanne:
Thanks for the first round. Pushed an update:
 - typo fixed
 - rebased on latest
 - moved the plugin version to the right section

Back to docs now, but I suppose this could be merged already.

Sanne:
Added the documentation changes.

DavideD:
Rebased and Merged https://github.com/hibernate/hibernate-ogm/pull/661

PR #666, opened 2016-05-06
OGM-1027 Support BSON function calls in the MongoDB native parser

gsmet:


DavideD:
Rebsed and Merged https://github.com/hibernate/hibernate-ogm/pull/666
Thanks!

PR #669, opened 2016-05-10
OGM-1046 Commit transaction after tests

DavideD:
https://hibernate.atlassian.net/browse/OGM-1046

This patch doesn't speed up things as I initially thought, it seems the tests take a lot  of time when the EntityManagerFactory is initialized.

That said, the transaction wasn't committed in the end, and with Neo4j remote it takes longer to execute the build without this fix.

### DIFF
--- a/core/src/test/java/org/hibernate/ogm/backendtck/batchfetching/MultiGetEmbeddedIdTest.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/batchfetching/MultiGetEmbeddedIdTest.java
@@ -52,16 +52,16 @@ public class MultiGetEmbeddedIdTest extends OgmTestCase {
 	private static final TupleContext TUPLECONTEXT = new TupleContextImpl( Arrays.asList( "id.name", "id.publisher" ), EMPTY_ASSOCIATION_METADATA, EMPTY_ROLES,
 			EmptyOptionsContext.INSTANCE );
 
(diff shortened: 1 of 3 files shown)


Sanne:
'transactino' is a small italian cocktail ;)

core/src/test/java/org/hibernate/ogm/backendtck/batchfetching/MultiGetEmbeddedIdTest.java
```diff
@@ -52,32 +52,38 @@
...
-			assertThat( tuples.get( 2 ).get( "id.name" ) ).isEqualTo( KING_OF_TOKYO.getId().getName() );
+				assertThat( tuples.get( 2 ).get( "id.publisher" ) ).isEqualTo( KING_OF_TOKYO.getId().getPublisher() );
+				assertThat( tuples.get( 2 ).get( "id.name" ) ).isEqualTo( KING_OF_TOKYO.getId().getName() );
 
-			session.getTransaction().commit();
+				session.getTransaction().commit();
+			}
+			catch (Exception e) {
+				rollback( tx );
+				throw e;
```
Sanne: not critical as it's just a test, but keep in mind that:
 - you could have an exception thrown during commit()
 - it's illegal to try to rollback a transaction which was committed (even if it failed to commit)

Sanne:
merged, fixed typo in the commit message

PR #671, opened 2016-05-11
OGM-854 Neo4j - Properly initialize ToOne relation for JP-QL and native queries

gsmet:


emmanuelbernard:
hey @gsmet 

Can you explain in layman terms what the indexedTypeName vs EntityMetadata does and why it fixes the issue?

Also I see that you create metadata objects during the runtime path (i.e. not cached) `new DefaultEntityKeyMetadata`. Can this be avoided and cached per query plan somehow? Maybe I'm not reading properly and there is no query plan per se.

gsmet:
@emmanuelbernard the core of the problem is that a Neo4jTupleSnapshot was created without the association information which are required to build the complete tuple: in MongoDB's case, the data is in the author_id column but in Neo4j's case, it's in a outer relation which Neo4jTupleSnapshot can get if we give it the proper information.

So the patch does just permit to get a proper TupleContext to pass to the Neo4jTupleSnapshot and we need the type for that. We cannot use a Class<?> as in the custom case, we only have a string.

As for building the DefaultEntityKeyMetadata, I will move it back where it was. I think you're right it's better to keep it the way it was.

gsmet:
@emmanuelbernard pushed an update to fix it

gsmet:
@emmanuelbernard I gave it some further thoughts and I'm wondering if we should integrate the entity type name into EntityKeyMetadata. I think this patch is proof that we need this information in this context so we might as well integrate it.

Anything against it? I prefer to check before as there are a lot of tests to adapt.

core/src/main/java/org/hibernate/ogm/query/impl/OgmQueryTranslator.java
```diff
@@ -80,9 +80,17 @@
 	 */
 	private SelectClause selectClause;
 
+	/**
+	 * When the query is only targeting one type, we register the relevant information here.
+	 */
 	private EntityKeyMetadata singleEntityKeyMetadata;
 
 	/**
+	 * When the query is only targeting one type, we register the relevangt information here.
```
DavideD: s/relevangt/relevant/

Not considering the syntax error could you add some details about what you mean with "relevant"? In particular, I don't think I understand the difference between, singleIndexedTypeName and singleEntityKeyMetadata. (Why the word "indexed"?)

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/query/nativequery/Neo4jSessionNativeQueryTest.java
```diff
@@ -29,9 +29,9 @@
  */
 public class Neo4jSessionNativeQueryTest extends OgmTestCase {
 
-	private final OscarWildePoem portia = new OscarWildePoem( 1L, "Portia", "Oscar Wilde", new GregorianCalendar( 1808, 3, 10, 12, 45 ).getTime() );
-	private final OscarWildePoem athanasia = new OscarWildePoem( 2L	, "Athanasia", "Oscar Wilde", new GregorianCalendar( 1810, 3, 10 ).getTime() );
-	private final OscarWildePoem ballade = new OscarWildePoem( 3L, "Ballade De Marguerite", "Oscar Wilde", new GregorianCalendar( 1881, 3, 1 ).getTime() );
+	private final OscarWildePoem portia = new OscarWildePoem( 1L, "Portia", "Oscar Wilde", new GregorianCalendar( 1808, 3, 10, 12, 45 ).getTime(), null );
+	private final OscarWildePoem athanasia = new OscarWildePoem( 2L	, "Athanasia", "Oscar Wilde", new GregorianCalendar( 1810, 3, 10 ).getTime(), null );
+	private final OscarWildePoem ballade = new OscarWildePoem( 3L, "Ballade De Marguerite", "Oscar Wilde", new GregorianCalendar( 1881, 3, 1 ).getTime(), null );
```
DavideD: No need to change, but why not to have a separate constructor for this?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/Neo4jTupleSnapshot.java
```diff
@@ -100,6 +98,7 @@ private Object readEmbeddedProperty(String column) {
 
 	private Object readProperty(Node otherNode, String targetColumnName) {
 		Object value = null;
+		otherNode.getAllProperties();
```
DavideD: The returned value is not used.

gsmet: Yeah, sorry, a leftover... Fixed it.

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/query/nativequery/Neo4jSessionNativeQueryTest.java
gsmet: You're right, I fixed it.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/Neo4jDialect.java
```diff
@@ -634,8 +639,9 @@ public void forEachTuple(ModelConsumer consumer, EntityKeyMetadata... entityKeyM
 		String nativeQuery = buildNativeQuery( backendQuery, queryParameters );
 		Result result = dataBase.execute( nativeQuery, parameters );
 
-		if ( backendQuery.getSingleEntityKeyMetadataOrNull() != null ) {
-			return new NodesTupleIterator( result, backendQuery.getSingleEntityKeyMetadataOrNull() );
+		if ( backendQuery.isSingleEntity() ) {
+			OgmEntityPersister persister = (OgmEntityPersister) sessionFactory.getEntityPersister( backendQuery.getSingleIndexedTypeNameOrNull() );
+			return new NodesTupleIterator( result, backendQuery.getSingleEntityKeyMetadataOrNull(), persister.getTupleContext() );
```
DavideD: Wouldn't make sense to add a parameter to `executeBackendQuery` as well? We will need it anyway in the future to get the id of the running transaction. Something like `QueryContext` (equivalent to `TupleContext`) the can return the additional information that we need.

core/src/main/java/org/hibernate/ogm/query/impl/OgmQueryTranslator.java
gsmet: Changed it to singleEntityTypeName.

There is not really a difference between the 2. It's just that the EntityKeyMetadata doesn't contain the type name and we require it so we have to have it there.

I considered adding it to EntityKeyMetadata following Emmanuel's comments but I'm not sure it's a good idea as it's part of the cache key.

gsmet:
FYI, I added a commit to clean up the contract of forEachTuple. As we require only one entity type, we might as well enforce it. All the callers were using only one entity type anyway.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/Neo4jDialect.java
gsmet: I have nothing against it but I'm wondering if it should be part of this patch?

gsmet:
@emmanuelbernard after thinking about it, I don't think it's a good idea to add the type name to the EntityKeyMetadata.

infinispan/src/test/java/org/hibernate/ogm/datastore/infinispan/test/dialect/impl/InfinispanDialectWithClusteredConfigurationTest.java
```diff
@@ -162,7 +162,7 @@ public void shouldApplyForEachTupleInClusteredMode() throws Exception {
 
 		// then
 		MyConsumer consumer = new MyConsumer();
-		dialect2.forEachTuple( consumer, keyMetadata );
+		dialect2.forEachTuple( consumer, null, keyMetadata );
```
DavideD: To be consistent with the rest of the test, this should  be `emptyTupleContext()`

core/src/main/java/org/hibernate/ogm/dialect/spi/GridDialect.java
```diff
@@ -172,10 +172,10 @@
 	 *            the instance that is going to be called for every {@link Tuple}
 	 * @param tupleContext
 	 *            contains additional information that might be used to build the tuple
-	 * @param entityKeyMetadatas
-	 *            the key metadata of the tables for which we want to apply the consumer
+	 * @param entityKeyMetadata
+	 *            the key metadata of the table for which we want to apply the consumer
 	 */
-	void forEachTuple(ModelConsumer consumer, TupleContext tupleContext, EntityKeyMetadata... entityKeyMetadatas);
+	void forEachTuple(ModelConsumer consumer, TupleContext tupleContext, EntityKeyMetadata entityKeyMetadata);
```
DavideD: @gunnarmorling Do you see any issues in integrating this for the release?

This method has been introduced for the MassIndexer but it seems that even in that case we only index one entity at a time. Also, is it OK to change the interface before the Candidate release?

I'm OK integrating the current patch

infinispan/src/test/java/org/hibernate/ogm/datastore/infinispan/test/dialect/impl/InfinispanDialectWithClusteredConfigurationTest.java
gsmet: Right, fixed!

core/src/main/java/org/hibernate/ogm/dialect/spi/GridDialect.java
gsmet: Note that as we now require the TupleContext, we cannot support different EntityKeyMetadata.

core/src/main/java/org/hibernate/ogm/dialect/spi/GridDialect.java
```diff
@@ -165,14 +165,17 @@
...
 	 *            the instance that is going to be called for every {@link Tuple}
-	 * @param entityKeyMetadatas
-	 *            the key metadata of the tables for which we want to apply the costumer
+	 * @param tupleContext
+	 *            contains additional information that might be used to build the tuple
+	 * @param entityKeyMetadata
+	 *            the key metadata of the table for which we want to apply the consumer
 	 */
-	void forEachTuple(ModelConsumer consumer, EntityKeyMetadata... entityKeyMetadatas);
+	void forEachTuple(ModelConsumer consumer, TupleContext tupleContext, EntityKeyMetadata entityKeyMetadata);
```
gunnarmorling: That last minute SPI change is a bit troublesome. Normally I'd say that'd grant another CR :/ I suppose there is no way around it?

core/src/main/java/org/hibernate/ogm/loader/nativeloader/impl/BackendCustomQuery.java
```diff
@@ -41,7 +41,8 @@
 	private final Set<String> querySpaces;
 	private final List<Return> customQueryReturns;
 
-	private final EntityKeyMetadata singleEntityKeyMetadata;
+	private EntityKeyMetadata singleEntityKeyMetadata;
```
gunnarmorling: Can we make that final again? I'm rather keen on making things truely immutable which grants safe publication across threads.

You could create a container type to be returned by the `determine...` method if it is about the need for multiple return values (which I suppose).

core/src/main/java/org/hibernate/ogm/dialect/spi/GridDialect.java
DavideD: I don't think there is, should we close an eye for this time?
The alternative is to include this in another release.

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/query/nativequery/Neo4jEntityManagerNativeQueryTest.java
```diff
@@ -133,6 +133,19 @@ public void testQueryWithCompositeId() throws Exception {
...
 
+	@Test
+	@TestForIssue(jiraKey = "OGM-854")
+	public void testToOneInitializedByQuery() throws Exception {
+		em.getTransaction().begin();
+
+		String nativeQuery = "MATCH ( n:" + TABLE_NAME + " { name:'Portia', author:'Oscar Wilde' } ) RETURN n";
+		OscarWildePoem poem = (OscarWildePoem) em.createNativeQuery( nativeQuery, OscarWildePoem.class ).getSingleResult();
+
+		assertThat( poem.getCritic() ).isNotNull();
```
gunnarmorling: Could you add an assertion on one of the associated props? Just to be sure...

gunnarmorling:
Ok, looks good to me overall. I think changing the SPI is not ideal but acceptable. Just let's add an entry to the migration notes.

@gsmet: Added some more comments inline, good to go after that.

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/query/nativequery/Neo4jEntityManagerNativeQueryTest.java
gsmet: Done!

core/src/main/java/org/hibernate/ogm/query/impl/OgmQueryTranslator.java
```diff
@@ -81,14 +81,9 @@
...
-	 * When the query is only targeting one type, we register the EntityKeyMetadata for this type.
+	 * When the query is only targeting one type, we register the EntityMetadataInformation for this type.
 	 */
-	private EntityKeyMetadata singleEntityKeyMetadata;
-
-	/**
-	 * When the query is only targeting one type, we register the name of the entity type.
-	 */
-	private String singleEntityTypeName;
+	private EntityMetadataInformation singleEntityMetadataInformation;
```
DavideD: This can be final, isn't it?

gsmet: No, it's initialized in doCompile().

core/src/main/java/org/hibernate/ogm/loader/nativeloader/impl/BackendCustomQuery.java
```diff
@@ -60,24 +60,27 @@ public BackendCustomQuery(String queryString, T query, NativeSQLQueryReturn[] qu
...
 			if ( queryReturn instanceof RootReturn ) {
-				if ( metadata != null ) {
+				if ( metadataInformation != null ) {
 					return null;
 				}
 				RootReturn rootReturn = (RootReturn) queryReturn;
 				OgmEntityPersister persister = (OgmEntityPersister) sessionFactory.getEntityPersister( rootReturn.getEntityName() );
-				metadata = new DefaultEntityKeyMetadata( persister.getTableName(), persister.getRootTableIdentifierColumnNames() );
+				metadataInformation = new EntityMetadataInformation(
+						new DefaultEntityKeyMetadata( persister.getTableName(), persister.getRootTableIdentifierColumnNames() ),
```
gunnarmorling: Couldn't the EKM from the persister be used?

core/src/main/java/org/hibernate/ogm/massindex/impl/BatchIndexingWorkspace.java
```diff
@@ -66,18 +66,15 @@ public BatchIndexingWorkspace(GridDialect gridDialect, SearchIntegrator search,
...
-		return new DefaultEntityKeyMetadata( persister.getTableName(), persister.getRootTableIdentifierColumnNames() );
-	}
-
 	@Override
 	public void run() {
 		ErrorHandler errorHandler = searchIntegrator.getErrorHandler();
 		try {
-			final EntityKeyMetadata keyMetadata = getEntityKeyMetadata();
+			OgmEntityPersister persister = (OgmEntityPersister) sessionFactory.getEntityPersister( indexedType.getName() );
+			final EntityKeyMetadata keyMetadata = new DefaultEntityKeyMetadata( persister.getTableName(), persister.getRootTableIdentifierColumnNames() );
```
gunnarmorling: I'm wondering why we don't re-use the EKM from the persister. I realize it's been like that before your change, but it seems not needed.

core/src/main/java/org/hibernate/ogm/loader/nativeloader/impl/BackendCustomQuery.java
gsmet: Yes, haven't looked after it as it was already instantiating a new one in the existing code.

gunnarmorling:
Merged; Thanks, @gsmet! Added one more commit to mark `BackendQuery` as experimental.

gunnarmorling:
Could you add a note on the changed method in `GridDialect` to the [migration notes](https://developer.jboss.org/wiki/HibernateOGMMigrationNotes)? Thanks!

PR #673, opened 2016-05-13
[5.1] OGM-982 - Upgrade to ORM 5.1

gunnarmorling:
The actual update wasn't too hard, that's done in the first commit. The more challenging issue was getting the integration tests to work. The reason being that WF 10 - naturally - still contains ORM 5.0. So that's what I did to make it work:

* Create a new module for ORM 5.1; Note that it doesn't contain the actual JARs but refers to them via Maven artifact coordinates; That requires a Maven repo on the system running WF, but I find that a good compromise for the sake of not futher increasing the size of the module ZIP
* Create a new module for HSEARCH 5.5.3-orm51; This is to make HSEARCH use ORM 5.1 instead of 5.0, as otherwise OGM and HSEARCH would work with conflicting ORM versions. This requires the 5.5.3 HSEARCH module ZIP to have been unzipped.

Let me know what you think. If we agree on the general approach, I'll update the docs.

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
+++ b/core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
@@ -23,7 +23,7 @@ import org.hibernate.LockMode;
 import org.hibernate.LockOptions;
 import org.hibernate.MappingException;
 import org.hibernate.StaleObjectStateException;
-import org.hibernate.bytecode.instrumentation.spi.LazyPropertyInitializer;
+import org.hibernate.bytecode.enhance.spi.LazyPropertyInitializer;
 import org.hibernate.cache.spi.access.EntityRegionAccessStrategy;
 import org.hibernate.cache.spi.access.NaturalIdRegionAccessStrategy;
 import org.hibernate.cache.spi.entry.CacheEntry;
@@ -607,6 +607,10 @@ public abstract class OgmEntityPersister extends AbstractEntityPersister impleme
 	public Object initializeLazyProperty(String fieldName, Object entity, SessionImplementor session)
 			throws HibernateException {
 
+		if ( true ) {
+			throw new RuntimeException( "BOOM" );
+		}
+
 		final Serializable id = session.getContextEntityIdentifier( entity );
 
 		final EntityEntry entry = session.getPersistenceContext().getEntry( entity );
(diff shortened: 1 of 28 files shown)


Sanne:
> Note that it doesn't contain the actual JARs but refers to them via Maven artifact coordinates; That requires a Maven repo on the system running WF

I'm not convinced by this choice, you're trading away usability for the sake of some notion that 60MB is ok while 70MB is not acceptable.

If you feel strongly about making the modules smaller, we could have a second set which uses this approach consistently (for all jars), but we need at least the full-weight modules for those running offline, copying Maven repos with "the right content" is a hurdle, not least the documentation becomes even more intimidating when you explain all these details.

gunnarmorling:
Indeed I see it as a first step to obtain all JARs via their artifacts. Then we are talking about dozens of MB vs. just a few module.xml.

I'm negative on maintaining two sets of modules. So if we cannot agree on the slimmed-down approach I'll add the new JARs here, too. But this thing is getting larger and larger. Would be interesting what others think.

> copying Maven repos with "the right content" is a hurdle

https://github.com/release-engineering/offliner might help with that for those who really need it, but that'd need some more investigation/documentation.

Sanne:
>  Would be interesting what others think.

+1, and it came up before

core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
```diff
@@ -607,6 +607,10 @@ private Tuple getResultsetById(Serializable id, SessionImplementor session) {
 	public Object initializeLazyProperty(String fieldName, Object entity, SessionImplementor session)
 			throws HibernateException {
 
+		if ( true ) {
+			throw new RuntimeException( "BOOM" );
+		}
```
DavideD: ???

DavideD:
For this specific issue I would stay consistent and copy the jar in the dist package.

Let's open a separate JIRA and talk about alternatives. I think I like the idea of two different module set but I'm worried about the increase in complexity related to integration tests and maintenance.

core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
gunnarmorling: Oops :)

gsmet:
Same here, we should keep it consistent with how it works now and open another issue if we want to look for alternatives.

gsmet:
@gunnarmorling do you plan to make progress on this or should I take it over?

Thanks!

gunnarmorling:
@gsmet, yes, I'll take it up.

gunnarmorling:
Closing it; Going to send a new one.

PR #674, opened 2016-05-13
OGM-1053 Minor various documentation updates for 5.0 release

Sanne:
Here it is.

I have been rephrasing some sentences as well, for improved clarity.. I think they are better, often because we have a better idea of what it does or how it's meant than when those sentences were originally written.

BUT

I guess some of these changes might be questionable. I'd suggest that the reviewer accepts the changes he likes and ignores the ones he doesn't, so to not delay the release with nit-picking on language.


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/674

Thanks!

PR #676, opened 2016-05-17
OGM-1038 Pass transaction id to dialects

DavideD:
https://hibernate.atlassian.net/browse/OGM-1038

This will make it possible for a dialect to access information about the current transaction. At the moment is only the transaction id.

I haven't added any tests for this specific part since none of the current dialects need this information.
It will be used by the Neo4j remote dialect in following PR.

Note that I added a `QueryContext` to the method executing native queries. It will also store the `TupleContext` removing the need to save the sessionFactory in the Neo4j dialect.

We already discussed about this PR in the pasts but we changed the codebase a bit so, maybe, you won't to have a quick look.

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/dialect/impl/EmptyTransactionContext.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/impl/EmptyTransactionContext.java
@@ -1,0 +1,32 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.dialect.impl;
+
+import org.hibernate.ogm.dialect.spi.TransactionContext;
+import org.hibernate.ogm.util.impl.Log;
+import org.hibernate.ogm.util.impl.LoggerFactory;
+
+/**
+ * A {@link TransactionContext} that can be used when the dialect does not need to know the transaction id.
+ *
+ * @author Davide D'Alto
+ */
+public class EmptyTransactionContext implements TransactionContext {
+
+	public static TransactionContext INSTANCE = new EmptyTransactionContext();
+
+	private static final Log LOG = LoggerFactory.make();
+
+	private EmptyTransactionContext() {
+	}
+
+	@Override
+	public Object getTransactionId() {
+		throw LOG.transactionIdIsNotAvailable();
+	}
+
+}
--- a/core/src/main/java/org/hibernate/ogm/dialect/spi/TransactionContext.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/spi/TransactionContext.java
@@ -1,0 +1,22 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.dialect.spi;
+
+/**
+ * Contains information about the running transaction.
+ *
+ * @author Davide D'Alto
+ */
+public interface TransactionContext {
+
+	/**
+	 * A value that can be used to identify the running transaction.
+	 *
+	 * @return the transaction identifier
+	 */
+	Object getTransactionId();
+}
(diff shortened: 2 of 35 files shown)


core/src/main/java/org/hibernate/ogm/dialect/spi/TransactionContext.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.dialect.spi;
+
+/**
+ * @author Davide D'Alto
```
gsmet: Maybe a small javadoc would be nice?

core/src/main/java/org/hibernate/ogm/dialect/impl/EmptyTransactionContext.java
```diff
@@ -0,0 +1,32 @@
...
+	public static TransactionContext INSTANCE = new EmptyTransactionContext();
+
+	private static final Log LOG = LoggerFactory.make();
+
+	private EmptyTransactionContext() {
+	}
+
+	@Override
+	public Object getTransactionId() {
+		throw LOG.transactionIdIsNull();
```
gsmet: I would have named it `transactionIdIsNotAvailable` as the message of the exception states it.

gsmet:
Looks good to me overall apart from the minor nitpicking.

That being said, not sure I understand all the implications of this so I would prefer to have a second opinion before merging.

DavideD:
I've applied @gsmet remarks

core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
```diff
@@ -1743,8 +1746,17 @@ public String getSubclassForDiscriminatorValue(Object value) {
...
 
-	public TupleContext getTupleContext() {
-		return tupleContext;
+	/**
+	 * Returns the {@link TupleContext}.
+	 *
+	 * @param session the current session, if null the {@link TupleContext#getTransactionContext()} will be null.
+	 * @return the tupleContext for the session
+	 */
+	public TupleContext getTupleContext( SessionImplementor session ) {
```
gunnarmorling: When will the session be null?

core/src/main/java/org/hibernate/ogm/util/impl/TupleContextHelper.java
```diff
@@ -0,0 +1,34 @@
...
+	/**
+	 * Given a {@link SessionImplementor} returns the {@link TupleContext} associated to an entity.
+	 *
+	 * @param session the current session
+	 * @param metadata the {@link EntityMetadataInformation} of the entity associated to the TupleContext
+	 * @return the TupleContext associated to the current session for the entity specified, or {@code null} if the
+	 * EntityMetadataInformation is {@code null}
+	 */
+	public static TupleContext tupleContext(SessionImplementor session, EntityMetadataInformation metadata) {
+		if ( metadata != null ) {
```
gunnarmorling: When would that be the case? Seems like null should be prevented in the first place?

DavideD: Check `BackendQuery#getSingleEntityMetadataInformationOrNull()`:
```
	/**
	 * Returns the {@link EntityMetadataInformation} of the entity type selected by this query.
	 *
	 * @return the {@link EntityMetadataInformation} of the entity type selected by this query or {@code null} in case this
	 * query does not select exactly one entity type (e.g. in case of scalar values or joins (if supported in future revisions)).
	 */
	public EntityMetadataInformation getSingleEntityMetadataInformationOrNull() {
		return singleEntityMetadataInformation;
	}

```

core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
DavideD: Good question ... checking ...

DavideD: It happens when massindexing, in that case there is not a session when the foreach method is called.
At the moment, Neo4j does not work with the massindexer because of different Lucene versions. 
It should be possible to implement it in the remote scenario but I haven't done that yet

gsmet:
As discussed on Thursday, you need to check if QueryContext is required as the TransactionContext in it seems redundant with the TransactionContext located in the TupleContext.

DavideD:
I removed the `QueryContext` interface and rebased.

Having a TupleContext in executeBackendQuery also allowed me to remove the SessionFactory from the Neo4j dialect. See [OGM-963](https://hibernate.atlassian.net/browse/OGM-963)

gsmet:
Merged! Thanks!

PR #677, opened 2016-05-17
OGM-1054 Additional type descriptors

DavideD:
https://hibernate.atlassian.net/browse/OGM-1054

New type descriptors for Neo4j remote, I've added them in core as they might come handy in the future for other dialects.

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/type/impl/ByteArrayStringType.java
+++ b/core/src/main/java/org/hibernate/ogm/type/impl/ByteArrayStringType.java
@@ -1,0 +1,38 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.type.impl;
+
+import org.hibernate.MappingException;
+import org.hibernate.engine.spi.Mapping;
+import org.hibernate.ogm.type.descriptor.impl.StringMappedGridTypeDescriptor;
+import org.hibernate.type.descriptor.java.ByteArrayTypeDescriptor;
+
+/**
+ * Persist an array of {@link Byte} as {@link String}.
+ *
+ * @see StringMappedGridTypeDescriptor
+ * @author Davide D'Alto
+ */
+public class ByteArrayStringType extends AbstractGenericBasicType<Byte[]> {
+
+	public static final ByteArrayStringType INSTANCE = new ByteArrayStringType();
+
+	public ByteArrayStringType() {
+		super( StringMappedGridTypeDescriptor.INSTANCE, ByteArrayTypeDescriptor.INSTANCE );
+	}
+
+	@Override
+	public String getName() {
+		return "byte_array_as_string";
+	}
+
+	@Override
+	public int getColumnSpan(Mapping mapping) throws MappingException {
+		return 1;
+	}
+
+}
--- a/core/src/main/java/org/hibernate/ogm/type/impl/ByteMappedType.java
+++ b/core/src/main/java/org/hibernate/ogm/type/impl/ByteMappedType.java
@@ -1,0 +1,40 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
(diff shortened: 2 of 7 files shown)


core/src/main/java/org/hibernate/ogm/type/impl/ByteArrayStringType.java
```diff
@@ -0,0 +1,35 @@
...
+
+	public static final ByteArrayStringType INSTANCE = new ByteArrayStringType();
+
+	public ByteArrayStringType() {
+		super( StringMappedGridTypeDescriptor.INSTANCE, ByteArrayTypeDescriptor.INSTANCE );
+	}
+
+	@Override
+	public String getName() {
+		return "Byte_array";
```
gsmet: Is the uppercase "Byte" expected? If so, I'm wondering if the other types should be consistent with this naming scheme. See ByteMappedType for instance.

Shouldn't we add the _as_string suffix to be consistent with PrimitiveByteArrayStringType?

core/src/main/java/org/hibernate/ogm/type/impl/BinaryMappedType.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.hibernate.MappingException;
+import org.hibernate.engine.spi.Mapping;
+import org.hibernate.ogm.type.descriptor.impl.ByteArrayMappedGridTypeDescriptor;
+import org.hibernate.type.descriptor.java.PrimitiveByteArrayTypeDescriptor;
+
+/**
+ * @author Davide D'Alto
+ */
+public class BinaryMappedType extends AbstractGenericBasicType<byte[]> {
```
gsmet: Is there a reason why it's called Binary? I would have expected PrimaryByteArrayMappedType?

core/src/main/java/org/hibernate/ogm/type/impl/BinaryMappedType.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.hibernate.ogm.type.impl;
+
+import org.hibernate.MappingException;
+import org.hibernate.engine.spi.Mapping;
+import org.hibernate.ogm.type.descriptor.impl.ByteArrayMappedGridTypeDescriptor;
+import org.hibernate.type.descriptor.java.PrimitiveByteArrayTypeDescriptor;
+
+/**
+ * @author Davide D'Alto
```
gsmet: Javadoc.

core/src/main/java/org/hibernate/ogm/type/impl/ByteArrayStringType.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.hibernate.ogm.type.impl;
+
+import org.hibernate.MappingException;
+import org.hibernate.engine.spi.Mapping;
+import org.hibernate.ogm.type.descriptor.impl.StringMappedGridTypeDescriptor;
+import org.hibernate.type.descriptor.java.ByteArrayTypeDescriptor;
+
+/**
+ * @author Davide D'Alto
```
gsmet: Javadoc.

core/src/main/java/org/hibernate/ogm/type/impl/ByteMappedType.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.hibernate.ogm.type.impl;
+
+import org.hibernate.MappingException;
+import org.hibernate.engine.spi.Mapping;
+import org.hibernate.ogm.type.descriptor.impl.ByteMappedGridTypeDescriptor;
+import org.hibernate.type.descriptor.java.ByteTypeDescriptor;
+
+/**
+ * @author Davide D'Alto
```
gsmet: Javadoc.

core/src/main/java/org/hibernate/ogm/type/impl/ByteArrayStringType.java
DavideD: We don't really have a convention for these names, actually this property is not even  used. I'll update anyway to keep a bit of consistency

core/src/main/java/org/hibernate/ogm/type/impl/BinaryMappedType.java
DavideD: None really, just a name I've seen used while creating the classes. I will rename it

DavideD:
Added some javadoc

gsmet:
Merged! Thanks!

PR #680, opened 2016-05-19
Set global UTC timezone for testing

schernolyas:
Set default timezone for testing globally. It needs for run  tests in other timezones. 

I am located in Moscow (UTC+3) .I got incorrect work of test org.hibernate.ogm.backendtck.type.BuiltInTypeTest

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -415,6 +415,7 @@
                         <systemPropertyVariables>
                            <!-- See test org.hibernate.ogm.test.utils.EnvironmentTest -->
                            <hibernate.service.allow_crawling>false</hibernate.service.allow_crawling>
+                           <user.timezone>UTC</user.timezone>
                         </systemPropertyVariables>
                     </configuration>
                 </plugin>


gunnarmorling:
@schernolyas, what's the exact error you get? That test already sets the TZ in via `@BeforeClass setDefaultTimeZone()`, so I'm wondering why you set it globally again?

Btw. `BuiltinTypeTest` also passes for me if I set the TZ to "Europe/Moscow" in `setDefaultTimeZone`. Would be great to know why it's failing for you.

schernolyas:
Hi @gunnarmorling !

I got broken test then I develop new module for Orientdb. The error occurs because OrientDB  initialized timezone before method @BeforeClass runned.  After it, I and @DavideD decided to  set timezone as JVM parameter. It is more reliable

gunnarmorling:
Ah, so are you saying that the test is failing only with the OrientDB backend you are working on?

> The error occurs because OrientDB initialized timezone before method @BeforeClass runned

But then the TZ should be UTC as set by the `@BeforeClass` method, no?

> set timezone as JVM parameter. It is more reliable

The rub is that it doesn't work when running the test from within the IDE. Hence I'd prefer a solution *within* the test class itself. But that's what we have already, so I'm still not quite sure why the proposed change is needed. Can I look at the backend you work on somewhere on GitHub by any chance?

Sanne:
I guess that some static initializers are triggered in OrientDB, so
bypassing the @beforeClass step.

We could also introduce such a static initializer, and with some care make
sure that's triggered first.

gunnarmorling:
Just talking to Sergey on HC about this :)

I'd rather see that the TZ used with the DB is set to a specific value
(UTC) when establishing the connection for tests.

schernolyas:
The function implemented by new method org.hibernate.ogm.util.TestHelper.prepareDatabase.

PR closed.

PR #681, opened 2016-05-19
OGM-356 Support object comparison in JPQL queries for MongoDB and Neo4j

gsmet:
https://hibernate.atlassian.net/browse/OGM-356

OK, so here it is... Took me a while to get it right with Neo4j.

## Oh man, it looks easy, it's just a NPE

The first part of the patch is designed to fix the issue people have mentioned in the JIRA: they get a NPE because we use `nullSafeSet` without a session and a session is required for `ManyToOneType`'s `nullSafeSet`.

I took the liberty to introduce a new method `convertToBackendType` which role is to convert the value to the backend type using a `SessionFactory` instead of being based on the session. I implemented it in the other types by calling the `nullSafeSet` method so no changes to expect except in the `ManyToOne` case.

I stayed on the safe side in `ManyToOneType` and I didn't change the `nullSafeSet` implementation based on the session.

This work alone makes it work with MongoDB. As long as we only use it for a single ToOne. Supporting more complex cases with MongoDB is left for a future patch.

## Neo4j in da place

Neo4j is a different beast as even *ToOne association are managed with relationships and as for now, only the embedded case was supported.

So I did the work to support associations, basing my work on what already existed for embedded. The main change is that we build the alias tree one step at a time to be sure we get all the required associations. This way, we even support the case for nested associations or nested embedded. See the test case for all the things that are supported now.


DavideD:
> This way, we even support the case for nested associations or nested embedded. See the test case for all the things that are supported now.

What do you mean with nested associations? Embedded association containing another association?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/query/parsing/impl/Neo4jAliasResolver.java
```diff
@@ -0,0 +1,179 @@
...
+import org.hibernate.ogm.datastore.neo4j.dialect.impl.NodeLabel;
+import org.hibernate.ogm.util.impl.StringHelper;
+
+/**
+ * Keep track of the aliases needed to create the Cypher query.
+ *
+ * @author Davide D'Alto
+ * @author Guillaume Smet
+ */
+public class Neo4jAliasResolver {
```
DavideD: So I guess this replaces `AliasResolver`, I wonder why you renamed it?
I think you forgot to commit the deletion of `AliasResolver` though

DavideD:
> What do you mean with nested associations? Embedded association containing another association?

Nevermind I get it

DavideD:
> Neo4j is a different beast as even *ToOne association are managed with relationships and as for now, only the embedded case was supported.

I think it would have made sense to keep it as separate commit

gsmet:
> I think it would have made sense to keep it as separate commit

Yes, the issue is it didn't work for Neo4j even after the general fix. That's why I made it a single commit.

But I will separate it.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/query/parsing/impl/Neo4jAliasResolver.java
gsmet: I didn't forget but looks like a rebase issue. Will fix.

The name was too general IMHO and I wanted to make it clear it was a Neo4j specific thing as it's based on the Neo4j relationships.

core/src/test/java/org/hibernate/ogm/backendtck/queries/Author.java
```diff
@@ -38,7 +38,7 @@
 	@OneToMany(mappedBy = "author")
 	private Set<Hypothesis> hypotheses;
 
-	@OneToOne(cascade = { CascadeType.PERSIST, CascadeType.REMOVE })
+	@ManyToOne(cascade = { CascadeType.PERSIST, CascadeType.REMOVE })
```
DavideD: I wonder if still have a test covering the OneToOne case

DavideD:
Looks good, I'm going to integrate it.
Just a few notes:
* I don't think we should change existing tests unless is required. After all they are giving us some confidence that things are still working after changes.
* I tend to rename a class before/after other changes, this make it easier to compare the content of the class with the previous commit it also make it to solve conflicts (if needed)

Thanks

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/681

Thanks @gsmet
Nice to see this done

gsmet:
@DavideD 

> I don't think we should change existing tests unless it's required. After all they are giving us some confidence that things are still working after changes.

As for the OneToOne -> ManyToOne change, the mapping was wrong as the other side was a OneToMany with a mappedBy. ORM fixes it automatically and changes it to a ManyToOne so I thought it would be better to have the correct mapping in the entity.

> I tend to rename a class before/after other changes, this make it easier to compare the content of the class with the previous commit, it also make it to solve conflicts (if needed)

Yeah, that's what I did again after seeing that the AliasResolver didn't get deleted. But I might have done something wrong because Git lost the information.

Thanks for the review!

DavideD:
> As for the OneToOne -> ManyToOne change, the mapping was wrong as the other side was a OneToMany with a mappedBy. ORM fixes it automatically and changes it to a ManyToOne so I thought it would be better to have the correct mapping in the entity.

Make sense, sorry for doubting :)
This means that some tests are currently missing for OneToOne associations, should we create a JIRA?

gsmet:
Done: https://hibernate.atlassian.net/browse/OGM-1062

PR #683, opened 2016-05-27
OGM-1067 Correctly size the Map/Set when a size is specified

gsmet:
https://hibernate.atlassian.net/browse/OGM-1067

The Maps/Sets initialized in CollectionHelper are systematically undersized leading to a systematic resize when we initially populate the Map/Set.

Fixing it by correctly calculating the capacity from the initial size specified.

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/util/impl/CollectionHelper.java
+++ b/core/src/main/java/org/hibernate/ogm/util/impl/CollectionHelper.java
@@ -20,6 +20,7 @@ import org.hibernate.ogm.dialect.query.spi.ClosableIterator;
  * Provides commonly used functionality around collections.
  *
  * @author Gunnar Morling
+ * @author Guillaume Smet
  */
 public class CollectionHelper {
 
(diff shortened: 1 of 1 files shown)


Sanne:
Wrong project ?

Sanne:
BTW I don't completely understand the rationale behind this, but I'll ask that on the *SEARCH* JIRA.

gsmet:
Yeah, as I said, bad headache... The issue was created in the wrong project. Fixed and commit message updated.

gsmet:
As for the rationale behind this, I added a few more elements in the JIRA.

Sanne:
> As for the rationale behind this, I added a few more elements in the JIRA.

The new explanation doesn't address the fact that buckets get shared ;)
Also the reference to a List is not helping as that doesn't share buckets so I totally see the point in that case.

Anyway, I can merge this the above comment is more to either have you reconsider in the light of bucket structure, or my curiosity about what I'm missing.

In terms of PR review, I think the javadoc should explain the same (why is it ok to ignore the load factor strategy in this case), but feel free to merge it if you think this is better explained another time.

Sanne:
@gunnarmorling just wondering, did you have similar questions? I remember you two talking about some collection sizing last few days.

gsmet:
@Sanne see for instance https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/Maps.java#L174

The Javadoc of Map is helpful too:
> The expected number of entries in the map and its load factor should be taken into account when setting its initial capacity, so as to minimize the number of rehash operations. If the initial capacity is greater than the maximum number of entries divided by the load factor, no rehash operations will ever occur.

PR #684, opened 2016-05-27
Automatic integration in WildFly

Sanne:
- https://hibernate.atlassian.net/browse/OGM-828

### DIFF
--- a/integrationtest/src/test/java/org/hibernate/ogm/test/integration/redis/RedisModuleMemberRegistrationWithTTLConfiguredIT.java
+++ b/integrationtest/src/test/java/org/hibernate/ogm/test/integration/redis/RedisModuleMemberRegistrationWithTTLConfiguredIT.java
@@ -11,17 +11,16 @@ import static org.junit.Assert.assertTrue;
 import javax.inject.Inject;
 import javax.persistence.EntityManagerFactory;
 
+import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.ogm.cfg.OgmProperties;
 import org.hibernate.ogm.datastore.redis.Redis;
 import org.hibernate.ogm.datastore.redis.RedisJsonDialect;
 import org.hibernate.ogm.datastore.redis.RedisProperties;
 import org.hibernate.ogm.datastore.redis.impl.RedisDatastoreProvider;
 import org.hibernate.ogm.datastore.spi.DatastoreProvider;
-import org.hibernate.ogm.jpa.impl.OgmEntityManagerFactory;
 import org.hibernate.ogm.test.integration.redis.model.PhoneNumber;
 import org.hibernate.ogm.test.integration.redis.service.PhoneNumberService;
 import org.hibernate.ogm.test.integration.testcase.ModuleMemberRegistrationScenario;
-import org.hibernate.ogm.test.integration.testcase.model.Member;
 import org.hibernate.ogm.test.integration.testcase.util.ModuleMemberRegistrationDeployment;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.junit.Arquillian;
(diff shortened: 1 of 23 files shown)


integrationtest/src/test/java/org/hibernate/ogm/test/integration/redis/RedisModuleMemberRegistrationWithTTLConfiguredIT.java
```diff
@@ -11,6 +11,8 @@
 import javax.inject.Inject;
 import javax.persistence.EntityManagerFactory;
 
+import org.hibernate.SessionFactory;
+import org.hibernate.engine.spi.SessionFactoryImplementor;
```
gsmet: Looks like you missed a CTRL + SHIFT + O.

modules/pom.xml
```diff
@@ -24,6 +24,7 @@
         <module.xml.aliases.basedir>${basedir}/src/main/aliases</module.xml.aliases.basedir>
         <hibernate.ogm.module.xml>${module.xml.basedir}/ogm/core/module.xml</hibernate.ogm.module.xml>
         <hibernate.ogm.module.slot>${parsedVersion.majorVersion}.${parsedVersion.minorVersion}</hibernate.ogm.module.slot>
+        <hibernate.ogm.jipijapa.module.slot>${project.version}</hibernate.ogm.jipijapa.module.slot>
```
gsmet: Is there any reason why we use `${project.version}` here while all other modules are using `${hibernate.ogm.module.slot}`?

Sanne: Yes. The "hibernate ogm module slot" is usually resolved to major.minor ( "5.0" or "5.1", etc..) so that people can refer to that, and we give them the latest within the chosen major.minor.

Some modules are internal though: people don't depend on them directly, but we rather inject the "right" versions for them. In this case we use the full version identifier, as we might be changing details which are not necessarily compatible (i.e. internal stuff).

for example, see the module "hibernate.ogm.neo4j-internal.module.slot" below: it also uses the full project version.

integrationtest/src/test/java/org/hibernate/ogm/test/integration/redis/RedisModuleMemberRegistrationWithTTLConfiguredIT.java
Sanne: what do you mean?

gsmet: The build failed because of unused imports in this file.

modules/pom.xml
gsmet: Thanks for the explanation!

integrationtest/src/test/java/org/hibernate/ogm/test/integration/redis/RedisModuleMemberRegistrationWithTTLConfiguredIT.java
Sanne: ah, now I get it, sorry! Looks like I didn't push my latest correctly, fixing now.. thanks a lot for noticing.

I should have noticed the build status :)

Sanne:
Updated to remove the redundant import statements. Thanks @gsmet !

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/684
Thanks

DavideD:
@Sanne, [OGM-828](https://hibernate.atlassian.net/browse/OGM-828) is titled: "Need to list entities when using Hibernate OGM on WildFly", does this patch solve the issue? It doesn't look like we have a test where we don't enlist the entities. Am I wrong?

gunnarmorling:
> It doesn't look like we have a test where we don't enlist the entities. Am I wrong?

He removed all the explicit class references from the integration tests, no?

DavideD:
> He removed all the explicit class references from the integration tests, no?

Right, my bad, I've checked the persistence.xml generated and they were still listed but it was an old one generated before I applied the PR. All good then.

Sanne:
> Right, my bad, I've checked the persistence.xml generated and they were still listed but it was an old one generated before I applied the PR. All good then.

If I missed some configuration files, feel free to clean them up and push :)

They are useful as regression tests to cover this too.

PR #685, opened 2016-06-01
Assorted small fixes to the batch support

gsmet:
https://hibernate.atlassian.net/browse/OGM-1068
https://hibernate.atlassian.net/browse/OGM-1069
https://hibernate.atlassian.net/browse/OGM-1070

Assorted small fixes to the batch support (extracted from a bigger patch).

### DIFF
--- a/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/dialectinvocations/GridDialectOperationInvocationsTest.java
+++ b/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/dialectinvocations/GridDialectOperationInvocationsTest.java
@@ -121,8 +121,6 @@ public class GridDialectOperationInvocationsTest extends OgmTestCase {
 				"getAssociation",
 				"getTuple",
 				"getAssociation",
-				"removeAssociation",
-				"removeTuple",
 				"executeBatch"
 				);
 	}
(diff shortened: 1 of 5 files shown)


mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/dialectinvocations/GridDialectOperationInvocationsTest.java
```diff
@@ -121,8 +121,6 @@ public void insertGetOfEntityAndElementCollectionAndDelete() throws Exception {
 				"getAssociation",
 				"getTuple",
 				"getAssociation",
-				"removeAssociation",
```
gunnarmorling: I suppose they are part of the subsequent `executeBatch()`? I might be wrong, but there should be a way for asserting the batch contents, too (if not we should create such testing facility as part of a separate issue).

gunnarmorling:
@gsmet Feel free to merge yourself, considering my one comment above.

Regarding JIRA issues, I suppose I'd have used only one - with the three commits as you did them. Readers of the changelog won't get much from the different JIRA issues, and for our own reference we cand decribe several minor things as these three within one issue.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/dialectinvocations/GridDialectOperationInvocationsTest.java
gsmet: Yes, they are included in the `executeBatch()` call. I agree on the need for a testing facility for that.

Created a new issue here to track this: https://hibernate.atlassian.net/browse/OGM-1071

gsmet:
Merged! Thanks!

PR #686, opened 2016-06-02
OGM-1064 Replace separate association management methods by upsertTupleWithAssociations()

gsmet:
https://hibernate.atlassian.net/browse/OGM-1064

Note: this is a work in progress

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/compensation/impl/InvocationCollectingGridDialect.java
+++ b/core/src/main/java/org/hibernate/ogm/compensation/impl/InvocationCollectingGridDialect.java
@@ -33,6 +33,7 @@ import org.hibernate.ogm.compensation.operation.impl.RemoveAssociationImpl;
 import org.hibernate.ogm.compensation.operation.impl.RemoveTupleImpl;
 import org.hibernate.ogm.compensation.operation.impl.RemoveTupleWithOptimisticLockImpl;
 import org.hibernate.ogm.compensation.operation.impl.UpdateTupleWithOptimisticLockImpl;
+import org.hibernate.ogm.dialect.batch.spi.GroupedChangesToEntityOperation;
 import org.hibernate.ogm.dialect.batch.spi.InsertOrUpdateAssociationOperation;
 import org.hibernate.ogm.dialect.batch.spi.InsertOrUpdateTupleOperation;
 import org.hibernate.ogm.dialect.batch.spi.Operation;
@@ -97,24 +98,14 @@ public class InvocationCollectingGridDialect extends ForwardingGridDialect<Seria
 			while ( operation != null ) {
 				newQueue.add( operation );
 
-				if ( operation instanceof InsertOrUpdateTupleOperation ) {
-					InsertOrUpdateTupleOperation insertOrUpdateTuple = (InsertOrUpdateTupleOperation) operation;
-					operations.add( new InsertOrUpdateTupleImpl( insertOrUpdateTuple.getEntityKey(), insertOrUpdateTuple.getTuple() ) );
+				if ( operation instanceof GroupedChangesToEntityOperation ) {
+					GroupedChangesToEntityOperation groupedChangesOnEntity = (GroupedChangesToEntityOperation) operation;
+					for (Operation groupedOperation : groupedChangesOnEntity.getOperations()) {
+						operations.add( getSimpleGridDialectOperations( groupedOperation ) );
+					}
 				}
-				else if ( operation instanceof RemoveTupleOperation ) {
-					RemoveTupleOperation removeTuple = (RemoveTupleOperation) operation;
-					operations.add( new RemoveTupleImpl( removeTuple.getEntityKey() ) );
-				}
-				else if ( operation instanceof InsertOrUpdateAssociationOperation ) {
-					InsertOrUpdateAssociationOperation insertOrUpdateAssociationOperation = (InsertOrUpdateAssociationOperation) operation;
-					operations.add( new InsertOrUpdateAssociationImpl(
-							insertOrUpdateAssociationOperation.getAssociationKey(),
-							insertOrUpdateAssociationOperation.getAssociation() )
-					);
-				}
-				else if ( operation instanceof RemoveAssociationOperation ) {
-					RemoveAssociationOperation removeAssociationOperation = (RemoveAssociationOperation) operation;
-					operations.add( new RemoveAssociationImpl( removeAssociationOperation.getAssociationKey() ) );
+				else {
+					operations.add( getSimpleGridDialectOperations( operation ) );
 				}
 
 				operation = queue.poll();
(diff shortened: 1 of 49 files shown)


core/src/main/java/org/hibernate/ogm/service/impl/OgmIntegrator.java
```diff
@@ -73,10 +72,8 @@ private void attachEventContextManagingListenersIfRequired(SessionFactoryService
 		eventListenerRegistry.addDuplicationStrategy( EventContextManagingFlushEventListenerDuplicationStrategy.INSTANCE );
 		eventListenerRegistry.getEventListenerGroup( EventType.FLUSH ).appendListener( new EventContextManagingFlushEventListener( stateManager ) );
 
-		if ( getIntegrator( JpaIntegrator.class, serviceRegistry ) != null ) {
```
gunnarmorling: How comes this _if_ has vanished?

gunnarmorling: One a more general note, do you know why something is attached to the persist lifecycle to begin with? I should know as I wrote it, but it completely slipped my mind. I'm finding the `executeBatch()` in `OperationsQueueLifecycle` a bit weird in the context of persist. Wondering whether that's alright.

core/src/main/java/org/hibernate/ogm/dialect/batch/spi/GroupableEntityOperation.java
```diff
@@ -0,0 +1,20 @@
...
+package org.hibernate.ogm.dialect.batch.spi;
+
+import org.hibernate.ogm.model.key.spi.EntityKey;
+
+/**
+ * Represents an {@link Operation} we can group by entity
+ *
+ * @author Guillaume Smet
+ */
+public interface GroupableEntityOperation extends Operation {
```
gunnarmorling: Is that interface actually needed? Or could/should we consider all operations to be scoped in the context of an (embedding) entity? I'd lean towards the latter on first thought.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -222,7 +224,7 @@ else if ( isInTheQueue( key, tupleContext ) ) {
 
 	private static boolean isInTheQueue(EntityKey key, TupleContext tupleContext) {
```
gunnarmorling: There also is `isInTheInsertionQueue()`. Seems one can go away.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1190,42 +1139,103 @@ private void executeBatchRemove(Map<DBCollection, BatchInsertionTask> inserts, R
...
 
-	private void executeBatchUpdate(Map<DBCollection, BatchInsertionTask> inserts, InsertOrUpdateTupleOperation tupleOperation) {
-		EntityKey entityKey = tupleOperation.getEntityKey();
-		Tuple tuple = tupleOperation.getTuple();
-		MongoDBTupleSnapshot snapshot = (MongoDBTupleSnapshot) tupleOperation.getTuple().getSnapshot();
-		WriteConcern writeConcern = getWriteConcern( tupleOperation.getTupleContext() );
-
-		if ( INSERT == snapshot.getSnapshotType() ) {
-			prepareForInsert( inserts, snapshot, entityKey, tuple, writeConcern );
+	private void executeBatchUpdate(Map<DBCollection, BatchInsertionTask> inserts, List<MongoDBTupleSnapshot> insertSnapshots,
```
gunnarmorling: Can we break down that method into several smaller ones? I cannot give a specific threshold, but this one seems much too large. Splitting it up helps with making things easier to follow, as it a) forces you into cutting sensible pieces and b) these pieces have a (method) name.

gunnarmorling:
Shouldn't there be a new method on `GridDialect`, `upsertTupleWithAssociations()` or similar? I don't think we should _require_ dialects to support batching per se. It's only that we'd re-use the batching machinery to group the operations by embedding entity. Then if a dialect does not support batching, we'd invoke that new method on `GridDialect`, otherwise we'd invoke `executeBatch()`, where the passed operations queue would contain a list of operation objects, each equivalent to an invocation of `upsertTupleWithAssociations()`.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1190,42 +1139,103 @@ private void executeBatchRemove(Map<DBCollection, BatchInsertionTask> inserts, R
...
+			GroupedChangesToEntityOperation groupedOperation) {
+		if ( groupedOperation.getOperations().isEmpty() ) {
```
gunnarmorling: Can that actually happen? Why would we pass an operation group object in that case?

core/src/main/java/org/hibernate/ogm/dialect/batch/spi/GroupableEntityOperation.java
gsmet: Currently, I haven't grouped the RemoveTuple operation. I will consider doing it and if so, this interface would indeed be useless.

core/src/main/java/org/hibernate/ogm/service/impl/OgmIntegrator.java
gsmet: I made a commit just for it ;). From what I've seen, the persist operation is used when the key is generated by the datastore.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
gsmet: That's the plan. I didn't do it right now because it was easier to follow the flow of operations and easier to debug it.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
gsmet: It did happen when I considered that a RemoveTupleOperation could clear the operations registered before but it wasn't a good idea as the association removals weren't managed. I'll remove this.

gsmet:
> Shouldn't there be a new method on GridDialect, upsertTupleWithAssociations() or similar?

This is the second part of the patch I'm working on right now. The fact is I was planning to talk to you before you reviewing this ;). Could you connect on HipChat?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
gsmet: They don't have the same parameters. But I should have renamed this one too. Will do.

core/src/main/java/org/hibernate/ogm/compensation/impl/InvocationCollectingGridDialect.java
```diff
@@ -97,24 +98,14 @@ public void executeBatch(OperationsQueue queue) {
 			while ( operation != null ) {
```
gunnarmorling: Now that batching is always enabled, I think we could unify these different operation object hierarchies?

gsmet: With this patch, batching is not always enabled. And I'm still not convinced it's a good idea to always enable it. I'm working on Redis JSON and Ehcache to see how it goes. I'll decide what's best from there.

gunnarmorling: Ok, but the batch delegator is always running, right? What I mean is, _something_ that collects grid dialect invocations in an object-style approach. Ideally, we should have this only once, maybe implementing different public interfaces geared towards different usages, but only paying once the penalty for object allocation.

gsmet:
Closing for now.

PR #687, opened 2016-06-06
OGM-366 Neo4j remote integration

DavideD:

@gsmet I've appliedthe refactoring we discussed, I also had to rebase to the latest changes.

I've decided not to add the `QueryContext` for now and continue to use the `TupleContext`. I deal with the case where a session or an `EntityMetdataInfo` is null by throwing an exception if the dialect try to access some of the methods in the TupleContext when it shouldn't. I prefer this solution because it seems there are already other ways to guard against this and I don't want to update the signature in this PR  that it is already quite big.


core/src/test/java/org/hibernate/ogm/backendtck/associations/manytoone/ManyToOneTest.java
```diff
@@ -42,8 +42,8 @@ public void testUnidirectionalManyToOne() throws Exception {
 		session.persist( emmanuel );
 		session.persist( jerome );
 		session.flush();
-		assertThat( getNumberOfEntities( sessions ) ).isEqualTo( 3 );
-		assertThat( getNumberOfAssociations( sessions ) ).isEqualTo( expectedAssociations() );
+		assertThat( getNumberOfEntities( session ) ).isEqualTo( 3 );
+		assertThat( getNumberOfAssociations( session ) ).isEqualTo( expectedAssociations() );
 		transaction.commit();
 		assertThat( getNumberOfEntities( sessions ) ).isEqualTo( 3 );
```
gsmet: Is there a reason why you didn't change all the calls and get rid of the SessionFactory based versions?

gsmet:
090fbd951a86bc485cb0683e9ba794ac8917b70b OGM-366 Create GridDialectType.NEO4J_REMOTE and skip tests

Typos in the commit comment:
suppose -> supposed
max indexer -> mass indexer

core/src/test/java/org/hibernate/ogm/backendtck/associations/manytoone/ManyToOneTest.java
gsmet: Apart from giving me less things to review, of course, which is a very valid reason ;)

core/src/test/java/org/hibernate/ogm/backendtck/associations/collection/manytomany/ManyToManyTest.java
```diff
@@ -135,7 +135,7 @@ private void delete(Session session, Object... entities) {
 	}
 
 	private int expectedAssociationNumber() {
-		if ( TestHelper.getCurrentDialectType().equals( GridDialectType.NEO4J ) ) {
+		if ( TestHelper.getCurrentDialectType().equals( GridDialectType.NEO4J ) || TestHelper.getCurrentDialectType().equals( GridDialectType.NEO4J_REMOTE )) {
```
gsmet: Minor nitpicking but we might as well be consistent in the way we test for this. See the change just above this one.

core/src/test/java/org/hibernate/ogm/backendtck/associations/manytoone/ManyToOneTest.java
```diff
@@ -117,7 +117,7 @@ public void canPersistAssociationToDetachedEntity() throws Exception {
 	}
 
 	private Long expectedAssociations() {
-		if ( TestHelper.getCurrentDialectType() == GridDialectType.NEO4J ) {
+		if ( TestHelper.getCurrentDialectType() == GridDialectType.NEO4J || TestHelper.getCurrentDialectType() == GridDialectType.NEO4J_REMOTE ) {
```
gsmet: Same here.

gsmet:
51f3d99988e7c46f69e2f635c353973ee7707740 Add classes mapping the JSON reponse from the Neo4j server

Typos in the commit comment:
a a -> ?
realtive -> relative

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/json/impl/Statements.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * The list of queries to execute via Rest.
+ *
+ * @author Davide D'Alto
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class Statements {
+
+	private List<Statement> statements = new ArrayList<Statement>( 5 );
```
gsmet: Why 5?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/DatabaseIdentifier.java
```diff
@@ -0,0 +1,93 @@
...
+	 * The name of the database
+	 *
+	 * @return the name of the database
+	 */
+	public String getDatabaseName() {
+		return databaseName;
+	}
+
+	/**
+	 * Returns the URI of the Neo4j server, e.g. "http:://localhost:5984".
```
gsmet: s/::/:/

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/DatabaseIdentifier.java
```diff
@@ -0,0 +1,93 @@
...
+	 *
+	 * @return the URI of the Neo4j server
+	 */
+	public URI getServerUri() {
+		return serverUri;
+	}
+
+	/**
+	 * Returns the URI of the database, e.g. "http:://localhost:5984/mydb".
```
gsmet: s/::/:/

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/DatabaseIdentifier.java
```diff
@@ -0,0 +1,93 @@
...
+	private final URI databaseUri;
+
+	public DatabaseIdentifier(String host, int port, String databaseName, String userName, String password) throws MalformedURLException, URISyntaxException {
+		this.host = host;
+		this.port = port;
+		this.databaseName = databaseName;
+		this.userName = userName;
+		this.password = password;
+
+		this.serverUri = new URL( PROTOCOL, host, port, "" ).toURI();
```
gsmet: Is there a reason why the protocol is hardcoded? I would have expected us to support HTTPS? That being said, we can fix it later.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/Neo4jClient.java
```diff
@@ -0,0 +1,120 @@
...
+ * Access point to the remote Neo4j server.
+ *
+ * @author Davide D'Alto
+ */
+public class Neo4jClient implements AutoCloseable {
+
+	/**
+	 * Size of the client connection pool used by the RestEasy HTTP client
+	 */
+	private static final int CONNECTION_POOL_SIZE = 10;
```
gsmet: Shouldn't we allow to configure this?

core/src/test/java/org/hibernate/ogm/backendtck/associations/manytoone/ManyToOneTest.java
DavideD: > Apart from giving me less things to review, of course, which is a very valid reason ;)
 
That's the main reason, I prefer to keep the amount of changes to a minimum. For this cases in particular, we might have some tests that don't have a session and in general there is nothing wrong about using the factory. This change is about giving an additional option for those scenario where you need to make some assertions inside a specific transaction.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/json/impl/Statements.java
DavideD: Good question, I think it was a guesstimate, but I'll remove it.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/Neo4jClient.java
```diff
@@ -0,0 +1,120 @@
...
+		}
+
+		clientBuilder.register( XStreamRequestHeaderFilter.INSTANCE );
+
+		// using a connection pool size > 1 causes a thread-safe pool implementation to be used under the hoods
+		return clientBuilder.connectionPoolSize( CONNECTION_POOL_SIZE ).build();
+	}
+
+	public Response authenticate(String username) {
+		return authenticationClient.authenticate( username );
```
gsmet: Do I understand things correctly: authenticate doesn't do a stateful authentication, it just tests that the authentication is OK and the authentication is done using basic authentication for each call?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/transaction/impl/Transaction.java
```diff
@@ -0,0 +1,40 @@
...
+	public void commit() {
+		remote.commit( txId );
+	}
+
+	public void rollback() {
+		remote.rollback( txId );
+	}
+
+	public Long getId() {
+		return txId.longValue();
```
gsmet: Either you want to return a primitive long or you don't need the longValue() call.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/DatabaseIdentifier.java
DavideD: No, no particular reason, this is a copy of an existing class.
 
I think it deserve a separate issue

gsmet:
0425d7e816699997b8ac3a7a02552f2885ce48bf Add Neo4j client
iwll -> will
althought -> although

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/Neo4jClient.java
DavideD: yes, I think it can be a follow-up commit or JIRA

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/Neo4jClient.java
DavideD: Yes, 
 it seems thought that it is an optional method and  I'm not using it anywhere. I think I'm going to remove it.

gsmet:
1e0fca3dd584699e73caf5e49eba5c4f5b763784 Extract reusable parts from the Neo4j embedded dialect

Typos in the commit comment:
extract -> extracts
part -> parts
tne -> the

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/Neo4jClient.java
gsmet: OK, maybe we could use it at startup to check that we can effectively connect to the Neo4j remote server?

core/src/main/java/org/hibernate/ogm/util/impl/Log.java
```diff
@@ -292,4 +292,7 @@
 
 	@Message(id = 86, value = "Transaction identifier not available")
 	HibernateException transactionIdIsNotAvailable();
+
+	@Message(id = 87, value = "The tuple context is not available, probably because we are dealing with more than a single entity")
```
gsmet: I think "single entity **type**" would be better.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/Neo4jClient.java
DavideD: ok, I'll check

core/src/main/java/org/hibernate/ogm/util/impl/TupleContextHelper.java
```diff
@@ -6,29 +6,103 @@
...
+			TupleContext tupleContext = new OnlyWithTransactionContext( transactionContext );
+			return tupleContext;
+		}
+		else {
+			return null;
+		}
+	}
+
+	/**
+	 * This {@link TupleContext} can be used for those use case where we don't necessarily have a tuple context but we
```
gsmet: use case -> use cases

core/src/main/java/org/hibernate/ogm/util/impl/TupleContextHelper.java
```diff
@@ -6,29 +6,103 @@
...
+	 * still need a {@link TransactionContext}.
+	 */
+	private static class OnlyWithTransactionContext implements TupleContext {
```
gsmet: I understand the point you exposed in your PR comment as the PR is already big enough. But I don't like this.

I think we should try to get rid of this once we got this PR merged. @gunnarmorling any opinion about this?

core/src/test/java/org/hibernate/ogm/backendtck/associations/collection/manytomany/ManyToManyTest.java
DavideD: I'm not sure I understand what you are referring to  ...

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/AssociationQueries.java
```diff
@@ -0,0 +1,604 @@
...
+import org.neo4j.graphdb.Result;
+
+/**
+ * Container for the queries related to one association family in Neo4j. Unfortunately, we cannot use the same queries
+ * for all associations, as Neo4j does not allow to parameterize on node labels which would be required, as the
+ * association table is stored as a label.
+ *
+ * @author Davide D'Alto
+ */
+public class AssociationQueries extends QueriesBase {
```
gsmet: Shouldn't this be abstract?

core/src/main/java/org/hibernate/ogm/util/impl/TupleContextHelper.java
DavideD: > I think we should try to get rid of this once we got this PR merged.

Sounds good

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/EntityQueries.java
```diff
@@ -0,0 +1,667 @@
...
+import org.hibernate.ogm.dialect.spi.TupleContext;
+import org.hibernate.ogm.model.key.spi.AssociatedEntityKeyMetadata;
+import org.hibernate.ogm.model.key.spi.EntityKey;
+import org.hibernate.ogm.model.key.spi.EntityKeyMetadata;
+import org.hibernate.ogm.util.impl.EmbeddedHelper;
+
+/**
+ * @author Davide D'Alto
+ */
+public class EntityQueries extends QueriesBase {
```
gsmet: Shouldn't this be abstract?

core/src/main/java/org/hibernate/ogm/datastore/impl/AvailableDatastoreProvider.java
```diff
@@ -19,6 +19,7 @@
 	EHCACHE( "org.hibernate.ogm.datastore.ehcache.impl.EhcacheDatastoreProvider" ),
 	MONGODB( "org.hibernate.ogm.datastore.mongodb.impl.MongoDBDatastoreProvider" ),
 	FONGO( "org.hibernate.ogm.datastore.mongodb.impl.FongoDBDatastoreProvider" ),
+	NEO4J( "org.hibernate.ogm.datastore.neo4j.remote.impl.RemoteNeo4jDatastoreProvider" ),
```
gsmet: Shouldn't it be called NEO4J_REMOTE?

neo4j/pom.xml
```diff
@@ -156,4 +156,25 @@
...
+            <properties>
+                <datastoreProvider>neo4j_embedded</datastoreProvider>
+            </properties>
+        </profile>
+    
+        <!-- Executes the tests against a separately started Neo4j server -->
+        <profile>
+            <id>neo4j-remote</id>
+            <properties>
+                <datastoreProvider>neo4j</datastoreProvider>
```
gsmet: I would have expected neo4j_remote here.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/RemoteNeo4jDialect.java
```diff
@@ -0,0 +1,559 @@
...
+ * A {@link Tuple} is saved as a {@link Node} where the columns are converted into properties of the node.<br>
+ * An {@link Association} is converted into a {@link Relationship} identified by the {@link AssociationKey} and the
+ * {@link RowKey}. The type of the relationship is the value returned by
+ * {@link AssociationKeyMetadata#getCollectionRole()}.
+ * <p>
+ * If the value of a property is set to null the property will be removed (Neo4j does not allow to store null values).
+ *
+ * @author Davide D'Alto &lt;davide@hibernate.org&gt;
+ */
+public class RemoteNeo4jDialect extends BaseNeo4jDialect {
```
gsmet: I'm still a bit worried about this class. It duplicates quite a lot of things.

Have you tried to use generics on BaseNeo4jDialect with a couple of methods like newEntityQueries(),... , newTupleSnapshot() and so on?

Not sure it's a good idea but it might be worth a try if you haven't already done so.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/Neo4jProperties.java
```diff
@@ -34,6 +34,31 @@
 	 */
 	public static final String SEQUENCE_QUERY_CACHE_MAX_SIZE = "hibernate.ogm.neo4j.sequence_query_cache_max_size";
 
+	/**
+	 * Socket inactivity timeout in milliseconds.
+	 */
+	public static String SOCKET_TIMEOUT = "hibernate.ogm.neo4j.client.socketTimeout";
```
gsmet: socket_timeout would be more consistent with the others?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/Neo4jProperties.java
```diff
@@ -34,6 +34,31 @@
...
+
+	/**
+	 * the timeout in millisecond to make an initial socket connection.
+	 */
+	public static final String ESTABLISH_CONNECTION_TIMEOUT = "hibernate.ogm.neo4j.client.establish_connection_timeout";
+
+	/**
+	 * Property for setting the password to connect with. Accepts {@code String}.
+	 */
+	public static String TTL_TIMEOUT = "hibernate.ogm.neo4j.client.socketTimeOut";
```
gsmet: There's something wrong here: the comment is wrong and the property is the same as SOCKET_TIMEOUT. Looks like a leftover as it's not used anywhere. I think we can remove it.

core/src/main/java/org/hibernate/ogm/datastore/impl/AvailableDatastoreProvider.java
DavideD: At the moment, I'm using NEO4J for the remote and NEO4J_EMBEDDED for the embedded with the assumption that neo4j remote is the most common use case. NEO4J is much nicer to use.

neo4j/pom.xml
DavideD: see previous comment

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/Neo4jProperties.java
```diff
@@ -34,6 +34,31 @@
...
 
+	/**
+	 * Socket inactivity timeout in milliseconds.
+	 */
+	public static String SOCKET_TIMEOUT = "hibernate.ogm.neo4j.client.socketTimeout";
+
+	/**
+	 * How long will we wait (in milliseconds) to get a connection?
+	 */
+	public static final String CONNECTION_CHECKOUT_TMEOUT = "hibernate.ogm.neo4j.client.connection_checkout_timeout";
```
gsmet: TMEOUT -> TIMEOUT

core/src/test/java/org/hibernate/ogm/test/util/impl/TupleContextHelperTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.hibernate.ogm.model.key.spi.EntityKeyMetadata;
+import org.hibernate.ogm.model.spi.EntityMetadataInformation;
+import org.hibernate.ogm.util.impl.TupleContextHelper;
+import org.hibernate.ogm.utils.OgmTestCase;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+/**
+ * Test the {@link TupleContextHelper} utility classe.
```
gsmet: classe -> class
you're so French :)

core/src/test/java/org/hibernate/ogm/test/util/impl/TupleContextHelperTest.java
```diff
@@ -0,0 +1,90 @@
...
+ * <p>
+ * An {@link EntityMetadataInformation} will be null when there is more than one entity is involved.
```
gsmet: there is more than one entity is involved -> there is more than one entity type involved

core/src/test/java/org/hibernate/ogm/test/util/impl/TupleContextHelperTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+	@Test
+	public void testTransactionContextIsNotNull() throws Exception {
+		TupleContext tupleContext = TupleContextHelper.tupleContext( (SessionImplementor) openSession(), null );
+		assertThat( tupleContext.getTransactionContext() ).isNotNull();
+	}
+
+	@Test
+	public void testTupleContextIsNull() throws Exception {
+		// This can happen when we start the mass-indexer without a session on more than one entity.
```
gsmet: one entity -> one entity type

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/Neo4jProperties.java
DavideD: I read the value in `Neo4jConfiguration` and then is set when the `Neo4jClient` is created (Neo4jClient#createRestClient())

core/src/test/java/org/hibernate/ogm/test/util/impl/TupleContextHelperTest.java
DavideD: In italian is "classe" :)

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/RemoteNeo4jDialect.java
DavideD: I'll have a look.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/Neo4jProperties.java
gsmet: Nope, not for this one. It's SOCKET_TIMEOUT that you read.

core/src/main/java/org/hibernate/ogm/datastore/impl/AvailableDatastoreProvider.java
gsmet: Yeah well, my problem is that it's not consistent at all with the code where Neo4j* is for embedded (most of the time) and RemoteNeo4j* for remote.

I'll post a general comment on this at the end of my review so that we can deal with this.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jSequenceGenerator.java
```diff
@@ -0,0 +1,253 @@
...
+
+	private final Neo4jClient neo4jDb;
+
+	public RemoteNeo4jSequenceGenerator(Neo4jClient neo4jDb, int sequenceCacheMaxSize) {
+		this.neo4jDb = neo4jDb;
+		this.queryCache = new BoundedConcurrentHashMap<String, Statements>( sequenceCacheMaxSize, 20, BoundedConcurrentHashMap.Eviction.LIRS );
+	}
+
+	/**
+	 * Create the sequence nodes setting the initial value if the node does not exists already.
```
gsmet: exists -> exist

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jSequenceGenerator.java
```diff
@@ -0,0 +1,253 @@
...
+ *
+ * @author Davide D'Alto &lt;davide@hibernate.org&gt;
+ * @author Gunnar Morling
+ */
+public class RemoteNeo4jSequenceGenerator extends BaseSequenceGenerator {
+
+	/**
+	 * Query for creating SEQUENCE nodes.
+	 */
+	protected static final String SEQUENCE_CREATION_QUERY = "MERGE (n:" + NodeLabel.SEQUENCE.name() + " {" + SEQUENCE_NAME_PROPERTY + ": {sequenceName}} ) RETURN n";
```
gsmet: Why is the creation query different from the embedded one? It doesn't take into account initialValue?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jSequenceGenerator.java
```diff
@@ -0,0 +1,253 @@
...
+
+	protected static final String SEQUENCE_LOCK_QUERY = "MATCH (n:" + NodeLabel.SEQUENCE.name() + ") WHERE n." + SEQUENCE_NAME_PROPERTY + " = {sequenceName} "
+			+ " SET n.__locked = true " + " RETURN n";
```
gsmet: Is there a reason why the locks are managed differently between embedded and remote?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jSequenceGenerator.java
```diff
@@ -0,0 +1,253 @@
...
+
+	/**
+	 * Generate the next value in a sequence for a given {@link IdSourceKey}.
+	 *
+	 * @param idSourceKey identifies the generator
+	 * @param increment the difference between to consecutive values in the sequence
+	 * @param initialValue the initial value of the given generator
+	 * @return the next value in a sequence
+	 */
+	public Number nextValue(IdSourceKey idSourceKey, int increment, int initialValue) {
```
gsmet: The parameters increment and initialValue are interverted between the embedded and the remote versions. Doesn't look like a good idea. Maybe having an abstract method in the parent class would be handy to avoid this?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jSequenceGenerator.java
```diff
@@ -0,0 +1,253 @@
...
+		queryBuilder.append( idSourceKeyMetadata.getValueColumnName() );
+		String query = queryBuilder.toString();
+		return query;
+	}
+
+	/**
+	 * Generate the next value in a sequence for a given {@link IdSourceKey}.
+	 *
+	 * @param idSourceKey identifies the generator
+	 * @param increment the difference between to consecutive values in the sequence
```
gsmet: to -> two

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jSequenceGenerator.java
```diff
@@ -0,0 +1,253 @@
...
+		StatementsResponse statementsResponse = neo4jDb.executeQueriesInNewTransaction( statements );
+		List<StatementResult> results = statementsResponse.getResults();
+		Integer nextValue = (Integer) results.get( 1 ).getData().get( 0 ).getRow().get( 0 );
+		// sequence nodes are expected to have been created up-front
+		if ( idSourceKey.getMetadata().getType() == IdSourceType.SEQUENCE ) {
+			if ( nextValue == null ) {
+				throw new HibernateException( "Sequence missing: " + sequenceName );
+			}
+		}
+		return nextValue.longValue() - increment;
```
gsmet: This one looks weird, why would you substract the increment from the nextValue?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jSequenceGenerator.java
DavideD: Cypher does not provide a way to acquire a lock on a node. The lock property only purpose is to make sure that a lock is acquired before updating the sequence.
In embedded mode, I can use a nice API and I can acquire a lock on a single node easily.
Therefore I prefer to use a different approach in the two cases.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jSequenceGenerator.java
DavideD: Mmmh... are you talking about RemoteNeo4jSequenceGenerator#nextvalue and Neo4jSequenceGenerator#nextValue? I don't see a difference...

But I see a difference between nextValue and the private methods that follow

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jEntityQueries.java
```diff
@@ -0,0 +1,334 @@
...
+	public RemoteNeo4jEntityQueries(EntityKeyMetadata entityKeyMetadata, TupleContext tupleContext) {
+		super( entityKeyMetadata, tupleContext );
+	}
+
+	public Node findEntity(Neo4jClient executionEngine, Long transactionId, Object[] columnValues) {
+		Map<String, Object> params = params( columnValues );
+		String query = getFindEntityQuery();
+		List<StatementResult> results = executeQuery( executionEngine, transactionId, query, params );
+		if ( results != null ) {
+			Row row = results.get( 0 ).getData().get( 0 );
```
gsmet: I'm a bit surprised by the structure of the results which look overly complex.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jSequenceGenerator.java
gsmet: OK, I must admit that I'm a bit worried about the level of duplications we have here. If we consider the embedded case to be a not so common case, maybe we should avoid as much duplication as possible even if it's a little less effective/nice. Apart from the execution of the queries/extraction of the result, I would have expected both classes to be essentially the same.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jSequenceGenerator.java
gsmet: Btw, same typo in the embedded one, probably a preexisting one.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jSequenceGenerator.java
gsmet: Oh, yes sorry. It's effectively between nextValue and all the private methods generating queries.

Btw, my comment on the abstract method still stands as the return type is not the same and it would probably be better for these methods to be consistent. (This if we cannot find a way to avoid the code duplication)

DavideD: > my comment on the abstract method still stands a ...

no problem, I will add it

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/RemoteNeo4jSchemaDefiner.java
```diff
@@ -0,0 +1,119 @@
...
+
+		Statements constraintStatements = new Statements();
+		provider.getSequenceGenerator().createSequencesConstraints( constraintStatements, sequences );
+		provider.getSequenceGenerator().createUniqueConstraintsForTableSequences( constraintStatements, idSourceKeyMetadata );
+
+		StatementsResponse response = provider.getDatabase().executeQueriesInNewTransaction( constraintStatements );
+		List<StatementResult> results = response.getResults();
+		// TODO: Check response for errors
+
+		// We create the sequencese in a separate transaction because
```
gsmet: sequencese -> sequences

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/RemoteNeo4jSchemaDefiner.java
```diff
@@ -0,0 +1,119 @@
...
+	private void createSequences(Database database, Iterable<IdSourceKeyMetadata> idSourceKeyMetadata, RemoteNeo4jDatastoreProvider provider) {
+		List<Sequence> sequences = sequences( database );
+
+		Statements constraintStatements = new Statements();
+		provider.getSequenceGenerator().createSequencesConstraints( constraintStatements, sequences );
+		provider.getSequenceGenerator().createUniqueConstraintsForTableSequences( constraintStatements, idSourceKeyMetadata );
+
+		StatementsResponse response = provider.getDatabase().executeQueriesInNewTransaction( constraintStatements );
+		List<StatementResult> results = response.getResults();
+		// TODO: Check response for errors
```
gsmet: Maybe open a JIRA if you don't plan to fix them in this PR?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/RemoteNeo4jSchemaDefiner.java
```diff
@@ -0,0 +1,119 @@
...
+
+		// We create the sequencese in a separate transaction because
+		// Neo4j does not allow the creation of constraints and graph elements in the same transaction
+		Statements sequenceStatements = new Statements();
+		provider.getSequenceGenerator().createSequences( sequenceStatements, sequences );
+		response = provider.getDatabase().executeQueriesInNewTransaction( sequenceStatements );
+		results = response.getResults();
+		// TODO: Check response for errors
```
gsmet: Maybe open a JIRA if you don't plan to fix them in this PR?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jSequenceGenerator.java
DavideD: The only way I found to make it work in a multi-threaded environment is to first increment the value and then read it . Our API allows for an initial value and to make sure that I'm actually reading the correct one the first time I need to decrement the value I obtain from the db.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/RemoteNeo4jSchemaDefiner.java
DavideD: Yes

DavideD: By the way, no need to add a coment for evey TODO (unless is something different), I will open a JIRA for them

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/transaction/impl/RemoteJtaTransactionCoordinator.java
```diff
@@ -0,0 +1,153 @@
...
+					failure();
+				}
+				else {
+					success();
+				}
+			}
+		}
+	}
+
+	private class RemoteTransactionDriver extends ForwardingTransactionDriver implements IdentifiableDriver {
```
gsmet: Is there a reason why you don't use the same pattern as for Neo4jJtaTransactionCoordinator? Neo4jJtaTransactionCoordinator just relies on Neo4jSynchronization.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/transaction/impl/RemoteResourceLocalTransactionCoordinator.java
```diff
@@ -0,0 +1,339 @@
...
+import org.hibernate.resource.transaction.TransactionCoordinator;
+import org.hibernate.resource.transaction.TransactionCoordinatorBuilder;
+import org.hibernate.resource.transaction.internal.SynchronizationRegistryStandardImpl;
+import org.hibernate.resource.transaction.spi.TransactionCoordinatorOwner;
+import org.hibernate.resource.transaction.spi.TransactionStatus;
+
+/**
+ * An implementation of TransactionCoordinator based on managing a transaction through a Neo4j Connection.
+ * <p>
+ * This is inspired by {@code org.hibernate.resource.transaction.backend.jdbc.internal.JdbcResourceLocalTransactionCoordinatorImp}.
```
gsmet: Imp -> Impl in the class name.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/transaction/impl/RemoteResourceLocalTransactionCoordinator.java
```diff
@@ -0,0 +1,339 @@
...
+ *
+ * @author Davide D'Alto
+ */
+public class RemoteResourceLocalTransactionCoordinator implements TransactionCoordinator {
```
gsmet: Maybe we could avoid a good part of the duplication between this and Neo4jResourceLocalTransactionCoordinator?

core/src/test/java/org/hibernate/ogm/backendtck/associations/collection/manytomany/ManyToManyTest.java
gsmet: See here:
https://github.com/hibernate/hibernate-ogm/pull/687/commits/090fbd951a86bc485cb0683e9ba794ac8917b70b#diff-0d9a048ba0f413c22966f840a46688a1L75
and here:
https://github.com/hibernate/hibernate-ogm/pull/687/commits/090fbd951a86bc485cb0683e9ba794ac8917b70b#diff-ca96314f78e37673af07f179b8c1f69eL138
or here:
https://github.com/hibernate/hibernate-ogm/pull/687/commits/090fbd951a86bc485cb0683e9ba794ac8917b70b#diff-c680f98d971b14f510ea306001b55d4aL120

Just a matter of consistency.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/DatabaseIdentifier.java
gsmet: I created a JIRA for this.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/Neo4jClient.java
gsmet: I created a JIRA for this.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/Neo4jClient.java
gsmet: If it's not possible, we might as well remove this method and the Rest interface.

gsmet:
OK, this is by far my largest review so far.

Overall, nice work, it does the job and it was quite easy to review the split commits.

2 general remarks:
- I wonder if we could avoid some of the duplication we have between the 2 dialects. I put some comments where I was worried about this. If it's not feasible, we can keep it like that but I think we should try to reduce the duplication as much as possible - as you already did it in several places;
- we are not consistent about the naming:
  - sometimes, it's Neo4j for embedded and Neo4jRemote for remote
  - sometimes it's the opposite (not in the code but in the configuration)
  - a few classes have the Neo4j for embedded but are only called Remote for remote (the transaction coordinator stuff)
  - for the SchemaDefiner, we have EmbeddedNeo4jSchemaDefiner for embedded
  - sometimes, the base implementation is called BaseNeo4j* and sometimes it doesn't have a prefix at all such as EntityQueries and AssociationQueries.


-> I think we should try to be as consistent as possible and I would propose the following naming scheme:
- EmbeddedNeo4j and RemoteNeo4j prefix for all the classes specific to a a dialect
- BaseNeo4j prefix for the abstract classes which are then in 2 flavors
- Neo4j prefix if the class is common to both without any difference (I don't propose to rename the existing class, just to be consistent for the ones which are added in 2 flavors)

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/impl/RemoteNeo4jSchemaDefiner.java
gsmet: There are only 2 of them!

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jSequenceGenerator.java
```diff
@@ -0,0 +1,253 @@
...
+	 * @param increment the difference between to consecutive values in the sequence
+	 * @param initialValue the initial value of the given generator
+	 * @return the next value in a sequence
+	 */
+	public Number nextValue(IdSourceKey idSourceKey, int increment, int initialValue) {
+		String sequenceName = sequenceName( idSourceKey );
+		Statements statements = updateNextValueQuery( idSourceKey, initialValue, increment );
+		StatementsResponse statementsResponse = neo4jDb.executeQueriesInNewTransaction( statements );
+		List<StatementResult> results = statementsResponse.getResults();
+		Integer nextValue = (Integer) results.get( 1 ).getData().get( 0 ).getRow().get( 0 );
```
gsmet: This looks a bit like black magic :)

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jSequenceGenerator.java
gsmet: OK got it, thanks. Also removed a comment above asking why the creation of the sequence was different and not taking into account the initialValue.

Might be worth a comment before `return nextValue.longValue() - increment;` though!

DavideD: Sure, I will add it

On 6 Jun 2016 17:37, "Guillaume Smet" <notifications@github.com> wrote:
>
> In
neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jSequenceGenerator.java:
>
> > + * @param initialValue the initial value of the given generator
> > + * @return the next value in a sequence
> > + */
> > + public Number nextValue(IdSourceKey idSourceKey, int increment, int
initialValue) {
> > + String sequenceName = sequenceName( idSourceKey );
> > + Statements statements = updateNextValueQuery( idSourceKey,
initialValue, increment );
> > + StatementsResponse statementsResponse =
neo4jDb.executeQueriesInNewTransaction( statements );
> > + List<StatementResult> results = statementsResponse.getResults(
> );
> > + Integer nextValue = (Integer) results.get( 1 ).getData().get( 0
).getRow().get( 0 );
> > + // sequence nodes are expected to have been created up-front
> > + if ( idSourceKey.getMetadata().getType() == IdSourceType.SEQUENCE ) {
> > + if ( nextValue == null ) {
> > + throw new HibernateException( "Sequence missing: " + sequenceName );
> > + }
> > + }
> > + return nextValue.longValue() - increment;
>
> OK got it, thanks. Also removed a comment above asking why the creation
of the sequence was different and not taking into account the initialValue.
>
> Might be worth a comment before return nextValue.longValue() - increment;
though!

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/transaction/impl/RemoteJtaTransactionCoordinator.java
DavideD: Yes, in the case of remote I need the transaction id to execute Rest API calls from the dialect

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jEntityQueries.java
DavideD: In a single request you can execute multiple queries and each query result can be returned as rows of a  table or as graph elements. All this make things a bit complicated

DavideD:
I've applied all the remarks:
* Fix typos (Hopefully I haven't added any new one)
* Added credential authentication when the client is created
* Added more comments around
* Remove TODO: Throw exception in case of error during sequences or constraints creation
* Change AvailableDatastoreProvider.NEO4J to AvailableDatastoreProvider.NEO4J_REMOTE
* Add the prefix RemoteNeo4j to cthe classes related to the remote Neo4j integration.

As for renaming the other classes I will create a separate JIRA for it since I was also thinking about having different packages for remote and embedded and this will have an impact on many files.

I will create a separate issue for adding back the QueryContext as well

DavideD:
> As for renaming the other classes I will create a separate JIRA for it since I was also thinking about having different packages for remote and embedded and this will have an impact on many files.

It might also be cause of discussion, so I would like to see this PR integrated first

DavideD:
> I wonder if we could avoid some of the duplication we have between the 2 dialects. I put some comments where I was worried about this. If it's not feasible, we can keep it like that but I think we should try to reduce the duplication as much as possible - as you already did it in several places;

I think as a first PR this is clean enough really, we can continue refactoring while working on other things. I think it will be easier to do it once I've renamed the existing classes and created some additional packages.

gsmet:
Superseeded by https://github.com/hibernate/hibernate-ogm/pull/690 and merged!

PR #688, opened 2016-06-06
OGM-1072 Update Neo4j to version 2.3.4

DavideD:
https://hibernate.atlassian.net/browse/OGM-1072

A simle one before upgrading to 3

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -27,7 +27,7 @@
         <infinispanVersion>8.2.1.Final</infinispanVersion>
         <mongodbVersion>3.2.2</mongodbVersion>
         <fongodbVersion>2.0.6</fongodbVersion>
-        <neo4jVersion>2.3.2</neo4jVersion>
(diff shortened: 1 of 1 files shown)


Sanne:
merged

Sanne:
@DavideD I see the JIRA issue is marking this for version 5.0.1 .
Does it need backporting, or did we not create a 5.0 maintenance branch yet?

DavideD:
We haven't created a maintainance branch yet

gunnarmorling:
For now I'd just merge to master. I don't think a 5.0.x branch is really needed at this point. Should we really wish to do a 5.0.1 (found a critical bug etc.) then we can discuss which issues to backport. So I'd also unset the 5.0.1 fix version from this issue until we really decided that it should go into 5.0.1.

DavideD:
Sounds good to me, I've updated the fixVersion field

PR #689, opened 2016-06-08
OGM-910 Support index creation for MongoDB

gsmet:
https://hibernate.atlassian.net/browse/OGM-910

Can we agree on the implementation before I write the documentation?

Thanks!


DavideD:
Jenkings, retest this, please

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/impl/MongoDBIndexOptions.java
```diff
@@ -0,0 +1,110 @@
...
+	/**
+	 * Specifies a value, in seconds, as a TTL to control
+	 * how long MongoDB retains documents in this collection
+	 */
+	private int expireAfterSeconds;
+
+	/**
+	 * Allows to define options for the storage engine.
+	 *
+	 * Must be a valid MongoDB document.
```
DavideD: Do you mean storage engine instead of document?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/logging/impl/Log.java
```diff
@@ -101,4 +102,28 @@
 
 	@Message(id = 1228, value = "Query must be executed using 'getResultList()' or 'getSingleResult()' method: %s")
 	HibernateException readQueryMustBeExecutedViaGetResultList(MongoDBQueryDescriptor queryDescriptor);
+
+	@Message(id = 1229, value = "Unable to create index %2$s on collection %1$s")
+	HibernateException unableToCreateIndex(String collection, String indexName, @Cause Exception e);
+
+	@LogMessage(level = ERROR)
+	@Message(id = 1230, value = "Unable to create text index %2$s on collection %1$s. A text index named %3$s already exists and MongoDB only supports one text index per collection.")
+	void unableToCreateTextIndex(String collection, String newIndexName, String existingIndexName);
```
DavideD: I don't think this should be an error, at most a warning.

DavideD: And if it's an error, wouldn't be better to throw an exception?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/impl/MongoDBIndexOptions.java
gsmet: No. It has to be a valid MongoDB document e.g. ~ a JSON document (but with some extension). It also has to be valid as a storage engine option document but this cannot be checked by us.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/logging/impl/Log.java
gsmet: I chose an error as the index cannot be created so it's definitely an error, not just a warn.

After giving it some thoughts, I think you're right and it should be an exception.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/index/MongoDBIndexSpec.java
```diff
@@ -0,0 +1,166 @@
...
+		this.unique = true;
+		this.addIndexKeys( uniqueKey.getColumnIterator(), uniqueKey.getColumnOrderMap() );
+	}
+
+	public MongoDBIndexSpec(Index index, MongoDBIndexOptions options) {
+		this.options = options;
+		this.collection = index.getTable().getName();
+		this.indexName = index.getName();
+		this.unique = false;
+		// TODO HHH-10760: the columnOrderMap is not accessible for an Index
```
DavideD: Do we have a JIRA t okeep track of this? I think I've seen another TODO referring to it

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/logging/impl/Log.java
DavideD: From the exception it seems that the index already exists in the DB, so it's not that it is not created. In Neo4jSchemaDefiner we use the property: hibernate.schema_update.unique_constraint_strategy to let the user decide what to do in this situations.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1234,6 +1239,60 @@ public ParameterMetadataBuilder getParameterMetadataBuilder() {
...
+
+		try {
+			// if the index is already present and with the same definition, MongoDB simply ignores the call
+			// if the definition is not the same, MongoDB throws an error, except in the case of a text index
+			// where it silently ignores the creation
+			collection.createIndex( indexSpec.getIndexKeysDBObject(), indexSpec.getIndexOptionsDBObject() );
+		}
+		catch (MongoException e) {
+			String indexName = indexSpec.getIndexName();
+			if ( e.getCode() == 85
```
DavideD: What's `85`?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/logging/impl/Log.java
gsmet: **A** text index already exists in the db for this collection. It might be a totally different index or an index with the same name but a different configuration. The issue is that only one text index per collection is supported and MongoDB silently ignores all the text index creation orders if there is already one for the collection.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/index/MongoDBIndexSpec.java
gsmet: Yup, I'll create the according JIRA once this gets in.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
gsmet: It's the MongoDB error code. It's not very precise as it seems it's the same for every error directly linked to the index creation.

DavideD: I usually use a constant with a descriptive name for these cases, mainly to let people know what kind of error the code represents

DavideD:
I will just add a summary of the conversation I had with @gsmet about this:

* Keep the GridDIalect and the SchemaDefiner separated. No need to call the dialect to apply these changes.
* We should add a property to skip the unique constraint creation, like for  Neo4j
* Add a test to check that the constraint is applied when a column is unique (via attribute property or annotation). See the tests for Neo4j UniqueConstraint*Test

core/src/main/java/org/hibernate/ogm/dialect/spi/GridDialect.java
```diff
@@ -8,6 +8,7 @@
 
 import org.hibernate.LockMode;
 import org.hibernate.dialect.lock.LockingStrategy;
+import org.hibernate.ogm.index.OgmIndexSpec;
```
gunnarmorling: Should this be in "spi"?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/MongoDBCollection.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * MongoDB specific definitions for a collection.
+ *
+ * @author Guillaume Smet
+ */
+@Target({ TYPE, METHOD, FIELD })
+@Retention(RUNTIME)
+@MappingOption(MongoDBCollectionConverter.class)
+public @interface MongoDBCollection {
```
gunnarmorling: I'd prefer a simple "list annotation", i.e. with the single element named `value()` and it's named after the element it contains. But I see how that collides with the element name which already is pluralized.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/index/MongoDBIndexSpec.java
gsmet: Done: https://hibernate.atlassian.net/browse/OGM-1080 . I referenced it in the TODO.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
gsmet: Done.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/MongoDBCollection.java
gsmet: The idea was to centralize here all the options specific to MongoDB which might be applied to the collection, instead of creating multiple options.

core/src/main/java/org/hibernate/ogm/dialect/spi/GridDialect.java
gsmet: It has been removed in my branch as the createIndex method has been moved to the SchemaDefiner.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/MongoDBCollection.java
gunnarmorling: What other options would that be? Me feeling is we should rather find options that are applicable to several stores. E.g. take `@TimeToLife`. Currently it's only there fore Redis, but it could be generalized and used for MongoDB, too. In your model `timeToLife()`, it could be a part of `@MongoDBCollection`, but I think it shouldn't so to provide a more unified experience when working with different stores. See where I'm coming from?

gsmet: Any option from https://docs.mongodb.com/manual/reference/method/db.createCollection/ ?

I find it far more interesting to have a fully documented vision of the options supported by my datastore, rather than having a bunch of annotations which might be or might not be supported depending on the datastore and having to check in the doc if they are or not.

Typically, it's far easier to migrate from one datastore to another, you just change the master annotation and you see which options are supported or not. Much better than letting your annotations there and having them ignored because they are not supported.

gsmet:
@DavideD I applied all the changes we discussed.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/MongoDBCollection.java
gunnarmorling: We can discuss that, but it'd be quite a shift from the principles OGM follows so far. We are looking to provide a unified programming model, finding useful abstractions as far as possible. What you describe essentially is more to have an "OGM for MongoDB", an "OGM for Cassandra" etc.

Also there is the case of polyglot persistence where an entity is mapped to several stores. Being able to re-use config here would be nice instead of having `@MongoCollection`, `@CassandraCollection` etc. all there.

That said, I see your point about being unsure of what's supported on a given store or not. But that's a general issue, well, challenge ;), with OGM.

@emmanuelbernard, what's your take?

core/src/main/java/org/hibernate/ogm/service/impl/DefaultSchemaInitializationContext.java
```diff
@@ -84,6 +86,19 @@ public Database getDatabase() {
 		return allIdSourceKeyMetadata;
 	}
 
+	@Override
+	public Map<String, Class<?>> getTableEntityTypeMapping() {
+		Map<String, Class<?>> mapping = new HashMap<>();
+		Map<String, EntityPersister> entityPersisters = factory.getEntityPersisters();
+
+		for ( Entry<String, EntityPersister> entityPersisterEntry : entityPersisters.entrySet() ) {
```
gunnarmorling: A table might contain several entities (inheritance). Wondering whether that may be a problem here?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -381,7 +384,14 @@ public void insertOrUpdateTuple(EntityKey key, Tuple tuple, TupleContext tupleCo
 			getCollection( key ).update( idObject, updater, true, false, writeConcern );
 		}
 		catch ( DuplicateKeyException dke ) {
-			throw new TupleAlreadyExistsException( key.getMetadata(), tuple, dke );
+			// This exception is used by MongoDB for all the unique indexes violation, not only the primary key
+			// so we determine if it concerns the primary key by matching on the message
+			if ( PRIMARY_KEY_CONSTRAINT_VIOLATION_MESSAGE.matcher( dke.getMessage() ).matches() ) {
```
gunnarmorling: Doh :/

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBSchemaDefiner.java
```diff
@@ -0,0 +1,350 @@
...
+			// where it silently ignores the creation
+			collection.createIndex( indexSpec.getIndexKeysDBObject(), indexSpec.getIndexOptionsDBObject() );
+		}
+		catch (MongoException e) {
+			String indexName = indexSpec.getIndexName();
+			if ( e.getCode() == INDEX_CREATION_ERROR_CODE
+					&& !StringHelper.isNullOrEmptyString( indexName )
+					&& preexistingIndexes.containsKey( indexName ) ) {
+				// The index already exists with a different definition and has a name: we drop it and we recreate it
+				collection.dropIndex( indexName );
```
gunnarmorling: How does ORM behave in this case? Is it also doing this?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/index/MongoDBIndexSpec.java
```diff
@@ -0,0 +1,182 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.datastore.mongodb.index;
```
gunnarmorling: How comes this is in the API? I don't think we expect the user to interact with that. If so, it should go somewhere into _impl_.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/MongoDBCollection.java
emmanuelbernard: These options as described by https://docs.mongodb.com/manual/reference/method/db.createCollection/ are all about "DDL" generation tuning and very very specific to MongoDB. Most are actually specific to new versions of MongoDB.

For the sake of simplicity, I would even go as far as offering a full bypass.

    @MongoDBCollectionOptions(name="customerCollection", options="{ capped : true, size : 5242880, max : 5000, storageEngine: { wiredTiger: { configString: \"<option>=<setting>\" } } }")

While we should try and align on runtime things like the target structure and maybe even whether a property is indexed or not, I don't think the DDL options are a worth it. To be honest even Index will be hard. For example Infinispan indexes are infinitely complex as we use the full power of Hibernate Search.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/MongoDBIndexOptions.java
```diff
@@ -0,0 +1,72 @@
...
+	 * Builds the index in the background so that building an index does not block other database activities.
+	 */
+	boolean background() default false;
+
+	/**
+	 * Allows to define a partial index by providing a filter expression.
+	 *
+	 * Must be a valid MongoDB document.
+	 */
+	String partialFilterExpression() default "";
```
gunnarmorling: Wow, this goes way beyond what we had in originally in mind. At least I just had thought about enabling usage of the existing JPA annotations for index creation. It ties into the discussion above really. I'm not sure whether we should expose all this.

After all, a user could create the "schema" manually themselves, using the native APIs. Actually that's what I'd expect to happen in practice anyways, based on my experiences with RDMBS. You always wanted to set some DB-specific magic switches in the DDL, never available through a generic tool such as Hibernate.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/MongoDBCollection.java
gsmet: IMHO, it would be an error to not expose the features of the underlying datastore in the most useful way. It's not about an "OGM for MongoDB", it's about OGM being useful. Each datastore has specific features and that's why people use them. It would be an error to think they are interchangeable like a standard SQL store would be (and even that is not entirely true, I was very annoyed by the fact that it was not possible to define partial indexes with JPA annotations).

Typically, the point of having Neo4j is not to store my objects in it as I would do it in MongoDB, it's about being able to leverage the features of a Graph database.

IMHO, part of our job is to do all the work to expose the interesting features of each datastore to the world using a common programming model.

BTW, you didn't answer to the point that we will probably need @MongoDBCollection to expose the options of a collection. So why not also store the index options specific to MongoDB here?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/impl/MongoDBCollection.java
```diff
@@ -0,0 +1,49 @@
...
+		}
+		if ( !indexOptions.containsKey( indexName ) ) {
+			return new MongoDBIndexOptions( indexName );
+		}
+
+		return indexOptions.get( indexName );
+	}
+
+	public Set<String> getReferencedIndexes() {
+		return Collections.unmodifiableSet( indexOptions.keySet() );
```
gunnarmorling: We could make the field immutable itself? There is also a "tagging" annotation for it somewhere.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/impl/MongoDBCollectionOption.java
```diff
@@ -0,0 +1,19 @@
...
+package org.hibernate.ogm.datastore.mongodb.options.impl;
+
+import org.hibernate.ogm.options.spi.UniqueOption;
+
+/**
+ * Option to specify MongoDB specific options for a collection.
+ *
+ * @author Guillaume Smet
+ */
+public class MongoDBCollectionOption extends UniqueOption<MongoDBCollection> {
```
gunnarmorling: No contents?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/options/impl/MongoDBIndexOptions.java
gunnarmorling: So one wouldn't specify something like "wiredTiger" but rather "{ \"storageEngine\" : \"wiredTiger\" }"? An example might be useful.

gsmet:
Closing for now.

PR #694, opened 2016-06-14
OGM-1084 In the case of an association stored in the entity, the TTL should be the one of the entity

gsmet:
, not the one of the association

Clarify this in the documentation.

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/dialect/spi/AssociationTypeContext.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/spi/AssociationTypeContext.java
@@ -20,11 +20,18 @@ import org.hibernate.ogm.options.spi.OptionsContext;
 public interface AssociationTypeContext {
 
 	/**
-	 * Provide access to the options set for an entity or property.
+	 * Provide access to the options set for the association.
 	 *
-	 * @return a context object providing access to the options effectively applying for a given entity or property.
+	 * @return a context object providing access to the options effectively applying for the association.
 	 */
 	OptionsContext getOptionsContext();
+
+	/**
+	 * Provide access to the options set for the entity owner of the association.
+	 *
+	 * @return a context object providing access to the options effectively applying for the owner entity.
+	 */
+	OptionsContext getOwnerEntityOptionsContext();
 
 	/**
 	 * Provides meta-data about the entity key on the other side of this association.
--- a/documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
+++ b/documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
@@ -141,6 +141,12 @@ public class Zoo {
 ----
 ====
 
+The annotation on the entity level expresses that all associations of the `Zoo`
+class should be stored in separate association documents.
+This setting applies to the `animals` and `employees` associations.
+Only the elements of the `visitors` association will be stored in the document of the corresponding `Zoo` entity
+as per the configuration of that specific property which takes precedence over the entity-level configuration.
+
 .Configuring the TTL/expiry using annotations
 ====
 [source, JAVA]
@@ -153,7 +159,12 @@ public class Zoo {
     private Set<Animal> animals;
 
     @OneToMany
+    @TTL(value = 3, unit = TimeUnit.DAYS)
     private Set<Person> employees;
+
+    @OneToMany
+    @AssociationStorage(AssociationStorageType.IN_ENTITY)
+    private Set<Person> visitors;
 
     //...
 }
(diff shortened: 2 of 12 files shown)


core/src/main/java/org/hibernate/ogm/dialect/spi/AssociationTypeContext.java
```diff
@@ -20,13 +20,20 @@
...
+	 * @return a context object providing access to the options effectively applying for the association.
 	 */
 	OptionsContext getOptionsContext();
 
 	/**
+	 * Provide access to the options set for the entity owner of the association.
+	 *
+	 * @return a context object providing access to the options effectively applying for the owner entity.
+	 */
+	OptionsContext getOwnerEntityOptionsContext();
```
gunnarmorling: Could this info not be exposed via the existing `getOptionsContext()`. The idea originally was that this exposes the effectively applying options for a given element. My concern is that the two separate methods further complicate things for dialect implementations.

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
```diff
@@ -165,11 +176,8 @@ Hibernate OGM allows to specify a TTL value on entities and associations. The TT
 the entity using the http://redis.io/commands/pexpire[PEXPIRE] command. Every write to Redis will
 set a new TTL.
 
-The annotation on the entity level expresses that all associations of the `Zoo`
-class should be stored in separate assocation documents.
-This setting applies to the `animals` and `employees` associations.
-Only the elements of the `visitors` association will be stored in the document of the corresponding `Zoo` entity
-as per the configuration of that specific property which takes precedence over the entity-level configuration.
+You can specify a different TTL on associations if they are stored as separate documents. In the case of `visitors`
```
gunnarmorling: Should we really go down that path? Is there a user request for this? TTL makes sense for entities to me, but association elements? What's the use case behind this?

gsmet: @gunnarmorling No idea! I also find it rather convoluted but it was preexisting to this patch. I only made it more consistent.

core/src/main/java/org/hibernate/ogm/dialect/spi/AssociationTypeContext.java
gsmet: The fact is that the decision to use one or the other is taken in the dialect depending on if the association is stored in the entity or not. And it might not be the same decision for each options. Typically, the case of TTL is quite special.

gunnarmorling: > The fact is that the decision to use one or the other is taken in the
dialect depending on if the association is stored in the entity or not

Yes, understood. What I meant is, can we somehow generalize that, though? I
suppose it's tricky as "in-entity" vs. "separate" is not a concept in core,
IIRC.

Sanne: > What I meant is, can we somehow generalize that, though?

That would be interesting as e.g. Infinispan also supports TTL, and I assume at least Ehcache would have it too.

But let's define the use case clearly first, and assign it a (low?) priority accordingly.

Let's also consider that both Ehcache and Infinispan can configure a TTL for the whole storage (the Cache, an of course support other smarter forms of eviction too), so supporting such a feature for Hibernate OGM is only interesting for use cases which require a per-instance override of the global eviction settings.

documentation/manual/src/main/asciidoc/en-US/modules/redis.asciidoc
gsmet: @gunnarmorling Found one during my sports session :). It might be useful if you consider a list of users visiting a page or a product, or the list of active users on a forum topic.

Redis is typically used for this sort of transient information.

core/src/main/java/org/hibernate/ogm/dialect/spi/AssociationTypeContext.java
gsmet: > What I meant is, can we somehow generalize that, though?

I don't think we can generalize it considering the behavior is going to be different depending on the option we consider.

Typically, you don't want to use the entity options context for options like AssociationStorage.

gsmet: > But let's define the use case clearly first, and assign it a (low?) priority accordingly.

Yes, I don't think it has a lot to do with this commit. This commit just fixes an existing buggy behavior which can have quite undesirable consequences.

gsmet:
@gunnarmorling any further thoughts on this?

mp911de:
Nice addition.

gsmet:
Rebased.

Could we commit this? It's sitting there for more than a month. Removing this feature might be a discussion for later but this feature is currently advertised as supported and I'd really like to get it fixed.

Thanks!

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/694

As you said, we can always change it later

PR #695, opened 2016-06-14
OGM-1023 Add a test and documentation about the behavior of the $regex operator

gsmet:


documentation/manual/src/main/asciidoc/en-US/modules/mongodb.asciidoc
```diff
@@ -3329,6 +3329,12 @@ int removed = session.createNativeQuery( removeQuery ).executeUpdate();
 ----
 ====
 
+[WARNING]
+====
+Support for the `$regexp` operator is limited to the string syntax. We do not support the `/pattern/` syntax as it is not
```
DavideD: Personally, I think an example would be much more effective. Are there any limitation in using one syntax compare to the other?

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/695

PR #696, opened 2016-06-14
OGM-1084 Add script to run docker containers for OGM databases

emmanuelbernard:


DavideD:
@emmanuelbernard Any particular reason to put his in the bin folder?

Under `bin` I have all the compiled classes form eclipse (I think). I would move it under `src/main/scripts` that's the place where we keep all our scripts

gunnarmorling:
@DavideD, `bin` makes sense as it's a commonly used pattern for keeping executables and scripts. Still +1 for putting this script next to the other ones.

DavideD:
Closing https://github.com/hibernate/hibernate-ogm/pull/696

I will move it and send a different PR

emmanuelbernard:
Sorry, I'm from an era where scripts and other binaries were easily accessible in `bin` and not 3 level deep in a normalized structure ;)

PR #699, opened 2016-06-15
OGM-910 Make the index tests work on Fongo and relax the regexp

gsmet:
identifying a constraint violation on the primary key

Note: I removed the dropIndexes call as it's useless and counter productive when we have several tests.


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/699

By the way, we usually separate the title from the body in commit messages. As ecplained here: g

Except for that, I don't think we have strict rules, but I follow the guidelines in the article as often as I can.

PR #705, opened 2016-06-17
OGM-1082 Include the exception message in the log when we batch several

gsmet:
entity insertions and we end up with a TupleAlreadyExistsException.

It's not perfect but it's far better than having the following message:
    Trying to insert an already existing entity: null
which is quite misleading.


Sanne:
merged

PR #706, opened 2016-06-17
OGM-1090 Use Central first to get the artifacts

gsmet:
https://hibernate.atlassian.net/browse/OGM-1090

In passing, remove the sed calls adding this for the Travis build.

Note, I'm using the default Central repository instead of the Google
mirror as apparently a few artifacts are missing on Google mirror.

The default Central repository is now hosted by Fastly and is very fast.


Sanne:
merged

PR #708, opened 2016-06-21
JDK9

Sanne:
This allows us to run the core testsuite on Java 9 and prevent regressions with our own code, but still needs some investigation to get the actual NoSQL technologies to work.. although that might not be our responsibility.

There are still issues with:
 - running on WildFly (WildFly isn't ready for it)
 - Ehcache - we're using version 2, I doubt the Ehcache team will want to fix this unless we upgrade to 3
 - Infinispan - needs further investigation from the Infinispan team
 - Neo4J - some banned NIO usage, might be simple?

Not least this disables distribution and documentation build.

https://hibernate.atlassian.net/browse/OGM-1095

### DIFF
--- a/cassandra/pom.xml
+++ b/cassandra/pom.xml
@@ -15,6 +15,10 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>org.bsc.maven</groupId>
+                <artifactId>maven-processor-plugin</artifactId>
+            </plugin>
+            <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -207,6 +207,8 @@
                     <source>1.7</source>
                     <target>1.7</target>
                     <encoding>UTF-8</encoding>
+                    <!-- Annotation processor is run as an independent step -->
+                    <proc>none</proc>
                 </configuration>
             </plugin>
             <plugin>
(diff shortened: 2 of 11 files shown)


pom.xml
```diff
@@ -669,12 +692,36 @@
 
     <profiles>
         <profile>
+            <id>non-jigsaw</id>
+            <activation>
+                <!-- The versions we're using of Ehcache, Infinispan and Neo4J won't work on JDK9. Needs upgrading.  -->
+                <jdk>[1.7,1.8]</jdk>
+            </activation>
+            <modules>
+                <module>ehcache</module>
```
gunnarmorling: Se we can't run them at all? Couldn't we exclude single tests? Or is it more global issues with them, causing all tests to fail?

pom.xml
```diff
@@ -669,12 +692,36 @@
...
+        </profile>
+        <profile>
             <id>integrationtest</id>
             <activation>
                 <property>
                     <name>skipITs</name>
                     <value>!true</value>
                 </property>
+                <!-- Skip on Java 9, both because of some NoSQL stores aren't ready and because of WildFly -->
+                <jdk>[1.7,1.8]</jdk>
```
gunnarmorling: We require 1.8 anyways, so you can simplify that a bit.

cassandra/pom.xml
```diff
@@ -15,6 +15,10 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>org.bsc.maven</groupId>
```
gunnarmorling: Interesting, we seem to oscillate between adding and removing that plug-in :)

pom.xml
Sanne: Regarding Infinispan and Neo4J they have global issues.

For ehcache we could exclude a single test but it doesn't give us much value: either way, we can't say it will work on Java9 so I think the better approach is to defer this to an Ehcache upgrade, or a full investigation of why/what Ehcache is going to do with Java9.

In general I'd prefer to merge this as "the core builds", and open follow-up investigations so we can break it up in smaller tasks.

Sanne:
@gunnarmorling re-pushed to change the JDK check on Java 8 only like you suggested (github seems to have lost your comment?)

gunnarmorling:
@Sanne any idea about the test failure on CI? Not related? Looks good otherwise. Only let's open a follow-up issue to investigate the remaining modules not working on JDK 9 yet.

Sanne:
I don't think that CI failure is related, but running some double-checks with both JDKs locally.
I'll push it if it works..

+1 to open follow up tasks, I've started such talks with the Infinispan team so I guess that one is mine.

Sanne:
merged, took me a while as I had some local issues with Cassandra.

PR #711, opened 2016-06-23
OGM-1103 Fix concurrency issue in EventContextManager

gsmet:
https://hibernate.atlassian.net/browse/OGM-1103

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/compensation/impl/InvocationCollectingGridDialect.java
+++ b/core/src/main/java/org/hibernate/ogm/compensation/impl/InvocationCollectingGridDialect.java
@@ -60,10 +60,15 @@ import org.hibernate.ogm.model.spi.Tuple;
  */
 public class InvocationCollectingGridDialect extends ForwardingGridDialect<Serializable> {
 
-	private final EventContextManager eventContext;
+	private EventContextManager eventContext;
 
-	public InvocationCollectingGridDialect(GridDialect gridDialect, EventContextManager eventContextManager) {
+	public InvocationCollectingGridDialect(GridDialect gridDialect) {
 		super( gridDialect );
+	}
+
+	@Override
+	public void setEventContextManager(EventContextManager eventContextManager) {
+		super.setEventContextManager( eventContextManager );
 		this.eventContext = eventContextManager;
 	}
 
--- a/core/src/main/java/org/hibernate/ogm/compensation/impl/InvocationCollectingGridDialect.java
+++ b/core/src/main/java/org/hibernate/ogm/compensation/impl/InvocationCollectingGridDialect.java
@@ -60,7 +60,7 @@ import org.hibernate.ogm.model.spi.Tuple;
  */
 public class InvocationCollectingGridDialect extends ForwardingGridDialect<Serializable> {
 
-	private EventContextManager eventContext;
+	private EventContextManager eventContextManager;
 
 	public InvocationCollectingGridDialect(GridDialect gridDialect) {
 		super( gridDialect );
@@ -69,7 +69,7 @@ public class InvocationCollectingGridDialect extends ForwardingGridDialect<Seria
 	@Override
 	public void setEventContextManager(EventContextManager eventContextManager) {
 		super.setEventContextManager( eventContextManager );
-		this.eventContext = eventContextManager;
+		this.eventContextManager = eventContextManager;
 	}
 
 	@Override
@@ -311,6 +311,6 @@ public class InvocationCollectingGridDialect extends ForwardingGridDialect<Seria
 	 * Returns the {@link OperationCollector}. Must not store this as a field as it is flush-cycle scoped!
 	 */
 	private OperationCollector getOperationCollector() {
-		return eventContext.get( OperationCollector.class );
+		return eventContextManager.get( OperationCollector.class );
 	}
 }
--- a/core/src/main/java/org/hibernate/ogm/dialect/eventstate/impl/EventContextManager.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/eventstate/impl/EventContextManager.java
@@ -11,15 +11,20 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 
+import org.hibernate.engine.config.spi.ConfigurationService;
 import org.hibernate.engine.spi.SessionImplementor;
 import org.hibernate.event.spi.EventSource;
 import org.hibernate.ogm.cfg.OgmProperties;
+import org.hibernate.ogm.compensation.impl.OperationCollector;
+import org.hibernate.ogm.dialect.batch.spi.OperationsQueue;
+import org.hibernate.ogm.dialect.eventstate.impl.EventStateLifecycles.OperationCollectorLifecycle;
+import org.hibernate.ogm.dialect.eventstate.impl.EventStateLifecycles.OperationsQueueLifecycle;
 import org.hibernate.ogm.dialect.impl.BatchOperationsDelegator;
 import org.hibernate.ogm.dialect.impl.GridDialects;
 import org.hibernate.ogm.dialect.spi.GridDialect;
 import org.hibernate.ogm.util.impl.Immutable;
 import org.hibernate.service.Service;
-import org.hibernate.service.spi.SessionFactoryServiceRegistry;
+import org.hibernate.service.spi.ServiceRegistryImplementor;
 
 /**
  * A service which provides access to state specific to one event cycle (currently (auto)-flush or persist).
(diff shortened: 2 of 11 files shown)


core/src/main/java/org/hibernate/ogm/dialect/eventstate/impl/EventContextManager.java
```diff
@@ -72,7 +72,9 @@ void onEventFinished() {
 
 		SessionImplementor session = (SessionImplementor) states.get( SessionImplementor.class );
 
-		for ( Entry<Class<?>, Object> state : states.entrySet() ) {
+		// we need to make a copy of the map as an OperationCollector entry might be added during
```
gunnarmorling: In which case does this exactly happen? I'm no fan of this right now, because - as you say below - the onFinish() will not be called for the newly added state. Seems like a potential cause for trouble.

gsmet: Yeah, I'm not a fan either. I pushed it to have your feedback.

When we use InvocationCollectingGridDialect, we add things to an `OperationCollector` which is stored in the states. When grouping, we might do it in the `OperationsQueue` onFinish call and if the `OperationCollector` is not already present in the state, we add it to the states map.

gsmet: A different approach would be to register all the necessary elements in the `onEventBegin` call. We could add an `onBegin` method to `EventStateLifecycle` and remove the create method.

Might be a good opportunity to only register the lifecycle if required by the dialect.

core/src/main/java/org/hibernate/ogm/compensation/impl/InvocationCollectingGridDialect.java
```diff
@@ -60,10 +60,15 @@
  */
 public class InvocationCollectingGridDialect extends ForwardingGridDialect<Serializable> {
 
-	private final EventContextManager eventContext;
```
gunnarmorling: Each time you remove a `final`, a kitten dies, cruely.

core/src/main/java/org/hibernate/ogm/dialect/eventstate/impl/EventContextManagerInitiator.java
```diff
@@ -6,18 +6,19 @@
...
+import org.hibernate.service.spi.SessionFactoryServiceInitiator;
 
 /**
  * Contributes the {@link EventContextManager} service.
  *
  * @author Gunnar Morling
  */
-@SuppressWarnings("rawtypes")
-public class EventContextManagerInitiator implements StandardServiceInitiator<EventContextManager> {
+public class EventContextManagerInitiator implements SessionFactoryServiceInitiator<EventContextManager> {
```
gunnarmorling: Why is this SF-specific now?

core/src/main/java/org/hibernate/ogm/compensation/impl/InvocationCollectingGridDialect.java
gsmet: I was expecting this! I'm allergic to cats.

More seriously, any idea on how to fix it without injecting the eventContext later?

gsmet: Btw, I added 2 finals below so I probably resurrected 2 kittens.

core/src/main/java/org/hibernate/ogm/dialect/eventstate/impl/EventContextManagerInitiator.java
gsmet: Mmmh, probably because I thought I would need a SessionFactoryServiceRegistry as the old isEventContextRequired method was using it.

But, it looks like it's useless.

If you're OK with the general approach, I'll withdraw that part.

core/src/main/java/org/hibernate/ogm/service/impl/OgmIntegrator.java
```diff
@@ -60,7 +60,7 @@ private void doIntegrate(Metadata metadata, SessionFactoryImplementor sessionFac
 	private void attachEventContextManagingListenersIfRequired(SessionFactoryServiceRegistry serviceRegistry) {
 		@SuppressWarnings("unchecked")
 		Map<Object, Object> settings = serviceRegistry.getService( ConfigurationService.class ).getSettings();
-		if ( !EventContextManager.isEventContextRequired( settings, serviceRegistry ) ) {
+		if ( !EventContextManager.isEventContextRequired( serviceRegistry ) ) {
```
gunnarmorling: Seems passing `settings` wasn't needed before. Good catch.

core/src/main/java/org/hibernate/ogm/dialect/eventstate/impl/EventContextManagerInitiator.java
```diff
@@ -30,7 +31,16 @@ private EventContextManagerInitiator() {
...
 	@Override
-	public EventContextManager initiateService(Map configurationValues, ServiceRegistryImplementor registry) {
-		return new EventContextManager();
+	public EventContextManager initiateService(SessionFactoryImplementor sessionFactory, SessionFactoryOptions sessionFactoryOptions,
+			ServiceRegistryImplementor registry) {
+		EventContextManager eventContextManager = new EventContextManager( registry );
+
+		GridDialect gridDialect = registry.getService( GridDialect.class );
+		// we don't use GridDialects.hasFacet here as we need to set it even for a {@link ForwardingGridDialect}
+		if ( gridDialect instanceof EventContextManagerAwareGridDialect ) {
```
gunnarmorling: Could you use `getDialectFacetOrNull()`?

core/src/main/java/org/hibernate/ogm/dialect/impl/GridDialectInitiator.java
```diff
@@ -50,27 +49,24 @@ public GridDialect initiateService(Map configurationValues, ServiceRegistryImple
...
 				.getValue();
 	}
 
 	private static class GridDialectInstantiator implements Instantiator<GridDialect> {
 
 		private final DatastoreProvider datastore;
 		private final boolean errorHandlerConfigured;
-		private final EventContextManager eventContext;
 
-		public GridDialectInstantiator(DatastoreProvider datastore, boolean errorHandlerConfigured, EventContextManager eventContext) {
```
gunnarmorling: Why is it we cannot do it like this anymore?

core/src/main/java/org/hibernate/ogm/dialect/eventstate/impl/EventContextManagerInitiator.java
gsmet: Nope, we wan't use the hasFacet machinery as it's not what we want. We also want to inject it in the ForwardingGridDialect if the underlying dialect is not a EventContextManagerAwareGridDialect. See the comment above.

core/src/main/java/org/hibernate/ogm/dialect/impl/GridDialectInitiator.java
gsmet: We can't because there is a dependency between the GridDialect service and the EventContextManager service.

See EventContextManager#isOperationsQueueRequired.

core/src/main/java/org/hibernate/ogm/dialect/eventstate/impl/EventContextManagerInitiator.java
gunnarmorling: Why is this though? Seems `ForwardingGridDialect` is only passing it on if the wrapped dialect actually is a `EventContextManagerAwareGridDialect`?

gsmet: BatchOperationsDelegator is a ForwardingGridDialect and it needs the eventContextManager. See https://github.com/hibernate/hibernate-ogm/pull/711/files#diff-d064aa3d9da51990e6a0d80d519b9b6aR49 .

Same for InvocationCollectingGridDialect.

gsmet:
@gunnarmorling I force pushed a new version to withdraw the changes to the initialization (making the initialization session factory dependent)

Sanne:
N.B. I changed the JIRA description from "Concurrency issue" as a ConcurrentModificationException isn't a concurrency issue, I found it confusing.

gunnarmorling:
@gsmet Do you have a test which actually exposes the original CME? I tried to reproduce it to wrap my head around it, but I see a `createTuple()` call first, which is non batchable and thus the state object is created "early", not only upon `onFinish()`.

gunnarmorling:
Ah, got something. I can trigger it via insert + flush + update + commit.

gunnarmorling:
@gsmet, created an alternative proposal here: https://github.com/hibernate/hibernate-ogm/pull/723

gsmet:
Closing as it is superseded by Gunnar's proposal.

PR #712, opened 2016-06-23
OGM-1104 Properly throw an exception in GridDialectInitiator

gsmet:
https://hibernate.atlassian.net/browse/OGM-1104

If the GridDialect does not have a valid constructor, we
ignored the error and we threw a NPE instead of an explicit error
message.

In passing, fix the error message.

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/util/impl/Log.java
+++ b/core/src/main/java/org/hibernate/ogm/util/impl/Log.java
@@ -66,7 +66,7 @@ public interface Log extends BasicLogger {
 	@Message(id = 11, value = "Cannot instantiate GridDialect class [%1$s]")
 	HibernateException cannotInstantiateGridDialect(@FormatWith(ClassObjectFormatter.class) Class<?> dialectClass, @Cause Exception e);
 
-	@Message(id = 14, value = "%1$s has no constructor accepting DatasourceProvider")
+	@Message(id = 14, value = "%1$s has no constructor accepting DatastoreProvider")
 	HibernateException gridDialectHasNoProperConstructor(@FormatWith(ClassObjectFormatter.class) Class<?> dialectClass);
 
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/hibernate/ogm/test/boot/GridDialectInitiatorTest.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+public class GridDialectInitiatorTest {
+
+	@Test
+	@TestForIssue(jiraKey = "OGM-1104")
+	public void testInvalidGridDialect() {
+		try {
+			StandardServiceRegistry registry = new StandardServiceRegistryBuilder()
+					.applySetting( OgmProperties.ENABLED, true )
+					.applySetting( OgmProperties.GRID_DIALECT, "org.hibernate.ogm.test.boot.InvalidGridDialect" )
```
gunnarmorling: You can pass the class object, too. Makes it a bit safer.

core/src/main/java/org/hibernate/ogm/util/impl/Log.java
```diff
@@ -66,7 +66,7 @@
 	@Message(id = 11, value = "Cannot instantiate GridDialect class [%1$s]")
 	HibernateException cannotInstantiateGridDialect(@FormatWith(ClassObjectFormatter.class) Class<?> dialectClass, @Cause Exception e);
 
-	@Message(id = 14, value = "%1$s has no constructor accepting DatasourceProvider")
+	@Message(id = 14, value = "%1$s has no constructor accepting DatastoreProvider")
```
gunnarmorling: I'd put the FQN here to avoid ambiguities.

gunnarmorling:
@gsmet Added two remarks; I'll fix them and then merge.

gunnarmorling:
Merged; thanks.

PR #715, opened 2016-06-27
OGM-1075 Make sure embedded are loaded correctly in Neo4j remote

DavideD:
https://hibernate.atlassian.net/browse/OGM-1075

  All embedded associated to an entity, should be loaded when the entity is loaded.
  This commit fix the previous behaviour of making a request to the server for each embedded
  column.

### DIFF
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/BaseNeo4jEntityQueries.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/BaseNeo4jEntityQueries.java
@@ -30,11 +30,16 @@ import org.hibernate.ogm.util.impl.EmbeddedHelper;
  */
 public abstract class BaseNeo4jEntityQueries extends BaseNeo4jQueries {
 
+	/**
+	 * The alias used when a query returns an entity as result.
+	 */
+	public static final String ENTITY_ALIAS = "owner";
+
 	private static final int CACHE_CAPACITY = 1000;
 	private static final int CACHE_CONCURRENCY_LEVEL = 20;
 
 	/**
-	 * true if we the keys are mapped with a single non embedded property
+	 * true if we the keys are mapped with a single property
 	 */
 	protected final boolean singlePropertyKey;
 	protected final String[] keyColumns;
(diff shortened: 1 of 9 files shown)


DavideD:
Jenkins, retest this please

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/BaseNeo4jEntityQueries.java
```diff
@@ -136,19 +150,19 @@ private String initRemoveToOneAssociation(EntityKeyMetadata entityKeyMetadata, T
 	 *
 	 * In this case the query depends on how many id we are retrieving and it will be completed later
 	 */
-	private static String initMultiGetEntitiesQuery(EntityKeyMetadata entityKeyMetadata) {
+	private static String initMultiGetEntitiesQuery(EntityKeyMetadata entityKeyMetadata, boolean includeEmbedded) {
 		StringBuilder queryBuilder = new StringBuilder( "MATCH " );
-		queryBuilder.append( "(n:" );
+		queryBuilder.append( "(owner:" );
```
gsmet: It would make sense to use the constant you declared above.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/BaseNeo4jEntityQueries.java
```diff
@@ -136,19 +150,19 @@ private String initRemoveToOneAssociation(EntityKeyMetadata entityKeyMetadata, T
...
 		queryBuilder.append( ENTITY );
 		queryBuilder.append( ":" );
 		appendLabel( entityKeyMetadata, queryBuilder );
 		queryBuilder.append( ") " );
 		queryBuilder.append( " WHERE " );
 		if ( entityKeyMetadata.getColumnNames().length == 1 ) {
-			queryBuilder.append( "n." );
+			queryBuilder.append( "owner." );
```
gsmet: Same here.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/BaseNeo4jEntityQueries.java
```diff
@@ -384,25 +398,45 @@ private static String initUpdateEmbeddedNodeQuery(EntityKeyMetadata entityKeyMet
...
+			queryBuilder.append( ", r" );
+		}
+		else {
+			queryBuilder.append( " RETURN " );
+			queryBuilder.append( ENTITY_ALIAS );
+		}
+	}
+
+	private static void appendOptionalMatchOwnerEmbeddedNodes(StringBuilder queryBuilder) {
+		queryBuilder.append( " OPTIONAL MATCH (owner) -[r*]->(:" );
```
gsmet: Same here.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/NodeWithEmbeddeds.java
```diff
@@ -0,0 +1,47 @@
...
+		this.embeddedNodes = embeddedNodes == null ? EMPTY_MAP : Collections.unmodifiableMap( embeddedNodes );
+	}
+
+	public Node getOwner() {
+		return owner;
+	}
+
+	/**
+	 * The set of Nodes associated to a specific path.
+	 * @return
```
gsmet: Could you either remove the @return or move the first line after the @return?

gsmet:
Wouldn't it make sense to do the exact same thing for embedded? Or do we consider embedded in maintenance mode from now on?

DavideD:
When you get a node in embedded mode you basically can already access all the connected nodes. Not sure if it's worth the effort. In Remote mode is different because you have to rest calls

gsmet:
Makes sense!

DavideD:
I've renamed the class, changed the javadoc and used the constant for alias inthe queries.

gsmet:
Merged! Thanks!

PR #716, opened 2016-06-27
OGM-1026 Support Neo4j 3.0.3

DavideD:
https://hibernate.atlassian.net/browse/OGM-1026

This version still requires Apache Lucene 5.5, I was wrong about it when we talked in chat.
 
Changes in the new version:

  * the Rest API response has an additional field `meta` that needs to be mapped in
    `org.hibernate.ogm.datastore.neo4j.remote.json.imp.Row`;
  * libraries need to be updated in the WildFly module;
  * the operator `HAS` in cypher has been replaced by `EXISTS`;
  * some changes in the signature of methods used when dealing with the embedded datastore.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -27,7 +27,7 @@
         <infinispanVersion>8.2.1.Final</infinispanVersion>
         <mongodbVersion>3.2.2</mongodbVersion>
         <fongodbVersion>2.0.6</fongodbVersion>
-        <neo4jVersion>2.3.4</neo4jVersion>
+        <neo4jVersion>3.0.3</neo4jVersion>
         <!-- Update dependency versions accordingly when updating C*  -->
         <cassandraVersion>2.2.0-rc3</cassandraVersion>
         <!-- Update dependency versions accordingly when updating lettuce  -->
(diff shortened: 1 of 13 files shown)


bom/pom.xml
```diff
@@ -36,10 +36,10 @@
         <hdrHistogramVersion>2.1.8</hdrHistogramVersion>
         <latencyUtilsVersion>2.0.3</latencyUtilsVersion>
         <!-- Neo4j supports different versions of Cypher -->
-        <neo4jCypherCompiler19Version>2.0.5</neo4jCypherCompiler19Version>
-        <neo4jCypherCompiler22Version>2.2.9</neo4jCypherCompiler22Version>
+        <neo4jCypherCompiler23Version>2.3.4</neo4jCypherCompiler23Version>
+        <neo4jCypherCompiler30Version>${neo4jVersion}</neo4jCypherCompiler30Version>
```
gsmet: Why do we include several versions of Cypher?

modules/wildfly/pom.xml
```diff
@@ -51,6 +51,13 @@
             <groupId>org.hibernate.ogm</groupId>
             <artifactId>hibernate-ogm-infinispan</artifactId>
             <optional>true</optional>
+            <exclusions>
+                <exclusion>
+                    <!-- The Lucene dependency is not compatible with Neo4j embedded -->
+                    <groupId>org.hibernate</groupId>
+                    <artifactId>hibernate-search-engine</artifactId>
+                </exclusion>
+            </exclusions>
```
gsmet: I'm wondering if it's a good idea to remove that (and the related stuff below - enforcer and so on). The fact that the versions are aligned is just pure luck, isn't it? There's a good chance minor versions might not be aligned if they stick to 5.5.0.

DavideD: The infinispan WildFly module is not using it (it uses the one included in WilfFly) there is really no reason to have it as dependency in this module. If I don't exclude it I cannot include the right version for Neo4j.

Also the integration tests will fail if for some reason a module is missing a library or it is using the wrong version.

bom/pom.xml
DavideD: These are Neo4j dependecies and an error is thrown if hey are not on the classpath

gsmet: Good reason :)

modules/wildfly/pom.xml
DavideD: In this pull request, the versions of Lucene are not aligned. Neo4j is using 5.5.0 while HIbernate Search is using 5.3.1

neo4j/pom.xml
```diff
@@ -32,13 +32,6 @@
                 <artifactId>maven-processor-plugin</artifactId>
             </plugin>
             <plugin>
-                <artifactId>maven-enforcer-plugin</artifactId>
```
DavideD: I removed it because the enforcer is inherited form the parent pom. The reason it was included in this pom is because we decided to keep it disable. Now we don't have conflict anymore, so I want to be alerted when something changes

bom/pom.xml
DavideD: Actaully this was true before, I'll check  quicklynow to make sure that it is still the case

DavideD:
I reverted the changes to the enforcer plugin and the exclusion of search-engine

gsmet:
Merged! Thanks!

PR #718, opened 2016-06-27
OGM-1106 Deprecate Neo4jDialect

DavideD:
https://hibernate.atlassian.net/browse/OGM-1106


gsmet:
Jenkins, retest this please

DavideD:
Jenkins, retest this please

core/src/test/java/org/hibernate/ogm/utils/GridDialectType.java
```diff
@@ -25,7 +25,7 @@
 
 	MONGODB( "org.hibernate.ogm.datastore.mongodb.MongoDBDialect", true, true ),
 
-	NEO4J( "org.hibernate.ogm.datastore.neo4j.Neo4jDialect", false, true),
+	NEO4J( "org.hibernate.ogm.datastore.neo4j.EmbeddedNeo4jDialect", false, true),
```
gsmet: Shouldn't we rename that NEO4J_EMBEDDED?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/Neo4jDialect.java
```diff
@@ -6,592 +6,17 @@
...
- * <p>
- * A {@link Tuple} is saved as a {@link Node} where the columns are converted into properties of the node.<br>
- * An {@link Association} is converted into a {@link Relationship} identified by the {@link AssociationKey} and the
- * {@link RowKey}. The type of the relationship is the value returned by
- * {@link AssociationKeyMetadata#getCollectionRole()}.
- * <p>
- * If the value of a property is set to null the property will be removed (Neo4j does not allow to store null values).
- *
- * @author Davide D'Alto &lt;davide@hibernate.org&gt;
+ * @deprecated use {@link EmbeddedNeo4jDialect}. We might decide to remove this class in the future or change its
```
gunnarmorling: There should be no "might" in here. Either we know it's going to go away and thus we deprecate it, or not. Be more authoritative :)

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/RemoteNeo4jDialect.java
```diff
@@ -270,7 +270,7 @@ private void validate(StatementsResponse readEntity, EntityKey key, Tuple tuple)
 		if (!readEntity.getErrors().isEmpty() ) {
 			ErrorResponse errorResponse = readEntity.getErrors().get( 0 );
 			switch ( errorResponse.getCode() ) {
-				case Neo4jDialect.CONSTRAINT_VIOLATION_CODE:
+				case EmbeddedNeo4jDialect.CONSTRAINT_VIOLATION_CODE:
```
gunnarmorling: Better move that constant to a commonly used place?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/Neo4jDialect.java
DavideD: I'll wait to deprecate it and leave it there for now, although it duplicates the Embedded dialect

DavideD:
In the ned I've added the NEO4J_EMBEDDED, I'm not sure if we should remove the NEO4J options right now. I decided to keep it and make it equivalent to NEO4J_EMBEDDED

core/src/main/java/org/hibernate/ogm/datastore/impl/AvailableDatastoreProvider.java
```diff
@@ -19,6 +19,7 @@
 	EHCACHE( "org.hibernate.ogm.datastore.ehcache.impl.EhcacheDatastoreProvider" ),
 	MONGODB( "org.hibernate.ogm.datastore.mongodb.impl.MongoDBDatastoreProvider" ),
 	FONGO( "org.hibernate.ogm.datastore.mongodb.impl.FongoDBDatastoreProvider" ),
+	NEO4J( "org.hibernate.ogm.datastore.neo4j.embedded.impl.EmbeddedNeo4jDatastoreProvider" ),
```
gsmet: Maybe we should mark it as @Deprecated?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/Neo4jDialect.java
```diff
@@ -6,592 +6,17 @@
...
-	private static final Log log = LoggerFactory.getLogger();
-
-	private final GraphDatabaseService dataBase;
-
-	private final EmbeddedNeo4jSequenceGenerator sequenceGenerator;
-
-	private Map<EntityKeyMetadata, EmbeddedNeo4jEntityQueries> entityQueries;
-
-	private Map<AssociationKeyMetadata, EmbeddedNeo4jAssociationQueries> associationQueries;
+public class Neo4jDialect extends EmbeddedNeo4jDialect {
```
gsmet: Is there really a point in keeping this dialect around?

Personally, I think we should simply remove it: we will have prominent information about the support of the Remote protocol in the next release so I think it's the best moment to remove it and explain to people they have to move to the Embedded one or consider using the Remote one.

If we deprecate it, they probably won't notice it and be surprised when we will remove it as we probably won't have a major change in the Neo4j dialects to advertise.

I'm quite new with this sort of things so I might not be in line with the usual way of doing things.

And I think we should do the same for AvailableDatastoreProvider.NEO4J. IMHO, it's kinda easy to explain we now have 2 Neo4j dialects and people have to choose between the two. If they want to keep the same behavior as before, they have to switch to the NEOJ_EMBEDDED one.

If we don't remove it now, please consider adding back the @Deprecated annotation and state that it will be removed in the next release. I really think we shouldn't keep these old things for too long: it's quite confusing for new users.

gsmet:
OK, so apart from the discussion on "should we remove/deprecate the old Neo4j things?", it's good to go.

I just want us to decide on the way to go, will commit as soon as we have a clear path.

core/src/main/java/org/hibernate/ogm/datastore/impl/AvailableDatastoreProvider.java
DavideD: I sort of like the idea of having a generic variable and choose the dialect based on what we have on the classpath and or properties. Sort of like if you set the database location to /opt/database or http:// localhost:7777 or bolt://...

Unless, we plan of remove it I guess we can wait a bit before deprecating it

DavideD:
@gsmet I've removed AvailableDatastore.NEO4J and deprecated Neo4jDialect

DavideD:
Rebased to the latest changes

gsmet:
Jenkins, retest this please

gsmet:
Merged! Thanks!

PR #719, opened 2016-06-27
Remove unused luceneVersion property from BOM

DavideD:
This variable is not used anywhere, I think it is the remaining of some old code


gsmet:
Merged! Thanks!

PR #721, opened 2016-06-30
OGM-944 Update scripts for release with Jenkins

DavideD:
https://hibernate.atlassian.net/browse/OGM-944

@gsmet Would like to have a look or I'l merge them and start the release of 5.0.1 as a test?


gsmet:
I'll have a quick look just now.

src/main/release-scripts/upload-distribution.sh
```diff
@@ -1,11 +1,27 @@
 #!/usr/bin/env bash
 
-DIST_PARENT_DIR=$1
-RELEASE_VERSION=$2
+if [ -z "$1" ]
+  then
+    echo "Release version not supplied"
+    exit 1
+fi
```
gsmet: You are aware that you inverted the order of the parameters?

src/main/release-scripts/validate-release.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+RELEASE_VERSION=$1
+WORKSPACE=${WORKSPACE:-'.'}
+CHANGELOG=$WORKSPACE/changelog.txt
+README=$WORKSPACE/readme.md
+
+if [ `git tag -l | grep $RELEASE_VERSION` ]
```
Sanne: interesting idea, but it might work better if you first fetch the tags from upstream?

gunnarmorling:
Nice to see this one rolling; @hferentschik 's heritage :)

gsmet:
I just added a minor comment to be sure you didn't make a mistake inverting 2 parameters.

Other than that, didn't spot anything obvious!

src/main/release-scripts/validate-release.sh
DavideD: You mean `git fetch --tags`? I'll add it, just in case.

src/main/release-scripts/upload-distribution.sh
DavideD: Yes, but I'll move the check on RELEASE_VERSION instead of 1 to make it clearer what I'm validating

src/main/release-scripts/validate-release.sh
Sanne: yes, because otherwise you don't have the tags locally and the check won't fail even if these exist already.

DavideD:
Ok, I've applied the changes, I'm going to merge it

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/721

DavideD:
Rebased before merge

PR #722, opened 2016-06-30
OGM-1112 Avoid creation of test jar in integration tests

DavideD:
https://hibernate.atlassian.net/browse/OGM-1112

  We used to create a jar with the integration tests when we had separate
  maven modules for several datastore. Now that all datastores are tested
  in the same project there is no need to create a jar that it is not
  deployed or used.

### DIFF
--- a/integrationtest/pom.xml
+++ b/integrationtest/pom.xml
@@ -89,29 +89,6 @@
                     </configuration>
                 </plugin>
                 <plugin>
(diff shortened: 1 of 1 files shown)


DavideD:
@gsmet Quick look?

gsmet:
Fixed a few typos and merged! Thanks!

PR #723, opened 2016-07-01
OGM-1103 Alternative proposal

gunnarmorling:
Hey @gsmet, I took another look at the issue with fresh mind this morning and propose this alternative:

* It initializes the state eagerly at event begin for all those lifecycles enabled
* It pushes the decision whether to enable or not to lifecycles themselves

I think it's a nice evolution of the overall design. The eager init should be fine, as we'll need the states later on anyways (only for the exotic case of a no-op flush cycle it'd be superfluous). It avoids creation of that new dialect facet, too.

I believe it also revealed another issue on the side, where we wouldn't initialize the context properly in non-JPA.

The work would need a bit more polishing (esp. the test), but maybe you could take it from here if you like the approach?


gsmet:
Closing as I made additional cleanups on the proposal.

PR #724, opened 2016-07-04
OGM-1115 Relax regexp used to detect primary key constraint violation

gsmet:
https://hibernate.atlassian.net/browse/OGM-1115

More modern versions of MongoDB have a different syntax.

The regexp we use now should be sufficiently future proof.

I'm not very fond of this and we should probably try to get the MongoDB people to differentiate the 2 exceptions but IIRC the server doesn't do any difference between the 2.


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/724

PR #725, opened 2016-07-04
[5.1] OGM-982 - Upgrade to ORM 5.1

gunnarmorling:
Should be good to go now. I re-use the existing ORM 5.1.1-SNAPSHOT module ZIP now, so no need for the further "skinny module" approach. We only need to upgrade to 5.1.1.Final once out.

PR #726, opened 2016-07-11
OGM-982 Update to ORM 5.1.1

DavideD:
https://hibernate.atlassian.net/browse/OGM-982
This should add the missing bits to  https://github.com/hibernate/hibernate-ogm/pull/725

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/boot/model/naming/impl/OgmImplicitNamingStrategy.java
+++ b/core/src/main/java/org/hibernate/ogm/boot/model/naming/impl/OgmImplicitNamingStrategy.java
@@ -6,8 +6,6 @@
  */
 package org.hibernate.ogm.boot.model.naming.impl;
 
-import java.util.regex.Pattern;
-
 import org.hibernate.boot.model.naming.ImplicitNamingStrategy;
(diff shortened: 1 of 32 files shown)


core/src/main/java/org/hibernate/ogm/boot/model/naming/impl/OgmImplicitNamingStrategy.java
```diff
@@ -23,19 +21,13 @@
...
-		// for element collections just use the simple name
-		if ( parts.length == 2 ) {
-			return parts[1];
-		}
-		else {
-			return attributePath.getFullPath();
-		}
+		String fullPath = attributePath.getFullPath();
+		String attribute = fullPath.replaceAll( ELEMENT_COLLECTION_NAME_PATTERN, "" );
+		return attribute;
```
gsmet: BTW, if you end up doing something like that, the idea is to not mess with strings anymore. You have to navigate in the `AttributePath` hierarchy and use the `isCollectionElement` method. Steve plans to change the `collection&&element` part in a near future.

DavideD: Thanks

DavideD:
Jenkins, retest this, please

PR #728, opened 2016-07-12
OGM-982 Update to ORM 5.1.1

DavideD:
Follows https://github.com/DavideD/hibernate-ogm/tree/OGM-982-UpdateORM
https://hibernate.atlassian.net/browse/OGM-982

@gsmet Will this work?

I think fixing this also exposed a bug in the mapping for cassandra. Since it is experimental I've changed the expected mapping. See https://github.com/hibernate/hibernate-ogm/commit/9da7dc46b7a8cce2e0753cbcfe10e6fbad4cbd8d


DavideD:
Rebased to the latest changes

This should be ready to be merged @gsmet

gsmet:
@Davide OK, I was hoping for some feedback on my mail on #hibernate-dev. But, it's probably better to merge it anyway. I'll add my test on the List<Embeddable> case and merge it.

DavideD:
We can wait for some feedback.

Sanne:
why is this also including an override of the modules version of Hibernate Search ?

I was hoping that we could treat that as an orthogonal matter, as in fact we want to (eventually) be able to test with either combination.

Not a blocker, but I'd like to understand why this had to be included.

Sanne:
It passes all tests for me, nice :+1: 

Looks good; I didn't investigate in the small Cassandra and Neo4J adjustments but I can trust you on that.

Not merging as you seem to have some open questions still? I'd be +1 to merge.

gsmet:
@Sanne see my post called "[OGM] Mapping of embedded" on hibernate-dev

Sanne:
Ok I was thinking to push it and move on, but seems ORM 5.1.1 didn't release yet (this PR is using a snapshot): I don't think that's acceptable to merge yet.

gsmet:
Current status: waiting for ORM 5.1.1 release.

DavideD:
@gsmet or @gunnarmorling could you have a quick look at this?

I rebased it and added a few commits to fix some exceptions that weren't present before

modules/src/main/assembly/dist.xml
```diff
@@ -225,8 +225,6 @@
                  <include>biz.paluch.redis:lettuce</include>
                  <include>org.apache.commons:commons-pool2</include>
                  <include>io.reactivex:rxjava</include>
-                 <include>org.latencyutils:LatencyUtils</include>
-                 <include>org.hdrhistogram:HdrHistogram</include>
               </includes>
```
gsmet: Good catch.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/dialect/impl/RemoteNeo4jTupleAssociationSnapshot.java
```diff
@@ -64,8 +64,6 @@ else if ( targetNode.containsKey( targetColumnName ) ) {
 				else {
 					// Ex: @ElementCollection List<Embedded> examples
 
-					targetColumnName = targetColumnName.substring( collectionRole.length()  + 1 );
-
```
gsmet: This comes a bit out of the blue. Do you have an explanation on why it suddenly breaks after the upgrade of ORM?

DavideD: I suspect is the same reason this has changed: https://github.com/hibernate/hibernate-ogm/pull/728/files#diff-378d3f25166f2e2a798f69b74262fe86L66

The column has already the right name and therfore we don't have remove the collectionrole from the string.

DavideD:
@gsmet Does my comment answer your question?

gsmet:
@DavideD yes, sorry for the delay.

Any chance you could include a test inspired by https://gist.github.com/gsmet/0652294523b2c48efe72668ccc0a6e1c ?

I suspect this commit will fix https://hibernate.atlassian.net/browse/OGM-893 and I would prefer to have a unit test for that.

Thanks!

DavideD:
@gsmet I've added the tests for [OGM-893](https://hibernate.atlassian.net/browse/OGM-893)
The mapping for collections seems correct now

gsmet:
Nice! Fixed a typo in the last 3 commits message and merged!

gunnarmorling:
Yeah :)

PR #729, opened 2016-07-13
OGM-1119 / OGM-1120 Improve test infrastructure to speed up Cassandra tests

gsmet:
* https://hibernate.atlassian.net/browse/OGM-1119
 * https://hibernate.atlassian.net/browse/OGM-1120

Various cleanups to the test infrastructure.

Introduce `OgmJpaTestCase` and `OgmJpaTestRunner` to improve performances of the Cassandra tests. They are as symmetrical as possible to `OgmTestCase` and `OgmTestRunner`.

Avoid as many calls to `cluster.connect()` as possible as they are really slow.

### DIFF
--- a/cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraDatastoreProvider.java
+++ b/cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraDatastoreProvider.java
@@ -107,7 +107,7 @@ public class CassandraDatastoreProvider extends BaseDatastoreProvider
 	public void start() {
 		if ( cluster == null ) {
 			if ( ! config.getHosts().isSingleHost() ) {
-				throw new HibernateException( "Hibernate OGM Cassandra backend does not yet support multiple hosts, Coming soom" );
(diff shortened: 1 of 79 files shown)


cassandra/src/main/java/org/hibernate/ogm/datastore/cassandra/impl/CassandraDatastoreProvider.java
```diff
@@ -119,14 +119,13 @@ public void start() {
 						.withCredentials( config.getUsername(), config.getPassword() )
 						.build();
 
-				Session bootstrapSession = cluster.connect();
-				bootstrapSession.execute(
+				session = cluster.connect();
```
Sanne: I have no experience with Cassandra, but by reading the javadoc it seems that we might be able to go a step further:

the `connect()` here is only needed to run the `CREATE KEYSPACE IF NOT EXISTS`.

I'm not sure if all users will want us to invoke that at each bootstrap, shouldn't it depend on the `hibernate.hbm2ddl.auto` ?

Anyway, that's just a thought for future.. the patch looks like better than the original so applying ;)

Sanne:
merged, thanks!

PR #731, opened 2016-07-18
OGM-1122 Take into account currentState in getRowKeys() even if cleared

gsmet:
https://hibernate.atlassian.net/browse/OGM-1122

Even if the association has been cleared, we have to take into account the
rows added in `currentState` after the clear. Prior to this patch,
`getRowKeys()` always returned an empty list if the association has been
cleared.


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/731

Thanks

PR #734, opened 2016-07-22
OGM-1124 Update doc server metadata on release

gsmet:
https://hibernate.atlassian.net/browse/OGM-1124

We need to install on the build server (if not already available):
 * wget
 * sed
 * jq


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/734

Thanks

PR #736, opened 2016-08-07
OGM-1129 Upgrade integration tests to Wildfly 10.1.0.CR1

Sanne:


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/736

Thanks

PR #738, opened 2016-08-08
OGM-1132 Document that we do not support named parameters for Cassandra

gsmet:
On Cassandra, we currently only support ordinal parameters for CQL
natives queries.


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/738

PR #742, opened 2016-08-09
OGM-1132 Make sure to have some feedback if somebody else fix OGM-1008

DavideD:
An optional additional commit to avoid having too many commonets in the unit tests and to give some feedback about the fact that a unit test already exist for the feature.


gsmet:
Good idea.

Merged! Thanks!

PR #744, opened 2016-08-09
OGM-1131 Control the dependency version of Apache Lucene explicitly, …

Sanne:
…upgrade to 5.5.2

 - https://hibernate.atlassian.net/browse/OGM-1131

### DIFF
--- a/modules/wildfly/pom.xml
+++ b/modules/wildfly/pom.xml
@@ -95,13 +95,12 @@
             <artifactId>hibernate-envers</artifactId>
             <optional>true</optional>
         </dependency>
-        <!-- This is to satisfy Neo4J's custom Lucene,
-          not to be confused with the one used by Hibernate Search -->
+        <!-- This is to satisfy Neo4J's custom Lucene: might not match the needs of Hibernate Search -->
         <dependency>
             <groupId>org.apache.lucene</groupId>
             <artifactId>lucene-core</artifactId>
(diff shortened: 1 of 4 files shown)


modules/wildfly/pom.xml
```diff
@@ -95,13 +95,12 @@
             <artifactId>hibernate-envers</artifactId>
             <optional>true</optional>
         </dependency>
-        <!-- This is to satisfy Neo4J's custom Lucene,
-          not to be confused with the one used by Hibernate Search -->
+        <!-- This is to satisfy Neo4J's custom Lucene: might not match the needs of Hibernate Search -->
```
gunnarmorling: Is this comment actually true any longer, now that it's only one version?

Sanne: Yes the point is that each of these would otherwise bring their own
transitive dependency. There are minor mismatches, like in this case a
micro version although we can manage some other combinations too. The only
reasonable strategy is to manage this explicitly ourselves rather than who
comes first transient..

gunnarmorling: Ok, got it. I think the "satisfy" is what confused me. You actually override Neo4j's own version.

Sanne: not really. In this section we don't pick the version, we're just adding the Lucene dependencies so that they get copied into the Neo4J modules.

PR #745, opened 2016-08-09
OGM-1135 Avoid depending on both geronimo-jta_1.1_spec and Narayana's…

Sanne:
… JTA spec

This was causing me loads of classloader miseries.

Added it to the "banned dependency list" of the Maven enforcer plugin so that it never happens again.

 - https://hibernate.atlassian.net/browse/OGM-1135


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/745

PR #750, opened 2016-08-10
Travis - Use 127.0.0.1 instead of localhost in .travis.yml

gsmet:
See https://github.com/travis-ci/travis-ci/issues/6449


gunnarmorling:
Rebased and applied.

PR #751, opened 2016-08-10
OGM-1116 Be consistent with Search/Validator in the way modules are built

gsmet:
* https://hibernate.atlassian.net/browse/OGM-1116


modules/pom.xml
```diff
@@ -16,8 +16,8 @@
     <artifactId>hibernate-ogm-modules</artifactId>
     <packaging>pom</packaging>
 
-    <name>Hibernate OGM JBoss Modules</name>
-    <description>Create JBoss Modules for Hibernate OGM</description>
+    <name>Hibernate OGM Wildfly Module</name>
+    <description>Hibernate OGM modules creation to use in JBoss AS / Wildfly</description>
```
gunnarmorling: It's WildFly only.

gunnarmorling:
Looks good. The nested structure was there because at some point we had two ZIPs: one for WF and one for EAP, but the latter got outdated and I don't think it will come back. So +1 for this change.

gsmet:
Fixed the wording and rebased.

gunnarmorling:
Merged; thanks, @gsmet!

gunnarmorling:
Ah, can you add an entry to the migration notes about the changed artifact id and classifier? Thanks!

PR #754, opened 2016-08-11
OGM-1145 Upgrade maven-jdocbook-plugin and pressgang

gsmet:
* https://hibernate.atlassian.net/browse/OGM-1145

It allows to generate valid HTML5 output and fixes a bug where the user
was not brought to the right place when clicking on an anchor.


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/754

DavideD:
Rebased before merge

PR #756, opened 2016-08-16
OGM-1147 IdSourceKey doesn't need to handle multi column types

Sanne:
- https://hibernate.atlassian.net/browse/OGM-1147

### DIFF
--- a/ehcache/src/main/java/org/hibernate/ogm/datastore/ehcache/persistencestrategy/kind/impl/SerializableIdSourceKey.java
+++ b/ehcache/src/main/java/org/hibernate/ogm/datastore/ehcache/persistencestrategy/kind/impl/SerializableIdSourceKey.java
@@ -10,7 +10,6 @@ import java.io.Externalizable;
 import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
-import java.util.Arrays;
 
 import org.hibernate.ogm.datastore.ehcache.persistencestrategy.common.impl.VersionChecker;
 import org.hibernate.ogm.model.key.spi.IdSourceKey;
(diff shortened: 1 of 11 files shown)


ehcache/src/main/java/org/hibernate/ogm/datastore/ehcache/persistencestrategy/kind/impl/SerializableIdSourceKey.java
```diff
@@ -100,24 +109,25 @@ else if ( !table.equals( other.table ) ) {
...
+		return "SerializableIdSourceKey [columnName='" + columnName + "', columnValues='" + columnValue + "']";
 	}
 
 	@Override
 	public void writeExternal(ObjectOutput out) throws IOException {
 		out.writeInt( VERSION );
 		out.writeUTF( table );
-		out.writeObject( columnNames );
-		out.writeObject( columnValues );
+		//Wrapping in String[] and Object[] respectively as this used to be the format,
```
gunnarmorling: Should we make this an option?

gunnarmorling:
Looks great apart from that one question whether the compat. behavior for ISPN should be an option. I.e. allowing users to keep the legacy format for now but don't use it e.g. in new apps. In 6, this option would go away.

Sanne:
Yes that's reasonable, but I'd rather open a follow up issue as it's not trivial - I'd need this change to move on with Hot Rod.

BTW the (future) encoding change would affect both Infinispan and Ehcache.

Sanne:
merged!

Sanne:
followed up by https://hibernate.atlassian.net/browse/OGM-1148

PR #757, opened 2016-08-19
OGM-1064 Add tests for Redis and MongoDB to show how many calls to the

gsmet:
datastore are performed on simple operations

This is extracted from OGM-1064 and has value by itself.


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/757

I've changed the word "resetted" to "reset" in the comment and rebased

PR #760, opened 2016-08-25
OGM-1151, OGM-893 Update comments and Jira in code

DavideD:
A couple of minor fixes that were left behind


gsmet:
Merged! Thanks!

PR #761, opened 2016-08-31
OGM-1152 WIP for whoever wi going to work on this issue. Add GridDialectOperationInvocationTest for @OneToOne @JoinColumn

emmanuelbernard:
This test is not complete but offers an overview of persist

Also I suspect the "LOOK_UP" driven operations will vary from dialect to dialect so we would need a test based on `GridDialect.getDuplicateInsertPreventionStrategy`.


emmanuelbernard:
hum, need to do more work. Closing

PR #762, opened 2016-08-31
OGM-1152 WIP Dialect operation invocation tests started for the person taking OGM-1152

emmanuelbernard:
Not a compete test suite but has explanation for each operation being called.


Sanne:
Could we merge this already, even if there's more work to do?

gsmet:
@Sanne yes, I plan to finish this work soon. It needs some more love (typically, it doesn't take into account batching so it won't work with MongoDB dialect).

emmanuelbernard:
@Sanne I'm pretty sure tests don't pass on all dialects as it is.

gsmet:
I can confirm that :). Closing this one as I'm working on OGM-1152 and will provide another PR.

PR #763, opened 2016-09-06
OGM-1152 Dialect operation invocation tests

gsmet:
This is a follow-up of https://github.com/hibernate/hibernate-ogm/pull/762 . The tests now pass for all the dialects (I had to ignore some of them anyway but most of them are covered).

I documented a few other things, mostly the behavior of MongoDB.

Please don't close the issue as we have to add tests for the other types of associations but I wanted to get this in, in case someone else wants to take a look at it.

### DIFF
--- a/core/src/test/java/org/hibernate/ogm/backendtck/associations/onetoone/GridDialectOperationInvocationForOneToOneTest.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/associations/onetoone/GridDialectOperationInvocationForOneToOneTest.java
@@ -1,0 +1,103 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.backendtck.associations.onetoone;
+
+import java.util.List;
+import java.util.Map;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import org.hibernate.Session;
+import org.hibernate.Transaction;
+import org.hibernate.ogm.cfg.OgmProperties;
+import org.hibernate.ogm.dialect.impl.GridDialects;
+import org.hibernate.ogm.dialect.spi.GridDialect;
+import org.hibernate.ogm.utils.InvokedOperationsLoggingDialect;
+import org.hibernate.ogm.utils.OgmTestCase;
+
+import static org.fest.assertions.Assertions.assertThat;
+
+/**
+ * @author Emmanuel Bernard emmanuel@hibernate.org
+ */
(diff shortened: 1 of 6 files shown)


core/src/test/java/org/hibernate/ogm/backendtck/dialectinvocations/GridDialectOperationInvocationForOneToOneTest.java
```diff
@@ -14,6 +14,8 @@
 
 import org.hibernate.Session;
 import org.hibernate.Transaction;
+import org.hibernate.ogm.backendtck.associations.onetoone.Husband;
```
emmanuelbernard: For info, I'm not a bit fan of tests suing model classes from other tests because when one adds new properties it can destabilize them. Not sure what to do.

core/src/test/java/org/hibernate/ogm/backendtck/associations/onetoone/GridDialectOperationInvocationForOneToOneTest.java
```diff
@@ -15,16 +15,25 @@
...
+import org.hibernate.ogm.utils.SkipByGridDialect;
 
 import static org.fest.assertions.Assertions.assertThat;
 
 /**
  * @author Emmanuel Bernard emmanuel@hibernate.org
+ * @author Guillaume Smet
  */
+@SkipByGridDialect(value = { GridDialectType.CASSANDRA, GridDialectType.NEO4J_REMOTE, GridDialectType.NEO4J_EMBEDDED, GridDialectType.REDIS_HASH },
+		comment = "For Cassandra and Neo4j, the getAssociation always return an association, thus we don't have the createAssociation call. " +
```
emmanuelbernard: Hum, interesting, I was not aware of this different behavior. Not sure about the implication :/

emmanuelbernard: A cursory look at the code shows to big problem as long as assocation.size()  and association.getRows() return 0 and an empty set respectively.

core/src/test/java/org/hibernate/ogm/backendtck/dialectinvocations/GridDialectOperationInvocationForOneToOneTest.java
gsmet: Yes, it bothered me too but I didn't like having these tests in different packages either as I think it's interesting to have a global vision of how it works.

DavideD: I would prefer having separate as well, but we usually don't change existing models so it does not bother me too much.

DavideD:
Rebased and merged https://github.com/hibernate/hibernate-ogm/pull/763

@gsmet Could you update the JIRA explainig the use cases that are missing?

DavideD:
Thanks

PR #766, opened 2016-09-08
Hot rod dialect v4

Sanne:
Some pending tasks are documented as sub-tasks on JIRA:
 - https://hibernate.atlassian.net/browse/OGM-353

(Although, I've started including some of the sub-tasks already since opening this PR; next working on the missing Sequencer tests and then documentation)

WildFly modules included
  - https://hibernate.atlassian.net/browse/OGM-1160

# Launching / Exploring tests

As usual, there's a `org.hibernate.ogm.datastore.infinispanremote.utils.InfinispanRemoteBackendTckHelper` to help you start a test from OGM/core on this new dialect.
It includes a `@ClassRule` which will start an Infinispan Server as needed, or you can launch it via the static `main()` method to keep the server running (manually).

The tests which are specific to this module will use the same JUnit Rule so the database is auto-started as needed.

When running it via Maven, the Surefire plugin is instructed to activate the testing listener `org.hibernate.ogm.datastore.infinispanremote.utils.HotrodServerLifecycle`, which will keep the Infinispan Server running for the duration of the tests, and rest its state as needed.

I like this new approach, as the test execution is very fast and the debugging / development sessions are more practical.. we might consider evolving other modules to use a similar approach.

For the Arquillian tests, the magic is handled by an Arquillian extension which invokes the same JUnit helpers as before, but in the right timing as it needs to happen before the deployment: `org.hibernate.ogm.test.integration.infinispanremote.HotRodServerLifecycleManager` 
 - https://github.com/Sanne/hibernate-ogm/blob/bed210d92f8a3cf3b4033b8971b5206bef0665d0/integrationtest/src/test/java/org/hibernate/ogm/test/integration/infinispanremote/HotRodServerLifecycleManager.java

# Schema & Encoders

This dialect needs a full schema to be deployed to the Server before any I/O can happen.

We already discussed the shape and strategy of the schema, but to remind you it's 1:1 of how ORM would map it to RDBMs Tables, I'll highlight the exceptions to this rule in the following sections.

The schema build process is driven by `org.hibernate.ogm.datastore.infinispanremote.impl.ProtobufSchemaInitializer`.

The core element representing a "Table" mapping is
`org.hibernate.ogm.datastore.infinispanremote.impl.schema.TableDefinition`
 - https://github.com/Sanne/hibernate-ogm/blob/bed210d92f8a3cf3b4033b8971b5206bef0665d0/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/schema/TableDefinition.java

These "TableDefinitions" are generated upfront during boot time, and are immutable after that.
They use "ProtofieldWriterSet" which encapsulates an ordered set of field definitions, each with its:
 - ORM column name
 - Protobuf field name
 - Encoder/Decoder references

So a single _TableDefinition_ is able to both export itself in string format to generate the schema DDL fragment, but is also containing the function to decode / encode the payload to the underlying buffers.

The actual driver for encoding / decoding is `org.hibernate.ogm.datastore.infinispanremote.impl.protostream.ProtoDataMapper` , so the main dialect will interact with the _ProtoDataMapper_

## Capturing the generated Schema

There's a pluggable SPI `org.hibernate.ogm.datastore.infinispanremote.spi.schema.SchemaCapture` to intercept / capture the generated protobuf schema. This is controlled by a configuration option, and the default is to log the generated schema but in some tests I make use of it to validate the schema generation.
The user could also make use of this to override the schema generation and deploy a custom one (not that I'd suggest doing that as our encoders wouldn't be affected).

## Sequences and Table based generators

A sequence is mapped to a key / value, where the key is the sequence name, and stored in a Cache having the sequence name.

When using a table generator, ORM can store multiple "lines" called segments; in this case we consider the Table name as Cache name and each combination segment name part of the ID for the key/value store.

What's interesting is that, no matter if ORM ops for Sequences or Table strategy, we're actually storing in the same way.

The actual sequences are implemented via versioned CAS operations:
 - https://github.com/Sanne/hibernate-ogm/blob/bed210d92f8a3cf3b4033b8971b5206bef0665d0/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/sequences/HotRodSequencer.java

## Bags & unordered collections

Where a collection is of a "bag" type, ORM would map this to a table lacking a primary key, I guess primarily to support duplicate entries.
In a Key/Value store, the key is hardly optional so I have found no way to map this, and noticed the Cassandra dialect doesn't support it either.
The good news is that we detect this early on (at bootstrap, during schema generation) and throw a helpful exception.

# Versioning and Optimistic locking

Optimistic locking is used as an implementation detail by the Sequence generator, but it's not used as a general storage feature by the dialect.

There is some commented-out code in the `InfinispanRemoteDialect` which I believe should make it easy to implement proper optimistic locking in the near future, but I suspect this requires the fix on number of operations by @gsmet first.

# InfinispanRemoteDialect

The main Dialect driver is org.hibernate.ogm.datastore.infinispanremote.InfinispanRemoteDialect<EK, AK, ISK>.

Among the optional features, it supports `MultigetGridDialect`.

The Dialect interacts with the usual Tuples from OGM but internally uses the pairs 
-  `ProtostreamPayload` to represent a ready-to-be-encoded Value
- `ProtostreamId` to represent a ready-to-be-encoded Key

These are the only types recognized by the two-way encoders so actual storage operations need to be handled using these payload objects.
Finally, all actual Cache IO operations have to be executed within the context of a "mapper.withinCacheEncodingContext( lambda )" so that the DataMapper can setup the right encoding scheme for the specific Cache. This is to workaround limitation in how the Protostream encoders are configured, as normally they have to be setup 1:1 to class types.. more on this on JIRA s to follow up so we can cleanup the ThreadLocal I'm using for this workaround.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -51,6 +51,7 @@
 
         <!-- Infinispan -->
         <infinispanVersion>8.2.4.Final</infinispanVersion>
+        <protostreamVersion>3.0.5.Final</protostreamVersion>
 
         <!-- Ehcache -->
         <ehcacheVersion>2.6.11</ehcacheVersion>
@@ -397,6 +398,18 @@
             </dependency>
             <dependency>
                 <groupId>${project.groupId}</groupId>
+                <artifactId>hibernate-ogm-infinispan-remote</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>${project.groupId}</groupId>
+                <artifactId>hibernate-ogm-infinispan-remote</artifactId>
+                <version>${project.version}</version>
+                <type>test-jar</type>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>${project.groupId}</groupId>
                 <artifactId>hibernate-ogm-mongodb</artifactId>
                 <version>${project.version}</version>
             </dependency>
@@ -507,6 +520,52 @@
                 <version>${infinispanVersion}</version>
                 <type>test-jar</type>
                 <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.infinispan</groupId>
+                <artifactId>infinispan-client-hotrod</artifactId>
+                <version>${infinispanVersion}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.infinispan.protostream</groupId>
+                <artifactId>protostream</artifactId>
+                <version>${protostreamVersion}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.infinispan</groupId>
+                <artifactId>infinispan-client-hotrod</artifactId>
+                <version>${infinispanVersion}</version>
+                <type>test-jar</type>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.infinispan</groupId>
+                <artifactId>infinispan-server-hotrod</artifactId>
+                <version>${infinispanVersion}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.infinispan</groupId>
+                <artifactId>infinispan-server-hotrod</artifactId>
+                <type>test-jar</type>
+                <version>${infinispanVersion}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.infinispan</groupId>
+                <artifactId>infinispan-remote-query-client</artifactId>
+                <version>${infinispanVersion}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.infinispan</groupId>
+                <artifactId>infinispan-query-dsl</artifactId>
+                <version>${infinispanVersion}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.infinispan</groupId>
+                <artifactId>infinispan-remote-query-server</artifactId>
+                <scope>test</scope>
+                <version>${infinispanVersion}</version>
             </dependency>
 
             <!-- MongoDB -->
--- a/core/src/main/java/org/hibernate/ogm/datastore/impl/DatastoreProviderType.java
+++ b/core/src/main/java/org/hibernate/ogm/datastore/impl/DatastoreProviderType.java
@@ -16,6 +16,7 @@ package org.hibernate.ogm.datastore.impl;
 public enum DatastoreProviderType {
 	MAP( "org.hibernate.ogm.datastore.map.impl.MapDatastoreProvider" ),
 	INFINISPAN( "org.hibernate.ogm.datastore.infinispan.impl.InfinispanDatastoreProvider" ),
+	INFINISPAN_REMOTE( "org.hibernate.ogm.datastore.infinispanremote.impl.InfinispanRemoteDatastoreProvider" ),
 	EHCACHE( "org.hibernate.ogm.datastore.ehcache.impl.EhcacheDatastoreProvider" ),
 	MONGODB( "org.hibernate.ogm.datastore.mongodb.impl.MongoDBDatastoreProvider" ),
 	FONGO( "org.hibernate.ogm.datastore.mongodb.impl.FongoDBDatastoreProvider" ),
--- a/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
+++ b/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
@@ -1,0 +1,380 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.datastore.infinispanremote;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Set;
+
+import org.hibernate.ogm.datastore.infinispanremote.impl.ProtoStreamMappingAdapter;
+import org.hibernate.ogm.datastore.infinispanremote.impl.ProtostreamAssociationMappingAdapter;
+import org.hibernate.ogm.datastore.infinispanremote.impl.VersionedTuple;
+import org.hibernate.AssertionFailure;
+import org.hibernate.ogm.datastore.infinispanremote.impl.InfinispanRemoteDatastoreProvider;
+import org.hibernate.ogm.datastore.infinispanremote.impl.protostream.ProtostreamId;
+import org.hibernate.ogm.datastore.infinispanremote.impl.protostream.ProtostreamPayload;
+import org.hibernate.ogm.datastore.infinispanremote.logging.impl.Log;
+import org.hibernate.ogm.datastore.infinispanremote.logging.impl.LoggerFactory;
+import org.hibernate.ogm.datastore.map.impl.MapAssociationSnapshot;
+import org.hibernate.ogm.dialect.multiget.spi.MultigetGridDialect;
+import org.hibernate.ogm.dialect.spi.AssociationContext;
+import org.hibernate.ogm.dialect.spi.AssociationTypeContext;
+import org.hibernate.ogm.dialect.spi.BaseGridDialect;
+import org.hibernate.ogm.dialect.spi.DuplicateInsertPreventionStrategy;
+import org.hibernate.ogm.dialect.spi.ModelConsumer;
+import org.hibernate.ogm.dialect.spi.NextValueRequest;
+import org.hibernate.ogm.dialect.spi.TupleAlreadyExistsException;
+import org.hibernate.ogm.dialect.spi.TupleContext;
+import org.hibernate.ogm.model.key.spi.AssociationKey;
+import org.hibernate.ogm.model.key.spi.AssociationKeyMetadata;
+import org.hibernate.ogm.model.key.spi.EntityKey;
+import org.hibernate.ogm.model.key.spi.EntityKeyMetadata;
+import org.hibernate.ogm.model.key.spi.RowKey;
+import org.hibernate.ogm.model.spi.Association;
+import org.hibernate.ogm.model.spi.AssociationOperation;
+import org.hibernate.ogm.model.spi.AssociationOperationType;
+import org.hibernate.ogm.model.spi.Tuple;
+import org.infinispan.client.hotrod.MetadataValue;
+import org.infinispan.client.hotrod.Search;
+import org.infinispan.client.hotrod.VersionedValue;
+import org.infinispan.commons.util.CloseableIterator;
+import org.infinispan.query.dsl.FilterConditionContext;
+import org.infinispan.query.dsl.Query;
+import org.infinispan.query.dsl.QueryBuilder;
+import org.infinispan.query.dsl.QueryFactory;
+
+/**
+ *  Some implementation notes for evolution:
+ *
+ *  - QueryableGridDialect can't be implemented as "native queries" in Hot Rod are DSL based
+ *    and have no String representation; this might change as Hot Rod exposes the underlying
+ *    query representation which is similar to HQL; alternatively we could look at exposing
+ *    native queries in some way other than a String.
+ *
+ *  - OptimisticLockingAwareGridDialect can't be implemented as it requires an atomic replace
+ *    operation on a subset of the columns, while atomic operations in Hot Rod have to involve
+ *    either the full value or the version metadata of Infinispan's VersionedValue.
+ *
+ *  - IdentityColumnAwareGridDialect can't work out of the box. I suspect we could do this but
+ *    would need extending the Infinispan server deployment with some extension such as a
+ *    custom script to be invoked from the client.
+ *
+ *  - BatchableGridDialect could probably be implemented.
+ *
+ * @author Sanne Grinovero
+ */
+public class InfinispanRemoteDialect<EK,AK,ISK> extends BaseGridDialect implements MultigetGridDialect {
+
+	private static final Log log = LoggerFactory.getLogger();
+
+	private final InfinispanRemoteDatastoreProvider provider;
+
+	public InfinispanRemoteDialect(InfinispanRemoteDatastoreProvider provider) {
+		this.provider = Objects.requireNonNull( provider );
+	}
+
+	@Override
+	public Tuple getTuple(EntityKey key, TupleContext tupleContext) {
+		final String cacheName = key.getTable();
+		ProtoStreamMappingAdapter mapper = provider.getDataMapperForCache( cacheName );
+		ProtostreamId idBuffer = mapper.createIdPayload( key.getColumnNames(), key.getColumnValues() );
+		VersionedValue<ProtostreamPayload> v = mapper.withinCacheEncodingContext( c -> c.getVersioned( idBuffer ) );
+		if ( v == null ) {
+			return null;
+		}
+		ProtostreamPayload payload = v.getValue();
+		if ( payload == null ) {
+			return null;
+		}
+		long version = v.getVersion();
+		VersionedTuple versionedTuple = payload.toVersionedTuple();
+		versionedTuple.setVersion( version );
+		return versionedTuple;
+	}
+
+	@Override
+	public Tuple createTuple(EntityKey key, TupleContext tupleContext) {
+		return new VersionedTuple( true );
+	}
+
+	@Override
+	public void insertOrUpdateTuple(EntityKey key, Tuple tuple, TupleContext tupleContext) {
+		VersionedTuple versionedTuple = (VersionedTuple) tuple;
+		final String cacheName = key.getTable();
+		log.debugf( "insertOrUpdateTuple for key '%s' on cache '%s'", key, cacheName );
+		ProtoStreamMappingAdapter mapper = provider.getDataMapperForCache( cacheName );
+		ProtostreamPayload valuePayload = mapper.createValuePayload( versionedTuple );
+		ProtostreamId idBuffer = mapper.createIdPayload( key.getColumnNames(), key.getColumnValues() );
+		boolean optimisticLockFailed;
+		if ( versionedTuple.isBeingInserted() ) {
+			optimisticLockFailed = null != mapper.withinCacheEncodingContext( c -> c.putIfAbsent( idBuffer, valuePayload ) );
+			if ( optimisticLockFailed ) {
+				throw new TupleAlreadyExistsException( key.getMetadata(), tuple );
+			}
+		}
+		else {
+			mapper.withinCacheEncodingContext( c -> c.put( idBuffer, valuePayload ) );
+		}
+	}
+
+	@Override
+	public void removeTuple(EntityKey key, TupleContext tupleContext) {
+		final String cacheName = key.getTable();
+		log.debugf( "removeTuple for key '%s' on cache '%s'", key, cacheName );
+		ProtoStreamMappingAdapter mapper = provider.getDataMapperForCache( cacheName );
+		ProtostreamId idBuffer = mapper.createIdPayload( key.getColumnNames(), key.getColumnValues() );
+		mapper.withinCacheEncodingContext( c -> c.remove( idBuffer ) );
+	}
+
+	@Override
+	public Association getAssociation(AssociationKey key, AssociationContext associationContext) {
+		Map<RowKey,Map<String, Object>> results = loadRowKeysByQuery( key );
+		return new Association( new MapAssociationSnapshot( results ) );
+	}
+
+	private Map<RowKey, Map<String, Object>> loadRowKeysByQuery(AssociationKey key) {
+		final String cacheName = key.getTable();
+		ProtostreamAssociationMappingAdapter mapper = provider.getCollectionsDataMapper( cacheName );
+		return mapper.withinCacheEncodingContext( c -> {
+			QueryFactory queryFactory = Search.getQueryFactory( c );
+			final String[] columnNames = key.getColumnNames();
+			QueryBuilder qb = queryFactory.from( ProtostreamPayload.class );
+			FilterConditionContext bqEnd = null;
+			boolean firstIteration = true;
+			for ( int i = 0; i < columnNames.length; i++ ) {
+				String fieldName = mapper.convertColumnNameToFieldName( columnNames[i] );
+				if ( firstIteration ) {
+					bqEnd = qb.having( fieldName ).eq( key.getColumnValues()[i] );
+					firstIteration = false;
+				}
+				else {
+					bqEnd = bqEnd.and().having( fieldName ).eq( key.getColumnValues()[i] );
+				}
+			}
+			Query query = bqEnd.toBuilder().build();
+			Map<RowKey,Map<String, Object>> resultsCollector = new HashMap<>();
+			try ( CloseableIterator<Entry<Object,Object>> iterator = c.retrieveEntriesByQuery( query, null, 100 ) ) {
+				while ( iterator.hasNext() ) {
+					Entry<Object,Object> e  = iterator.next();
+					ProtostreamPayload value = ( (ProtostreamPayload) e.getValue() );
+					Map<String, Object> entryObject = value.toMap();
+					RowKey entryKey = value.asRowKey( key );
+					resultsCollector.put( entryKey, entryObject );
+				}
+			}
+			return resultsCollector;
+		} );
+	}
+
+	@Override
+	public Association createAssociation(AssociationKey key, AssociationContext associationContext) {
+		Map<RowKey, Map<String, Object>> associationMap = new HashMap<RowKey, Map<String,Object>>();
+		return new Association( new MapAssociationSnapshot( associationMap ) );
+	}
+
+	@Override
+	public void insertOrUpdateAssociation(AssociationKey key, Association association, AssociationContext associationContext) {
+		final String cacheName = key.getTable();
+		final ProtoStreamMappingAdapter mapper = provider.getDataMapperForCache( cacheName );
+		final boolean embeddedInEntity = associatonTypeIsForeignKeyEmbeddedInEntity( key, association, mapper );
+		if ( embeddedInEntity ) {
+			insertOrUpdateAssociationEmbeddedInEntity( key, association, associationContext, mapper, cacheName );
+		}
+		else {
+			insertOrUpdateAssociationMappedAsDedicatedEntries( key, association, associationContext, mapper, cacheName );
+		}
+	}
+
+	private void insertOrUpdateAssociationMappedAsDedicatedEntries(AssociationKey key, Association association, AssociationContext associationContext,
+			ProtoStreamMappingAdapter mapper, String cacheName) {
+		log.debugf( "insertOrUpdateAssociation for key '%s' on cache '%s', mapped as dedicated entries in ad-hoc table", key, cacheName );
+		final List<AssociationOperation> operations = association.getOperations();
+		for ( AssociationOperation ao : operations ) {
+			AssociationOperationType type = ao.getType();
+			RowKey rowKey = ao.getKey();
+			ProtostreamId idBuffer = mapper.createIdPayload( rowKey.getColumnNames(), rowKey.getColumnValues() );
+			switch ( type ) {
+				case PUT:
+					ProtostreamPayload valuePayloadForPut = mapper.createValuePayload( ao.getValue() );
+					mapper.withinCacheEncodingContext( c -> c.put( idBuffer, valuePayloadForPut ) );
+					break;
+				case REMOVE:
+					mapper.withinCacheEncodingContext( c -> c.remove( idBuffer ) );
+					break;
+				case CLEAR:
+					throw new AssertionFailure( "Request for CLEAR operation on an association mapped to dedicated entries. Makes no sense?" );
+			}
+		}
+	}
+
+	private void insertOrUpdateAssociationEmbeddedInEntity(AssociationKey key, Association association, AssociationContext associationContext,
+			ProtoStreamMappingAdapter mapper, String cacheName) {
+		log.debugf( "insertOrUpdateAssociation for key '%s' on cache '%s', mapped as in-entity foreign keys", key, cacheName );
+		final List<AssociationOperation> operations = association.getOperations();
+		for ( AssociationOperation ao : operations ) {
+			AssociationOperationType type = ao.getType();
+			RowKey rowKey = ao.getKey();
+			Tuple sourceTuple = ao.getValue();
+			ProtostreamId idBuffer = mapper.createIdPayload( rowKey.getColumnNames(), rowKey.getColumnValues() );
+			Tuple targetTuple;
+			ProtostreamPayload existingPayload = mapper.withinCacheEncodingContext( c -> c.get( idBuffer ) );
+			if ( existingPayload == null ) {
+				targetTuple = new Tuple();
+			}
+			else {
+				targetTuple = existingPayload.toTuple();
+			}
+			switch ( type ) {
+				case PUT:
+					for ( String columnName : rowKey.getColumnNames() ) {
+						targetTuple.put( columnName, sourceTuple.get( columnName ) );
+					}
+					ProtostreamPayload valuePayloadForPut = mapper.createValuePayload( targetTuple );
+					mapper.withinCacheEncodingContext( c -> c.put( idBuffer, valuePayloadForPut ) );
+					break;
+				case REMOVE:
+					for ( String columnName : key.getColumnNames() ) {
+						targetTuple.remove( columnName );
+					}
+					ProtostreamPayload valuePayloadForRemove = mapper.createValuePayload( targetTuple );
+					mapper.withinCacheEncodingContext( c -> c.put( idBuffer, valuePayloadForRemove ) );
+					break;
+				case CLEAR:
+					throw new AssertionFailure( "Request for CLEAR operation on an association mapped as foreign key embedded an an entity. Makes no sense?" );
+			}
+		}
+	}
+
+	private boolean associatonTypeIsForeignKeyEmbeddedInEntity(AssociationKey key, Association association, ProtoStreamMappingAdapter mapper) {
+		//if a RowKey includes all columns making up the Table's PK then the whole row is to be deleted (A "bridge" table)
+		//Otherwise, tuple operations must be applied to the properties of the Entity
+		List<AssociationOperation> operations = association.getOperations();
+		if ( operations.isEmpty() ) {
+			//Result doesn't matter as there are no operations to apply
+			return false;
+		}
+		else {
+			RowKey rowKey = operations.get( 0 ).getKey();
+			return mapper.columnSetExceedsIdRequirements( rowKey.getColumnNames() );
+		}
+	}
+
+	@Override
+	public void removeAssociation(AssociationKey key, AssociationContext associationContext) {
+		Map<RowKey, Map<String, Object>> rowsMap = loadRowKeysByQuery( key );
+		final String cacheName = key.getTable();
+		log.debugf( "removeAssociation for key '%s' on cache '%s'", key, cacheName );
+		final ProtoStreamMappingAdapter mapper = provider.getDataMapperForCache( cacheName );
+		for ( RowKey rowKey : rowsMap.keySet() ) {
+			ProtostreamId idBuffer = mapper.createIdPayload( rowKey.getColumnNames(), rowKey.getColumnValues() );
+			mapper.withinCacheEncodingContext( c -> c.remove( idBuffer ) ) ;
+		}
+	}
+
+	@Override
+	public boolean isStoredInEntityStructure(AssociationKeyMetadata associationKeyMetadata, AssociationTypeContext associationTypeContext) {
+		return false;
+	}
+
+	@Override
+	public Number nextValue(NextValueRequest request) {
+		return provider.getSequenceHandler().getSequenceValue( request );
+	}
+
+	@Override
+	public void forEachTuple(ModelConsumer consumer, TupleContext tupleContext, EntityKeyMetadata entityKeyMetadata) {
+		final String cacheName = entityKeyMetadata.getTable();
+		ProtoStreamMappingAdapter mapper = provider.getDataMapperForCache( cacheName );
+		VersionedValue<ProtostreamPayload> v = mapper.withinCacheEncodingContext( c -> {
+			//TODO extract the batch constant to an option?
+			try ( CloseableIterator<Entry<Object,MetadataValue<Object>>> iterator
+					= c.retrieveEntriesWithMetadata( null, 100 ) ) {
+				while ( iterator.hasNext() ) {
+					Entry<Object, MetadataValue<Object>> next = iterator.next();
+					ProtostreamPayload obj = (ProtostreamPayload) next.getValue().getValue();
+					long version = next.getValue().getVersion();
+					VersionedTuple tuple = obj.toVersionedTuple();
+					tuple.setVersion( version );
+					consumer.consume( tuple );
+				}
+			}
+			return null;
+		} );
+	}
+
+	@Override
+	public DuplicateInsertPreventionStrategy getDuplicateInsertPreventionStrategy(EntityKeyMetadata entityKeyMetadata) {
+		//We can implement duplicate insert detection by this by using Infinispan's putIfAbsent
+		//support and verifying the return on any insert
+		//TODO Not implemented yet as Hot Rod's support for atomic operations is complex, so default to the naive impl;
+		// In particular:
+		// - CAS operations such as putIfAbsent require Caches to use Transactions on the server (but the Hot Rod client can't participate)
+		// - Multi-Get and Multi-Put operations don't honour the Version API
+		return DuplicateInsertPreventionStrategy.LOOK_UP;
+	}
+
+	@Override
+	public boolean supportsSequences() {
+		//For reasons to keep this to 'false' see implementation comments on HotRodSequenceHandler
+		return false;
+	}
+
+	// [Optional] implement MultigetGridDialect:
+	@Override
+	public List<Tuple> getTuples(EntityKey[] keys, TupleContext tupleContext) {
+		Objects.requireNonNull( keys );
+		if ( keys.length == 0 ) {
+			return Collections.emptyList();
+		}
+		else if ( keys.length == 1 ) {
+			return Collections.singletonList( getTuple( keys[0], tupleContext ) );
+		}
+		else {
+			final String cacheName = keys[0].getTable();
+			final ProtoStreamMappingAdapter mapper = provider.getDataMapperForCache( cacheName );
+			final Map<EntityKey,ProtostreamId> keyConversionMatch = new HashMap<>();
+			final Set<ProtostreamId> convertedKeys = new HashSet<>();
+			for ( EntityKey ek : keys ) {
+				if ( ek == null ) {
+					continue;
+				}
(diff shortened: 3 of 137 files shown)


gunnarmorling:
Hey @Sanne, awesome to see this PR; looking forward to have the remote support in finally!

I was wondering a bit how to approach this PR; do you think you could provide some pointers to the design from a more high-level point of view, important things/classes to look at and things like that? The change is huge so I'd like to make sure we can make the review efficient by getting a good understanding of the overall picture and focus on the most important things. Thanks!

Sanne:
> I was wondering a bit how to approach this PR

You're on holidays, isn't that a good hint on how to approach this? ;-)

But nice idea, I'll write a mini-guide in the PR description.

gunnarmorling:
> I'll write a mini-guide in the PR description

Cool, thanks!

emmanuelbernard:
I'll look at it a bit but do not count on me for a full review. So someone else, please review Sanne's work.

emmanuelbernard:
I think I would prefer to map associations within the entity itself and not as a separate entry by default. First off, it's not possible today due to a limitation of Infinispan's ProtoStream: @Sanne needs to keep pushing the Infinispan team to address the remaining issues (this one and the ability to push schema come to mind).

But assuming we can, we did discuss it a bit with Sanne, but let me try and summarize my points:

By having the association data (ids in this case) within the entity entry, you allow more queries (in a wide sense) to be executed by Infinispan. In particular all of the queries requiring both an entity property and an association state. Queries not involving both can be handled by a separate query to the association cache.
This is particularly acute for `@ElementCollection` which naturally fit within the entity. Embedding the state is how Teiid maps things into Infinispan (see Van).

Granted, you need the option to store the association state in a separate entry (like we do for Mongo) in case the entry becomes big. We should reuse the existing option annotations.

By Infinispan query, I mean all means of getting the data except look up: dist exec, distributed streams, spark, Infinispan query, Full-text search etc.

TODO: capture that discussion in a JIRA to execute on it once Infinispan fix the limitations.

emmanuelbernard:
== Bags & unordered collections

The way you describe the problem made me wonder how you store this info @Sanne.
Is there one association cache for all possible associations? Or one association cache per association type.
And that question aside, how do you store entries? one key per entry in the association or rather one key per association and the value contains a collection of entries?
If the former, how do you retrieve all elements of an association (say to satisfy a a.`getBs()` call)?
I imagine in this case that you could give synthetic UUID keys as neither order matter and Hibernate ORM does ask to remove all entries matching the columns (i.e. all the duplicates) and recreates the expected amount of duplicates to match the new state. For example if you have 3 times `[nick:joe, age:18]` and remove one, Hibernate ORM asks to remove all entries containing `[nick:joe, age:18]` and then asks you to add two entries `[nick:joe, age:18]`. Additions of course are not a problem. So you could walk over all UUID keys and remove the duplicates by value.

If the latter (value is a collection of entries), you can apply the same logic.

gunnarmorling:
> Sanne added some commits on *18 Dec 2015*

Holy moly ... :)

gunnarmorling:
> I think I would prefer to map associations within the entity itself and not as a separate entry by default.

Same here. In fact I think we even agreed on this already at one of the previous meetings?!

Given ProtoBuf allows for the same sort of nesting as JSON/BSON does, the mapping really could and should like we do it for the document stores (MongoDB).

emmanuelbernard:
## Versioning and Optimistic locking

I remember you mentioned some specific feature that would benefit from this during our discussion, but I forgot. It was not about the exception vs double read for tuple creation. But something like that.

Sanne:
@gunnarmorling 

>> I think I would prefer to map associations within the entity itself and not as a separate entry by default.

> Same here. In fact I think we even agreed on this already at one of the previous meetings?!

Yes you all told me your preference, and I tried, and I'm sure I already explained several times that this wasn't going to work out, at least not without further patches in Infinispan.

To refresh your memory: the problem is not Protobuf, but ProtoStream. I switched 3 times between mapping strategies, also trying to bypass ProtoStream to accomodate for your mapping wishes, but ultimately the Infinispan team strongly suggested to stick to ProtoStream so I did fall back again to the Infinispan approach; The protostream library is meant to map user code directly and requires 1:1 mapping between encoders for a message and class model, so I can't nest things freely as one could by writing a Protobuf message with native APIs. On the other hand, Protobuf is meant to use code generation taking the schema as an input so we don't have many options.

N.B. each change in approach to follow "outsider suggestions" costed me weeks of work; the code now looks quite "clean" I hope.. or at least I tried hard for that, but this approach really is the only survivor of many attempts. I regularly shared the reasons of failed approaches so I'm not too happy to see you all shocked by the taken approach now.

All considered, I also think the strategy I adopted is safer for a general purpose default. I see the value in the other mapping strategy, but that needs to be controlled by an option so that implies you need to either have both options, or only the one I implemented. (The problem I'm thinking of is with huge collections: embedding it all wouldn't allow people to control loading with options such as ExtraLazy collection mappings ).

Having hopefully clear that the implemented mapping is needed too - at least as an option -, I see two possible evolutions to implement the second option (nesting of collections):
 - evolve ProtoStream (by either waiting for the Infinispan team or helping out)
 - bypass ProtoStream yet carefully mimic all of its storage idiosyncrasies (it is quite peculiar)

I suspect the best value is in evolving ProtoStream, yet that implies we limit our compatibility to some future release of Infinispan which has quite an impact on our roadmap plans.

So these are my consideration for having chosen the current mapping.

Sanne:
@emmanuelbernard RE 
## Bags & unordered collections

Good point I think your UUID idea could work; I opened the following issue to track this as a follow-up:

https://hibernate.atlassian.net/browse/OGM-1168

As suggested on the issue, I also think that unerdered bags should be stored embedded, at least by default. Maybe we should even take the option of alternative storage away in this case, as UUIDs can be costly and I suspect a bag would always need to be loaded as a whole anyway (hard to pick a subset of it).

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protostream/OgmProtoStreamMarshaller.java
```diff
@@ -23,9 +23,7 @@
  */
 public final class OgmProtoStreamMarshaller extends ProtoStreamMarshaller {
 
-	//FIXME all static threadlocals should die
-	// See https://hibernate.atlassian.net/browse/OGM-1163
-	private static final ThreadLocal<SerializationContext> currentSerializationContext = new ThreadLocal<>();
+	private final ThreadLocal<SerializationContext> currentSerializationContext = new ThreadLocal<>();
```
emmanuelbernard: Why do you still need a thread local? `SerializationContext` is not concurrent safe?

Sanne:
@emmanuelbernard RE
## Versioning and Optimistic locking

as mentioned in the PR description, some usage of Versioning was implemented but ultimately commented out as it needs to make sure the Association isn't deleted and then re-inserted when updated, as I lose track of version numbers at that point.

So this version does not implement optimistic locking.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protostream/OgmProtoStreamMarshaller.java
Sanne: There's only one instance of `OgmProtoStreamMarshaller` for the whole Hot Rod Client.

That means all our parallel Session are using the same marshaller - yet concurrently I need it to instruct which schema to use for a specific IO operation.

A typical Infinispan client app would have a different Class registered in the marshaller for each entity; we don't have this luxury, or I'd have to bytecode-generate a Class definition for each "Table" we want to map.

bom/pom.xml
```diff
@@ -508,6 +521,52 @@
...
+            </dependency>
+            <dependency>
+                <groupId>org.infinispan</groupId>
+                <artifactId>infinispan-client-hotrod</artifactId>
+                <version>${infinispanVersion}</version>
+                <type>test-jar</type>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.infinispan</groupId>
```
emmanuelbernard: why do we need the server? To run it during tests?

Sanne: Yes: see RemoteHotRodServerRule

core/src/main/java/org/hibernate/ogm/datastore/impl/DatastoreProviderType.java
```diff
@@ -16,6 +16,7 @@
 public enum DatastoreProviderType {
 	MAP( "org.hibernate.ogm.datastore.map.impl.MapDatastoreProvider" ),
 	INFINISPAN( "org.hibernate.ogm.datastore.infinispan.impl.InfinispanDatastoreProvider" ),
```
emmanuelbernard: Deprecate and add `INFINISPAN_EMBEDDED` ?

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
```diff
@@ -0,0 +1,396 @@
...
+import org.infinispan.commons.util.CloseableIterator;
+import org.infinispan.query.dsl.FilterConditionContext;
+import org.infinispan.query.dsl.Query;
+import org.infinispan.query.dsl.QueryBuilder;
+import org.infinispan.query.dsl.QueryFactory;
+
+/**
+ *  Some implementation notes for evolution:
+ *
+ *  - QueryableGridDialect can't be implemented as "native queries" in Hot Rod are DSL based
```
emmanuelbernard: aren't the supported set of the Infinispan DSL anways a subset of JP-QL? We could ask people to use JP-QL anyways.

core/src/main/java/org/hibernate/ogm/datastore/impl/DatastoreProviderType.java
Sanne: Ok: https://hibernate.atlassian.net/browse/OGM-1169

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
Sanne: The IQL is not exposed yet, it will be in future versions (?)

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
```diff
@@ -0,0 +1,396 @@
...
+ *    and have no String representation; this might change as Hot Rod exposes the underlying
+ *    query representation which is similar to HQL; alternatively we could look at exposing
+ *    native queries in some way other than a String.
+ *
+ *  - OptimisticLockingAwareGridDialect can't be implemented as it requires an atomic replace
+ *    operation on a subset of the columns, while atomic operations in Hot Rod have to involve
+ *    either the full value or the version metadata of Infinispan's VersionedValue.
```
emmanuelbernard: Can you explain what `VersionedValue` is for the uninitiated?

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protostream/OgmProtoStreamMarshaller.java
emmanuelbernard: thanks, please add this as a comment in the code.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
emmanuelbernard: yes but from the parser tree, you can call Infinispan DSL to build the query in the mean time. Granted it's not string -> string, it's string -> DSL object

Sanne: Yes we could do that: not a technical issue. My concern is that we'd essentially be exposing IQL already before the datastore advertises (and stabilises) it?

Until Infinispan actually exposes it, you'd have OGM expose a capability which the store itself hasn't even yet documented.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
```diff
@@ -0,0 +1,396 @@
...
+import org.infinispan.query.dsl.QueryBuilder;
+import org.infinispan.query.dsl.QueryFactory;
+
+/**
+ *  Some implementation notes for evolution:
+ *
+ *  - QueryableGridDialect can't be implemented as "native queries" in Hot Rod are DSL based
+ *    and have no String representation; this might change as Hot Rod exposes the underlying
+ *    query representation which is similar to HQL; alternatively we could look at exposing
+ *    native queries in some way other than a String.
```
DavideD: I  don't think QueryableGridDialect expect a String representation, you can actually use a different object. For example, in MongoDB we have `MongoDBQueryDescriptor`.

I'm not saying that you have to implement it now, but I think the it can be done.

emmanuelbernard:
> To refresh your memory: the problem is not Protobuf, but ProtoStream. I switched 3 times between mapping strategies, also trying to bypass ProtoStream to accomodate for your mapping wishes, but ultimately the Infinispan team strongly suggested to stick to ProtoStream so I did fall back again to the Infinispan approach; The protostream library is meant to map user code directly and requires 1:1 mapping between encoders for a message and class model, so I can't nest things freely as one could by writing a Protobuf message with native APIs. On the other hand, Protobuf is meant to use code generation taking the schema as an input so we don't have many options.
> 
> N.B. each change in approach to follow "outsider suggestions" costed me weeks of work; the code now looks quite "clean" I hope.. or at least I tried hard for that, but this approach really is the only survivor of many attempts. I regularly shared the reasons of failed approaches so I'm not too happy to see you all shocked by the taken approach now.
> 
> All considered, I also think the strategy I adopted is safer for a general purpose default. I see the value in the other mapping strategy, but that needs to be controlled by an option so that implies you need to either have both options, or only the one I implemented. (The problem I'm thinking of is with huge collections: embedding it all wouldn't allow people to control loading with options such as ExtraLazy collection mappings ).

Don't get annoyed. Remember that we have not lived daily on this topic nor had weekly status update, peek. Assume in our discussions that we have forgotten *everything* (which is pretty much my case anyways).

We keep pushing for this because from a philosophy approach, we want OGM to store data like a user would have. And a user would have embedded the information.
So yes we can release this version as is but also we should push for a ProtoStream "dynamic" mode (my preferred approach over the alternatives you proposed).
How long it would take us to implement it ourselves for example (say asking a Yoann to go a run with it)?

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
emmanuelbernard: Well, we do expose *JP-QL*. That's our job :)

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
```diff
@@ -0,0 +1,396 @@
...
+
+	@Override
+	public Tuple createTuple(EntityKey key, TupleContext tupleContext) {
+		return new VersionedTuple( true );
+	}
+
+	@Override
+	public void insertOrUpdateTuple(EntityKey key, Tuple tuple, TupleContext tupleContext) {
+		VersionedTuple versionedTuple = (VersionedTuple) tuple;
+		final long knownVersion = versionedTuple.getVersion();
```
DavideD: The code that it is using this varibale is commented

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
Sanne: A read from Hot Rod:

    VersionedValue<ProtostreamPayload> v = client.getVersioned( key );
    ProtostreamPayload payload = v.getValue();
    long version = v.getVersion();

A write:

     boolean successfull = c.replaceWithVersion( key, payload, expectedVersion )

Alternatively, you can do a full comparison of the payload value:


     boolean success = c.replace( key, oldValue, newValue )

In this case I like the Infinispan API, but I don't think it matches the specific intent of the `OptimisticLockingAwareGridDialect` contract so optimistic locking would have to be implemented differently.

I'm not sure I understood the working of `OptimisticLockingAwareGridDialect` though. For example, it doesn't seem to have an "insertUnlessExists" semantics.

Sanne: P.S. the critical difference is that the version is auto-incremented by Hot Rod, never explicitly written by the client.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
Sanne: Which API is that .. I'm looking at

`org.hibernate.ogm.dialect.query.spi.QueryableGridDialect.parseNativeQuery(String)`

Also conceptually, the public exposed API is `org.hibernate.SharedSessionContract.createSQLQuery(String)`, right?

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
Sanne: thanks. It's needed by the block of code which is commented out below.. I'll remove all commented code and this.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
Sanne: Could you open a JIRA yourself in this case? I'm not sure I understood what you're proposing.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
DavideD: Yes, you are right, my bad.

I thought we had some way to pass something different since in `OgmSession` we have a `createNativeQuery` method but it is the same thing as `createSQLQuery`.

Sanne:
I've re-pushed it to rebase and incorporate some of the feedback so far; especially improving comments and removed some dead code.

Had to disable the new tests GridDialectOperationInvocationForOneToOneTest and GridDialectOperationInvocationForElementCollectionTest as they fail after rebase.. I'll have to look at that.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
emmanuelbernard: https://hibernate.atlassian.net/browse/OGM-1171

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/InfinispanRemoteDatastoreProvider.java
```diff
@@ -0,0 +1,192 @@
...
+		} );
+		if ( failedCacheNames.size() > 1 ) {
+			throw log.expectedCachesNotDefined( failedCacheNames );
+		}
+		else if ( failedCacheNames.size() == 1 ) {
+			throw log.expectedCacheNotDefined( failedCacheNames.iterator().next() );
+		}
+	}
+
+	private void setMappedCacheNames(Set<String> tableNames) {
```
emmanuelbernard: Never used.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/InfinispanRemoteDatastoreProvider.java
```diff
@@ -0,0 +1,192 @@
...
+
+	//Useful to allow people to dump the generated schema,
+	//we use it to capture the schema in tests too.
+	private SchemaCapture schemaCapture;
+
+	private ServiceRegistryImplementor serviceRegistry;
+
+	private SchemaOverride schemaOverrideService;
+
+	private Set<String> mappedCacheNames;
```
emmanuelbernard: Not important but I think `mappedCacheNames` can be removed and one can use `perCacheSchemaMappers.keySet()`

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
DavideD: I think the idea of this interface is to allow dialects without a native versioning system to still support optimistic locking using atomic find and update operations.

I suspect you can still implemente the insert in the uipdate method but I would need to check this.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/InfinispanRemoteDatastoreProvider.java
Sanne: `perCacheSchemaMappers.keySet()` is a subset of all mapped caches: we also use some for e.g. Sequences

Yet, I think you're right that I could change that: nothing would break as this is largely used just by testing utilities which don't care for Sequences.. but I like the explicitness of the Set in an explicit field, useful during debugging.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/InfinispanRemoteDatastoreProvider.java
Sanne: thanks, removing.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java
```diff
@@ -0,0 +1,52 @@
...
+ * Properties for configuring the Infinispan Remote datastore via {@code persistence.xml} or
+ * {@link StandardServiceRegistryBuilder}.
+ */
+public final class InfinispanRemoteProperties implements KeyValueStoreProperties {
+
+	/**
+	 * The configuration property to use as key to define a custom configuration resource
+	 * for the Hot Rod (Infinispan remote) client.
+	 */
+	public static final String CONFIGURATION_RESOURCE_NAME = "hibernate.ogm.infinispan_remote.configuration_resource_name";
```
emmanuelbernard: Does that means the generic host / port won't work and someone must provide some xml configuration file? Or is that optional? It might be good to guide users in the JavaDoc or maybe better doc as to what to prefer.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java
```diff
@@ -0,0 +1,52 @@
...
+	 * Useful for testing, or to dump the schema somewhere else.
+	 */
+	public static final String SCHEMA_CAPTURE_SERVICE = "hibernate.ogm.infinispan_remote.schema_capture_service";
+
+	/**
+	 * You can inject an instance of {@link org.hibernate.ogm.datastore.infinispanremote.spi.schema.SchemaOverride} into
+	 * the configuration properties to override the Protobuf schema being generated.
+	 * This will not affect how entities are encoded, so the alternative schema must be compatible.
+	 */
+	public static final String SCHEMA_OVERRIDE_SERVICE = "hibernate.ogm.infinispan_remote.schema_override_service";
```
emmanuelbernard: Since it's such an advanced feature, should we make it private?

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protostream/ProtostreamPayload.java
```diff
@@ -0,0 +1,98 @@
...
+		this.loadedSnapshot = Objects.requireNonNull( loadedSnapshot );
+		this.tuple = null;
+	}
+
+	public ProtostreamPayload(Tuple tuple) {
+		this.tuple = Objects.requireNonNull( tuple );
+		this.loadedSnapshot = null;
+	}
+
+	public void forEach(BiConsumer<String, Object> actionOnNamedTouples) {
```
emmanuelbernard: s/touple/tuple/

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protostream/ProtostreamPayload.java
```diff
@@ -0,0 +1,98 @@
...
+		if ( loadedSnapshot != null ) {
+			loadedSnapshot.getMap().forEach( actionOnNamedTouples );
```
emmanuelbernard: what does that do? Update the snapshot? Snapshots are supposed to be immutable I think.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
```diff
@@ -0,0 +1,380 @@
...
+
+	public InfinispanRemoteDialect(InfinispanRemoteDatastoreProvider provider) {
+		this.provider = Objects.requireNonNull( provider );
+	}
+
+	@Override
+	public Tuple getTuple(EntityKey key, TupleContext tupleContext) {
+		final String cacheName = key.getTable();
+		ProtoStreamMappingAdapter mapper = provider.getDataMapperForCache( cacheName );
+		ProtostreamId idBuffer = mapper.createIdPayload( key.getColumnNames(), key.getColumnValues() );
```
emmanuelbernard: Why is it called an idBuffer exactly?

emmanuelbernard:
For people who wonder while processing the PR, associations are stored in a cache by their table name. And each row is an full entry in the cache. So when loading an association, you essentially do a query. 

TODO @sanne do one to many clash with the owning entity table?
TODO That is very different from what we do for the other backends including Infinispan proper. Discuss with voice together to see if we are fine with that.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
```diff
@@ -0,0 +1,380 @@
...
+			switch ( type ) {
+				case PUT:
+					ProtostreamPayload valuePayloadForPut = mapper.createValuePayload( ao.getValue() );
+					mapper.withinCacheEncodingContext( c -> c.put( idBuffer, valuePayloadForPut ) );
+					break;
+				case REMOVE:
+					mapper.withinCacheEncodingContext( c -> c.remove( idBuffer ) );
+					break;
+				case CLEAR:
+					throw new AssertionFailure( "Request for CLEAR operation on an association mapped to dedicated entries. Makes no sense?" );
```
emmanuelbernard: @sanne I think you got the wrong picture. It's not a clear of the association table. It's a clear of all the rowkey for that specific association entry (i.e. the assoc info dedicated to a specific entity).

So it would clear all entries returned by the query done in `getAssociation` (that's virtual of course and you can rely on the snapshot for that list).

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
```diff
@@ -0,0 +1,380 @@
...
+	public Association createAssociation(AssociationKey key, AssociationContext associationContext) {
+		Map<RowKey, Map<String, Object>> associationMap = new HashMap<RowKey, Map<String,Object>>();
+		return new Association( new MapAssociationSnapshot( associationMap ) );
+	}
+
+	@Override
+	public void insertOrUpdateAssociation(AssociationKey key, Association association, AssociationContext associationContext) {
+		final String cacheName = key.getTable();
+		final ProtoStreamMappingAdapter mapper = provider.getDataMapperForCache( cacheName );
+		final boolean embeddedInEntity = associatonTypeIsForeignKeyEmbeddedInEntity( key, association, mapper );
```
emmanuelbernard: What does that mean exactly? The many to one case? or rather `AssociationType.ONE_TO_ONE`

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
```diff
@@ -0,0 +1,380 @@
...
+				case REMOVE:
+					mapper.withinCacheEncodingContext( c -> c.remove( idBuffer ) );
+					break;
+				case CLEAR:
+					throw new AssertionFailure( "Request for CLEAR operation on an association mapped to dedicated entries. Makes no sense?" );
+			}
+		}
+	}
+
+	private void insertOrUpdateAssociationEmbeddedInEntity(AssociationKey key, Association association, AssociationContext associationContext,
```
emmanuelbernard: In this case if I understand correctly, you literally store all of the associaton columns (fk?) flat in the object.

```
User {
   String name;
   String addressId1;
   String addressId2;
}
// assuming Address is a composite key of id1 and id2
```

Correct?

In mongodb, we do create an intermediary object, something like

```
```
User {
   String name;
   AddressId address {
       String addressId1;
       String addressId2;
   }
}
// assuming Address is a composite key of id1 and id2
```

PS @gunnarmorling, I am correct on MongoDB right?

Comment (gunnarmorling) [inline infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java]: Yes, for MongoDB a composite key would be [mapped into a sub-node](http://docs.jboss.org/hibernate/ogm/5.0/reference/en-US/html_single/#identifiers_3) in the document:

    {
      "_id" : {
          "author" : "Guillaume",
          "title" : "How to use Hibernate OGM ?"
      },
      "content" : "Simple, just like ORM but with a NoSQL database"
    }

**Comment by gunnarmorling** [comment]
> need to discuss to have a team understanding of the choices.

I suggest that we allocate an hour or so then in tomorrow morning's hangout for this.

File: infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java (hunk 10 above)
Comment (Sanne) [inline infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java]: It's not an XML file but a properties file.
This file is mandatory to configure the Hot Rod client.

E.g. this is the one we use for testing:
 - https://github.com/Sanne/hibernate-ogm/blob/5f93b5bcf8974c412233ff260e817dfaa80d0062/infinispan-remote/src/test/resources/hotrod-client-testingconfiguration.properties

We could read such properties from the Hibernate properties? Last time I asked, you suggested to keep the external resource.

There's a follow up JIRA open to strictly validate the properties define in there, as we can't deal with some options being set.

File: infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java (hunk 11 above)
Comment (Sanne) [inline infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java]: I'm open to suggestions, but it seemed useful for end users.
They can use it to:

1. dump the generated schema to something else (a text file?)
2. intercept the generation and patch it before it's deployed
3. override the generation completely with their own resource

Of course no guarantees that the dialect will still work with a different schema than expected..

Comment (Sanne) [inline infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java]: Obviously I created this for my own testing needs; tried to make it re-usable as secondary goal only.

File: infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java (hunk 16 above)
Comment (Sanne) [inline infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java]: Both of them, if they are the owners of the relation: as long as the foreign key to the linked entity is stored within this same table, and not in an external bridge table.

File: infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java (hunk 17 above)
Comment (Sanne) [inline infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java]: Yes the storage is flat: remember I can't store nested structures.

File: infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java (hunk 10 above)
Comment (emmanuelbernard) [inline infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java]: @gunnarmorling for the other drivers like Neo or MongoDB don't we allow to pass through properties to the driver? We probably could align Infinispan's property injection as well by providing it a map from our own properties instead of using a file.

Comment (emmanuelbernard) [inline infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java]: @Sanne ok, I think I disagree with younger @emmanuelbernard right now ;) At least we should do what we do for the other dialects.

File: infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java (hunk 17 above)
Comment (emmanuelbernard) [inline infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java]: Arrrrrrggggggggg, fuck ProtoStream and its flat structure limitation!

File: infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java (hunk 10 above)
Comment (gunnarmorling) [inline infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java]: > for the other drivers like Neo or MongoDB don't we allow to pass through properties to the driver?

Yes, that's right. For MongoDB we all to pass through props like so:

    hibernate.ogm.mongodb.driver.serverSelectionTimeout=1000

File: infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java (hunk 11 above)
Comment (gunnarmorling) [inline infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java]: I'd say let's expose as few options as possible. At the very least mark it with `@Experimental`.

File: infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protostream/ProtostreamPayload.java (hunk 12 above)
Comment (gunnarmorling) [inline infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protostream/ProtostreamPayload.java]: Or toupet ;)

File: infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java (hunk 17 above)
Comment (anistor) [inline infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java]: @sanne I did not follow all the comments in this conversation, so why cannot you store nested structures? Protobuf/Protostream do allow it.

File: infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protobuf/BigDecimalProtofieldWriter.java [hunk 18]
```diff
@@ -0,0 +1,43 @@
...
+ * A BigDecimal is encoded as a byte array after converting to a BigInteger.
+ * The byte array conversion of BigInteger is well defined.
+ *
+ * @see java.math.BigInteger#toByteArray
+ */
+public class BigDecimalProtofieldWriter extends BaseProtofieldWriter<BigDecimal> implements ProtofieldWriter<BigDecimal> {
+
+	public BigDecimalProtofieldWriter(int tag, String name, boolean nullable, String columnName) {
+		super( tag, name, nullable, columnName,
+				(ProtoStreamWriter outProtobuf, BigDecimal value) -> outProtobuf.writeBytes( name, value.toBigInteger().toByteArray() ),
```
gunnarmorling: The Lambdas look nice, but out of curiousity, is there any strong advantage over simply foreseeing methods in `ProtofieldWriter` to be implemented in the specific sub-classes? Just trying to figure when to use one over the other.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
```diff
@@ -0,0 +1,380 @@
...
+ * @author Sanne Grinovero
+ */
+public class InfinispanRemoteDialect<EK,AK,ISK> extends BaseGridDialect implements MultigetGridDialect {
+
+	private static final Log log = LoggerFactory.getLogger();
+
+	private final InfinispanRemoteDatastoreProvider provider;
+
+	public InfinispanRemoteDialect(InfinispanRemoteDatastoreProvider provider) {
+		this.provider = Objects.requireNonNull( provider );
```
gunnarmorling: No need to change from my side, but there is the `Contracts` utility which we've used for this in OGM so far.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protobuf/CompositeProtobufCoDec.java
```diff
@@ -0,0 +1,185 @@
...
+import org.hibernate.ogm.datastore.infinispanremote.impl.protostream.ProtostreamPayload;
+import org.hibernate.ogm.datastore.map.impl.MapTupleSnapshot;
+import org.hibernate.ogm.dialect.spi.TupleContext;
+import org.hibernate.ogm.model.key.spi.EntityKey;
+import org.hibernate.ogm.model.spi.Tuple;
+import org.infinispan.client.hotrod.RemoteCache;
+import org.infinispan.protostream.MessageMarshaller.ProtoStreamReader;
+import org.infinispan.protostream.MessageMarshaller.ProtoStreamWriter;
+
+public final class CompositeProtobufCoDec implements MainOgmCoDec {
```
gunnarmorling: Why is it a "composite" codec? And why is it split into interface and impl?

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/InfinispanRemoteDatastoreProvider.java
```diff
@@ -0,0 +1,195 @@
...
+
+	//For each cache we have a schema and a set of encoders/decoders to the generated protobuf schema
+	private Map<String,ProtoDataMapper> perCacheSchemaMappers;
```
Sanne: Nice, I wasn't aware. Will add this to several other locations too.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protobuf/EnumProtofieldWriter.java
```diff
@@ -0,0 +1,136 @@
...
+ * String to map it to native enums.
+ * See also: https://developers.google.com/protocol-buffers/docs/proto#enum
+ * @author Sanne Grinovero
+ */
+public class EnumProtofieldWriter implements ProtofieldWriter<Enum> {
+
+	private final int tag;
+	private final String name;
+	private final Class<? extends Enum> type;
+	private final Enum[] enumConstants;
```
gunnarmorling: Seems unused.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protostream/ProtoDataMapper.java
```diff
@@ -0,0 +1,92 @@
...
+
+public final class ProtoDataMapper implements ProtoStreamMappingAdapter, ProtostreamAssociationMappingAdapter {
+
+	private final SerializationContext serContext;
+	private final MainOgmCoDec delegate;
+	private final OgmProtoStreamMarshaller marshaller;
+
+	public ProtoDataMapper(MainOgmCoDec delegate, SerializationContext serContext, OgmProtoStreamMarshaller marshaller) throws DescriptorParserException, IOException {
+		this.marshaller = marshaller;
+		this.delegate = Objects.requireNonNull( delegate );
```
gunnarmorling: Here we have access to the actual marshaller and the serialization context. Couldn't this be used to initialize the marshaller for once with the right context instead of setting this via the thread local again and again?

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
Sanne: @anistor hey, I was going to ping you about this ;)
Remember our conversation in London? I'm constrained by the Marshaller API to require a 1:1 relation between class types and protobuf message names. We agreed we'd need to evolve that API, but in the meantime I decided to go ahead and implement this dialect w/o supporting nested structures yet. We'll talk again about this in Konstanz soon..

Sanne:
I've been applying the package renames and class renames as suggested, so my next diff will look terrible. Closing this PR..

Thanks everyone! Reopening soon on a fresh branch.

infinispan-remote/src/test/resources/protoschema-expectations/Cloud_SnowFlake.protobuf
```diff
@@ -0,0 +1,40 @@
...
+	required string Cloud_id = 1;
+	required string producedSnowFlakes_id = 2;
+}
+
+message Cloud_id {
+	required string id = 1;
+}
+
+message Cloud {
+	required string id = 1;
```
gunnarmorling: We may have discussed it before, but did we decide on whether to include the id in the payload again or not? Is there a strong reason to do so?

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protobuf/BigDecimalProtofieldWriter.java
Sanne: I think it's just verbosity, but there's also the aspect of the "name" variable being captured in the lambda bytecode. It effectively becomes a constant.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
Sanne: This is the Java standard way; it wasn't available before Java 8

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
Sanne: @DavideD I agree. And since Infinispan has native support for it, I'll not implement the interface for the fallback option.

But also I'm having some concerns on the protocol proposed by `OptimisticLockingAwareGridDialect`.. I'm not sure that comparing all property values is correct as you might still fail to detect some forms of concurrent writes.

PR #767, opened 2016-09-12
OGM-1064 Operation grouping

gsmet:
So... looks like I finally got something in shape.

I tried to separate the commits in logical units even if it's probably not perfect. I spent quite some time splitting, rebasing and so on so I hope it will help the review.

I implemented grouping for:
- Redis Json & Hash
- CouchDB (I don't think it brings us much due to the fact that we have generated properties but it was interesting as I found a few bugs)
and improved the existing batching for:
- MongoDB

I haven't touched Neo4j and Infinispan as there was a lot of work going on there in parallel.

### ~~OgmEntityEntryState and inverse associations~~

~~I had to comment one test:
ManyToOneGlobalTest>ManyToOneTest.testRemovalOfTransientEntityWithAssociation~~

~~I'll send a mail about that to the list shortly so that we can discuss it.~~

-> **Fixed, we now remove the navigational information when we delete an entity.**

### Properties generated by the datastore

I did what we discuss the other day: if there are generated properties, I flush the pending operations to the datastore.

In the case of CouchDB, as we usually have a generated property for the revision, the grouping is mostly useless...

Funny enough, I found a bug in the new CouchDBDialect and once fixed, the initial issue goes away. The fact is that the revision is updated in the Tuple when we save the document to CouchDB - the protocol returns the updated revision.

I don't think we can make the difference between "the only generated property is the revision" and "we have other generated properties" so, even if suboptimal, it's probably the best we can do.

### OGM-1156

See https://hibernate.atlassian.net/projects/OGM/issues/OGM-1156 . My patch mitigates this issue *except* if there are generated properties (so in the case of CouchDB, always...). I think we should probably try to improve the situation anyway.

### OgmEntityExtraState lifecycle

We need to decide what should be the lifecycle of the values stored in the OgmEntityExtraState.

### Share more code

Once we agree on the general approach, we should probably try to share more code between the grouping aware dialects.

I suspect Infinispan and Neo4j will bring different approaches to the plate so I would like to wait for them to be grouping aware.

### Merge BatchableGridDialect and GroupingByEntityDialect

They started as 2 different beasts but they now have the same behavior so I think we should probably merge them.

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/dialect/batch/spi/OperationsQueue.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/batch/spi/OperationsQueue.java
@@ -6,8 +6,10 @@
  */
 package org.hibernate.ogm.dialect.batch.spi;
 
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
+import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
 
@@ -23,6 +25,7 @@ import org.hibernate.ogm.util.impl.LoggerFactory;
  *
  * @author Guillaume Scheibel &lt;guillaume.scheibel@gmail.com&gt;
  * @author Davide D'Alto &lt;davide@hibernate.org&gt;
+ * @author Guillaume Smet
  */
 public class OperationsQueue {
 
@@ -40,14 +43,34 @@ public class OperationsQueue {
 
 	private final Queue<Operation> operations = new LinkedList<Operation>();
 
-	private final Set<EntityKey> entityKeys = new HashSet<EntityKey>();
+	private final Map<EntityKey, GroupedChangesToEntityOperation> groupedOperations = new HashMap<>();
+
+	private final Set<EntityKey> insertionQueue = new HashSet<>();
 
 	private boolean closed = false;
 
+	public OperationsQueue() {
+	}
+
 	public void add(InsertOrUpdateTupleOperation operation) {
+		add( (GroupableEntityOperation) operation );
+		insertionQueue.add( operation.getEntityKey() );
+	}
+
+	public void add(GroupableEntityOperation operation) {
 		validate();
-		entityKeys.add( operation.getEntityKey() );
-		addOperation( operation );
+		GroupedChangesToEntityOperation groupedOperation = getOrCreateGroupedChangesOnEntityOperation( operation.getEntityKey() );
+		groupedOperation.addOperation( operation );
+	}
+
+	private GroupedChangesToEntityOperation getOrCreateGroupedChangesOnEntityOperation(EntityKey entityKey) {
+		GroupedChangesToEntityOperation groupedOperation = groupedOperations.get( entityKey );
+		if ( groupedOperation == null ) {
+			groupedOperation = new GroupedChangesToEntityOperation( entityKey );
+			groupedOperations.put( entityKey, groupedOperation );
+			addOperation( groupedOperation );
+		}
+		return groupedOperations.get( entityKey );
 	}
 
 	public void add(Operation operation) {
@@ -72,8 +95,9 @@ public class OperationsQueue {
 	}
 
 	public void clear() {
-		entityKeys.clear();
+		groupedOperations.clear();
 		operations.clear();
+		insertionQueue.clear();
 	}
 
 	public void close() {
@@ -89,8 +113,8 @@ public class OperationsQueue {
 	 * @param key the {@link EntityKey} that identify the element
 	 * @return true if an {@link InsertOrUpdateTupleOperation} is bound to the key, false otherwise
 	 */
-	public boolean contains(EntityKey key) {
-		return entityKeys.contains( key );
+	public boolean isInTheInsertionQueue(EntityKey key) {
+		return insertionQueue.contains( key );
 	}
 
 	/**
--- a/core/src/main/java/org/hibernate/ogm/dialect/batch/spi/OperationsQueue.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/batch/spi/OperationsQueue.java
@@ -52,15 +52,35 @@ public class OperationsQueue {
 	public OperationsQueue() {
 	}
 
-	public void add(InsertOrUpdateTupleOperation operation) {
-		add( (GroupableEntityOperation) operation );
+	public void add(Operation operation) {
+		validate();
+
+		log.debugf( "Add batched operation %1$s", operation );
+
+		if ( operation instanceof InsertOrUpdateTupleOperation ) {
+			addInsertOrUpdateTupleOperation( (InsertOrUpdateTupleOperation) operation );
+		}
+		else if ( operation instanceof GroupableEntityOperation ) {
+			addGroupableEntityOperation( (GroupableEntityOperation) operation );
+		}
+		else {
+			addOperation( operation );
+		}
+	}
+
+	private void addInsertOrUpdateTupleOperation(InsertOrUpdateTupleOperation operation) {
+		addGroupableEntityOperation( operation );
 		insertionQueue.add( operation.getEntityKey() );
 	}
 
-	public void add(GroupableEntityOperation operation) {
+	private void addGroupableEntityOperation(GroupableEntityOperation operation) {
 		validate();
 		GroupedChangesToEntityOperation groupedOperation = getOrCreateGroupedChangesOnEntityOperation( operation.getEntityKey() );
 		groupedOperation.addOperation( operation );
+	}
+
+	private void addOperation(Operation operation) {
+		operations.add( operation );
 	}
 
 	private GroupedChangesToEntityOperation getOrCreateGroupedChangesOnEntityOperation(EntityKey entityKey) {
@@ -73,20 +93,10 @@ public class OperationsQueue {
 		return groupedOperations.get( entityKey );
 	}
 
-	public void add(Operation operation) {
-		validate();
-		addOperation( operation );
-	}
-
 	private void validate() {
 		if ( isClosed() ) {
 			throw log.closedOperationQueue();
 		}
-	}
-
-	private void addOperation(Operation operation) {
-		log.debugf( "Add batched operation %1$s", operation );
-		operations.add( operation );
 	}
 
 	public Operation poll() {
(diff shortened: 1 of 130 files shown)


DavideD:
> They started as 2 different beasts but they now have the same behaviour so I think we should probably > merge them.

Yes, it seems to make sense to have only one.  The`BatchableGridDialect` was introduce to exploit the native batching mechanism of some datastore (like MongoDB), no reason for using it when availble with the collected operation (I think).

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -387,29 +386,30 @@ private static void addSubQuery(String operator, BasicDBObject query, String col
...
+		subQuery.removeField( column );
+		if ( subQuery.isEmpty() ) {
+			query.removeField( operator );
 		}
 	}
 
 	@Override
+	public void insertOrUpdateTuple(EntityKey key, TuplePointer tuplePointer, TupleContext tupleContext) {
+		throw new UnsupportedOperationException( "Method not supported in GridDialect anymore" );
+	}
```
DavideD: Can we move this in `AbstractGroupingByEntityDialect` as a default implementation?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1378,15 +1373,64 @@ private static void flushInserts(Map<DBCollection, BatchInsertionTask> inserts)
...
 	}
 
 	private static WriteConcern getWriteConcern(AssociationContext associationContext) {
 		return associationContext.getAssociationTypeContext().getOptionsContext().getUnique( WriteConcernOption.class );
 	}
 
-	private static ReadPreference getReadPreference(TupleContext tupleContext) {
-		return tupleContext.getOptionsContext().getUnique( ReadPreferenceOption.class );
+	@SuppressWarnings("deprecation")
+	private static WriteConcern mergeWriteConcern(WriteConcern original, WriteConcern writeConcern) {
```
DavideD: What's the logic behind this? A short comment might help

core/src/test/java/org/hibernate/ogm/utils/OgmTestRunner.java
```diff
@@ -140,6 +140,8 @@ protected void runChild(FrameworkMethod method, RunNotifier notifier) {
 		}
 		finally {
 			if ( testMethodScopedSessionFactory != null ) {
+				cleanUpPendingTransactionIfRequired();
+				TestHelper.dropSchemaAndDatabase( testScopedSessionFactory );
 				testMethodScopedSessionFactory.close();
```
DavideD: Nice one.

It seems to me there are a lot of commits in this PR that are unrelated to the grouping and that could be considered as separate issues. Since I tend to do the same thing, it makes me wonder if there is something that we could do to improve the creation of separate issues (make it faster). This way PR requests would be shorter and less intimidating.

By thew way, I'm just thinking out loud, it's on a critic on your PR

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1284,42 +1231,99 @@ private void executeBatchRemove(Map<DBCollection, BatchInsertionTask> inserts, R
...
+	private void executeBatchUpdate(Map<DBCollection, BatchInsertionTask> inserts, List<Tuple> insertTuples,
+			GroupedChangesToEntityOperation groupedOperation) {
+		EntityKey entityKey = groupedOperation.getEntityKey();
+		DBCollection collection = getCollection( entityKey );
+		DBObject insertStatement = null;
+		BasicDBObject updateStatement = new BasicDBObject();
+		WriteConcern writeConcern = null;
+
+		for (Operation operation : groupedOperation.getOperations()) {
+			if ( operation instanceof InsertOrUpdateTupleOperation ) {
```
DavideD: All this instance of are a bit ugly, isn't there a pattern that we can use as a future optimization?

core/src/main/java/org/hibernate/ogm/persister/impl/OgmCollectionPersister.java
```diff
@@ -618,6 +618,7 @@ public void recreate(PersistentCollection collection, Serializable id, SessionIm
 		}
 	}
 
+	@SuppressWarnings("unchecked")
```
DavideD: Is this needed?

core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
```diff
@@ -1229,6 +1229,12 @@ private boolean isAllOrDirtyOptLocking() {
 	}
 
 	public void checkVersionAndRaiseSOSE(Serializable id, Object oldVersion, SessionImplementor session, Tuple resultset) {
+		// The tuple has been deleted
+		if ( resultset == null ) {
+			raiseStaleObjectStateException( id );
+			return;
+		}
+
```
DavideD: Is there a test for this scenario?

DavideD:
Thanks @gsmet.

This PR looks good to me and introduce several nice improvements.
After rebasing it to the latest changes I've noticed a failure though,  should be simple to solve, could you have a look at it?

gsmet:
@DavideD yes the failure is expected. I suppose it's related to the new tests counting the operations we introduced with Emmanuel?

DavideD:
@gsmet Yes, I think so

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
gsmet: Done.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
gsmet: I added a comment to explain the rationale behind this: the idea is that each separate operation might have write concern requirements and we need to build the stricter write concern for the aggregated operation.

core/src/test/java/org/hibernate/ogm/utils/OgmTestRunner.java
gsmet: Yes, it's a tough choice. Last time I extracted a lot of small changes like that, people complained about it because it was hard to see what was the overall purpose of all these changes. So, for this one, I extracted the bugs I found in separate issues and kept the rest in it. Moreover extracting this sort of things adds quite a lot of overhead to the work too. Anyway, I agree with you, it's not easy to find the right balance and I might be on the bad side on this one.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
gsmet: Yes, Gunnar also expressed concerns about the fact that the method was bigger than what we usually recommend. I must admit that it's done like this on purpose even if it's not pretty: it's far easier to debug and see how it interacts like that (and trust me, I spent a lot of time debugging it ;)).

I suppose it's a general comment: for the AbstractGroupingEntityDialect implementations, it might change if we can move some code to the base class.

core/src/main/java/org/hibernate/ogm/persister/impl/OgmCollectionPersister.java
gsmet: Nope, removed!

core/src/main/java/org/hibernate/ogm/dialect/batch/spi/OperationsQueue.java
```diff
@@ -40,14 +43,34 @@ public boolean isClosed() {
...
 
+	public OperationsQueue() {
+	}
+
 	public void add(InsertOrUpdateTupleOperation operation) {
+		add( (GroupableEntityOperation) operation );
+		insertionQueue.add( operation.getEntityKey() );
+	}
+
+	public void add(GroupableEntityOperation operation) {
```
gsmet: Yes, you're right. I did that because I'm not a big fan of the instanceof constructs but it's definitely not a good idea. That being said, considering we must have the dispatching in `add(Operation)`, I removed the other methods: I think it's clearer like that.

core/src/main/java/org/hibernate/ogm/dialect/impl/BatchOperationsDelegator.java
```diff
@@ -70,7 +74,10 @@ public void executeBatch(OperationsQueue operationsQueue) {
 		log.tracef( "Executing batch" );
```
gsmet: Yup, done!

core/src/main/java/org/hibernate/ogm/util/impl/AssociationPersister.java
```diff
@@ -139,25 +137,56 @@ public AssociationKey getAssociationKey() {
...
+				}
+			}
+			else {
+				association = gridDialect.getAssociation( key, getAssociationContext() );
+			}
+
+			if ( association == null ) {
 				association = gridDialect.createAssociation( key, getAssociationContext() );
 			}
+			if ( hostingEntity != null ) {
```
gsmet: I also want to save it in the case at line 155. But you're right, it might not be a good idea to store it in every case just to save one call.

core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
gsmet: Yes, there is a test failing somewhere if this code is not there :).

gsmet:
It has been rebased and I addressed all the comments so far.

DavideD:
Is this OK to be integrated or there is something else that needs to be done?

DavideD:
I mean, I'm going to integrate it otherwise

gsmet:
No, I think it's OK. For now, I prefer to have BatchableGridDialect and GroupingByEntityDialect separated. We can merge them later when all the dialects will be ported to grouping and we are sure it makes sense.

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/767

Thanks

DavideD:
@gsmet I'm going to close [OGM-1064](https://hibernate.atlassian.net/browse/OGM-1064) and create follow up JIRA for Neo4j and HotRod. Is that OK?

gsmet:
@DavideD yes, please. We need to implement it for Cassandra, Neo4j *, Infinispan * and EhCache.

I am not sure it will be interesting for Cassandra and EhCache as associations are always stored in separate objects so I'm not sure the optimization makes sense.

PR #769, opened 2016-09-16
OGM-1169 Deprecate DatastoreProviderType.INFINISPAN and migrate to Da…

Sanne:
…tastoreProviderType.INFINISPAN_EMBEDDED

 - https://hibernate.atlassian.net/browse/OGM-1169

Open questions:

1. rename the maven artifact id as well?
2. Should we similarly rename the class  `org.hibernate.ogm.datastore.infinispan.impl.InfinispanDatastoreProvider` ?

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/datastore/impl/DatastoreProviderType.java
+++ b/core/src/main/java/org/hibernate/ogm/datastore/impl/DatastoreProviderType.java
@@ -24,9 +24,9 @@ public enum DatastoreProviderType {
 	 * @deprecated use {@link #INFINISPAN_EMBEDDED} instead to avoid ambiguities.
 	 */
 	@Deprecated
-	INFINISPAN( "org.hibernate.ogm.datastore.infinispan.impl.InfinispanDatastoreProvider" ),
+	INFINISPAN( "org.hibernate.ogm.datastore.infinispan.impl.InfinispanEmbeddedDatastoreProvider" ),
 
-	INFINISPAN_EMBEDDED( "org.hibernate.ogm.datastore.infinispan.impl.InfinispanDatastoreProvider" ),
+	INFINISPAN_EMBEDDED( "org.hibernate.ogm.datastore.infinispan.impl.InfinispanEmbeddedDatastoreProvider" ),
 	EHCACHE( "org.hibernate.ogm.datastore.ehcache.impl.EhcacheDatastoreProvider" ),
 	MONGODB( "org.hibernate.ogm.datastore.mongodb.impl.MongoDBDatastoreProvider" ),
 	FONGO( "org.hibernate.ogm.datastore.mongodb.impl.FongoDBDatastoreProvider" ),
--- a/core/src/main/java/org/hibernate/ogm/datastore/impl/DatastoreProviderType.java
+++ b/core/src/main/java/org/hibernate/ogm/datastore/impl/DatastoreProviderType.java
@@ -6,6 +6,9 @@
  */
 package org.hibernate.ogm.datastore.impl;
 
+import org.hibernate.ogm.util.impl.Log;
+import org.hibernate.ogm.util.impl.LoggerFactory;
+
 /**
  * This enumeration describes all available datastore providers by providing some shortcuts.
  * It's used for the Datastore Provider initialization to find the provider to instantiate.
@@ -14,8 +17,16 @@ package org.hibernate.ogm.datastore.impl;
  * @author Gunnar Morling
  */
 public enum DatastoreProviderType {
+
 	MAP( "org.hibernate.ogm.datastore.map.impl.MapDatastoreProvider" ),
+
+	/**
+	 * @deprecated use {@link #INFINISPAN_EMBEDDED} instead to avoid ambiguities.
+	 */
+	@Deprecated
 	INFINISPAN( "org.hibernate.ogm.datastore.infinispan.impl.InfinispanDatastoreProvider" ),
+
+	INFINISPAN_EMBEDDED( "org.hibernate.ogm.datastore.infinispan.impl.InfinispanDatastoreProvider" ),
 	EHCACHE( "org.hibernate.ogm.datastore.ehcache.impl.EhcacheDatastoreProvider" ),
 	MONGODB( "org.hibernate.ogm.datastore.mongodb.impl.MongoDBDatastoreProvider" ),
 	FONGO( "org.hibernate.ogm.datastore.mongodb.impl.FongoDBDatastoreProvider" ),
@@ -24,6 +35,8 @@ public enum DatastoreProviderType {
 	COUCHDB_EXPERIMENTAL( "org.hibernate.ogm.datastore.couchdb.impl.CouchDBDatastoreProvider" ),
 	CASSANDRA_EXPERIMENTAL( "org.hibernate.ogm.datastore.cassandra.impl.CassandraDatastoreProvider" ),
 	REDIS_EXPERIMENTAL( "org.hibernate.ogm.datastore.redis.impl.RedisDatastoreProvider" );
+
+	private static final Log log = LoggerFactory.make();
 
 	private String datastoreProviderClassName;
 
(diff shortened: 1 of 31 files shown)


DavideD:
I would rename both, the artifact id and the the class.
In Neo4j we renamed the class as well (`EmbeddedNeo4j* and RemoteNeo4j*), but having only one artifact we kept it as `hibernate-ogm-neo4j`

Sanne:
talked about this with @emmanuelbernard , we decided to not rename the Maven id for now as we expect more renaming soon.

Sanne:
@DavideD added a second, independent commit to rename `org.hibernate.ogm.datastore.infinispan.impl.InfinispanDatastoreProvider`

core/src/main/java/org/hibernate/ogm/util/impl/Log.java
```diff
@@ -295,4 +295,9 @@
 
 	@Message(id = 87, value = "The tuple context is not available, probably because we are dealing with more than a single entity type")
 	HibernateException tupleContextNotAvailable();
+
+	@LogMessage(level = WARN)
+	@Message(id = 88, value = "Configuration is referring to deprecated DatastoreProvider name '%1$s'. Please use the new form '%2$s' instead.")
```
gunnarmorling: I'd prefer "datastore provider name" over "DatastoreProvider name".

documentation/manual/src/main/asciidoc/en-US/modules/infinispan.asciidoc
```diff
@@ -74,7 +74,7 @@ which is a good starting point for new users - you don't have to set any propert
 
 .Hibernate OGM properties for Infinispan
 `hibernate.ogm.datastore.provider`::
-Set it to `infinispan` to use Infinispan as the datastore provider.
+Set it to `infinispan_embedded` to use Infinispan as the datastore provider.
```
gunnarmorling: "use Infinispan in-memory/embedded"?

infinispan/src/main/java/org/hibernate/ogm/datastore/infinispan/InfinispanEmbedded.java
```diff
@@ -0,0 +1,33 @@
...
+import org.hibernate.ogm.datastore.infinispan.options.navigation.impl.InfinispanEntityContextImpl;
+import org.hibernate.ogm.datastore.infinispan.options.navigation.impl.InfinispanGlobalContextImpl;
+import org.hibernate.ogm.datastore.infinispan.options.navigation.impl.InfinispanPropertyContextImpl;
+import org.hibernate.ogm.datastore.spi.DatastoreConfiguration;
+import org.hibernate.ogm.options.navigation.spi.ConfigurationContext;
+
+/**
+ * Allows to configure options specific to the Infinispan data store.
+ */
+public class InfinispanEmbedded implements DatastoreConfiguration<InfinispanGlobalContext> {
```
gunnarmorling: Will we actually need different config APIs for embedded and remote?

core/src/main/java/org/hibernate/ogm/datastore/impl/DatastoreProviderType.java
```diff
@@ -25,6 +36,8 @@
 	CASSANDRA_EXPERIMENTAL( "org.hibernate.ogm.datastore.cassandra.impl.CassandraDatastoreProvider" ),
 	REDIS_EXPERIMENTAL( "org.hibernate.ogm.datastore.redis.impl.RedisDatastoreProvider" );
 
+	private static final Log log = LoggerFactory.make();
```
yrodiere: Why isn't this used below? If there's a reason, maybe it deserves a comment... ?

infinispan/src/main/java/org/hibernate/ogm/datastore/infinispan/InfinispanDialect.java
```diff
@@ -55,9 +55,9 @@
  */
 public class InfinispanDialect<EK,AK,ISK> extends BaseGridDialect {
```
yrodiere: Since this class relies on an embedded provider, maybe we'd better rename this to `InifinispanEmbeddedDialect`? If it's a matter of not breaking compatibility, we can always add another deprecated class with the legacy name (`InfinispanDialect`), extending this one.
Of course, if this class is ultimately expected to be used for both embedded and remote datastore providers, and we just don't have an interface for `InfinispanDatastoreProvider` yet, then it's okay.

documentation/manual/src/main/asciidoc/en-US/modules/configuration.asciidoc
```diff
@@ -36,7 +36,7 @@ The provider name is [classname]`org.hibernate.ogm.jpa.HibernateOgmPersistence`.
             <property name="hibernate.transaction.jta.platform"
                       value="JBossTS" />
             <property name="hibernate.ogm.datastore.provider"
-                      value="infinispan" />
+                      value="infinispan_embedded" />
```
yrodiere: Running `grep -R '"infinispan"'` returned several remaining occurrences of the deprecated string in the tests. Is that on purpose?

core/src/main/java/org/hibernate/ogm/util/impl/Log.java
Sanne: ok, changing.

documentation/manual/src/main/asciidoc/en-US/modules/infinispan.asciidoc
Sanne: I'll change this instance, but I wouldn't go further at this stage. With the introduction of the remote dialect we'll likely need to refactor the whole "Infinispan" chapter to have a common introduction and then a split in subsections.

infinispan/src/main/java/org/hibernate/ogm/datastore/infinispan/InfinispanEmbedded.java
Sanne: Good question. I assumed yes, especially as we want to keep the dependencies isolated from each other so that we can upgrade them independently.

Some options will necessarily be different as they are quite different; but indeed they will share some concepts. I'd stick to share names were it makes sense, but different API and package.

core/src/main/java/org/hibernate/ogm/datastore/impl/DatastoreProviderType.java
Sanne: Good catch. I intended to use it, fixing..

documentation/manual/src/main/asciidoc/en-US/modules/configuration.asciidoc
Sanne: One was left on purpose, to test the previous name. But you're right, I missed some more.. fixing. Thanks!

infinispan/src/main/java/org/hibernate/ogm/datastore/infinispan/InfinispanDialect.java
gsmet: @Sanne looks like you missed this comment?

gsmet:
I'm a little late to the game but I just wanted to point out that, for Neo4j, we have:
* `EmbeddedNeo4jDatastoreProvider`
* `RemoteNeo4jDatastoreProvider` (soon to be `BoltNeo4jDatastoreProvider` and `HttpNeo4jDatastoreProvider`)

and for Infinispan, it looks like you chose `InfinispanEmbeddedDatastoreProvider` with Embedded after Infinispan. Shouldn't we try to be consistent with the naming?

infinispan/src/main/java/org/hibernate/ogm/datastore/infinispan/InfinispanDialect.java
Sanne: > looks like you missed this comment?

I run out of time :)

Will have to resume this task soon.

DavideD:
Rebased and Merged https://github.com/hibernate/hibernate-ogm/pull/769

I've created [OGM-1186](https://hibernate.atlassian.net/browse/OGM-1186) to remember to review naming conventions

PR #770, opened 2016-09-19
OGM-1089 Integration with Neo4j using Bolt

DavideD:
https://hibernate.atlassian.net/browse/OGM-1089

This PR will add a new dialect that uses the Bolt protocol instead of the HTTP API.

I opted for the following prefixes:
- BOLT: for the bolt protocol
- HTTP: for the HTTP api
- REMOTE: for the classes that are common to the Bolt and Http ones

The package containing the classes are under 
`org.hibernate.ogm.datastore.neo4j.remote.[bolt|http|common]` 

The mapping is identical to the existing ones, infact it reuses the existing queries. 
The code could be rafctor more but I think this is good as a first stage.

At the moment the tests under the integrationtest module will only work if the default port is used for Bolt.

Is it possible to run the test via:
`mvn install -pl neo4j -P neo4j-bolt`

The connection via bolt is manged using the `neo4j-java-driver`


core/src/test/java/org/hibernate/ogm/backendtck/batchfetching/MultiGetSingleColumnIdTest.java
```diff
@@ -185,6 +186,19 @@ private void rollback(Transaction tx) {
 		return new Class<?>[]{ BoardGame.class };
 	}
 
+	// The conversion to the right type is done later and some datastore don't keep track of the exact type.
+	// For example Neo4j in remote mode, will return a Long instead of an Integer.
+	private Integer id(Tuple tuple) {
+		Object object = tuple.get(  "id" );
```
gsmet: Checkstyle (2 spaces after the parenthesis)

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/BaseNeo4jSchemaDefiner.java
```diff
@@ -24,47 +28,80 @@
...
  */
-public abstract class BaseNeo4jSchemaDefiner<T> extends BaseSchemaDefiner {
+public abstract class BaseNeo4jSchemaDefiner extends BaseSchemaDefiner {
+
+	public static class UniqueConstraintDetails {
+		private final Label label;
+		private final String property;
+
+		public UniqueConstraintDetails(Label label, String property) {
+			super();
```
gsmet: Maybe remove the useless `super()`?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/BaseNeo4jSchemaDefiner.java
```diff
@@ -24,47 +28,80 @@
...
 						Column column = columnIterator.next();
 						if ( column.isUnique() ) {
-							createUniqueConstraintIfMissing( neo4jDb, label, column.getName() );
+							constraints.add( new UniqueConstraintDetails( label, column.getName() ) );
 						}
 					}
 					Iterator<UniqueKey> uniqueKeyIterator = table.getUniqueKeyIterator();
 					while ( uniqueKeyIterator.hasNext() ) {
-						createConstraint( neo4jDb, table, label, uniqueKeyIterator.next() );
+						createConstraint( constraints, table, label, uniqueKeyIterator.next() );
```
gsmet: Maybe rename this `addConstraint` as it does not create the constraints anymore.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/common/dialect/impl/RemoteNeo4jAssociationSnapshot.java
```diff
@@ -29,6 +31,11 @@ public RemoteNeo4jAssociationSnapshot(Map<RowKey, Tuple> tuples) {
 	@Override
 	public Tuple get(RowKey rowKey) {
 		Tuple tuple = tuples.get( rowKey );
+		for ( Entry<RowKey, Tuple> entry : tuples.entrySet() ) {
+			if ( RemoteNeo4jHelper.matches( entry.getKey(), rowKey ) ) {
+				return entry.getValue();
+			}
+		}
```
gsmet: The logic is weird here. Maybe you should directly return the tuple from the first line if not null? That would work in the simple case and if you don't find it you execute the fallback logic.

At least, that's how I understand what you're trying to do but I may be wrong.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/common/impl/RemoteNeo4jDatastoreProvider.java
```diff
@@ -0,0 +1,97 @@
...
+		return false;
+	}
+
+	protected RemoteNeo4jDatabaseIdentifier getDatabaseIdentifier() {
+		if ( !configuration.getHosts().isSingleHost() ) {
+			logger.doesNotSupportMultipleHosts( configuration.getHosts().toString() );
+		}
+		Hosts.HostAndPort hostAndPort = configuration.getHosts().getFirst();
+		try {
+			return new RemoteNeo4jDatabaseIdentifier( getProtocol(), getConfiguration() );
```
gsmet: Very minor but we should be consistent and use directly `protocol` and `configuration` (I had to check if there was a good reason to use the getters for this specific call).

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/common/request/impl/RemoteStatements.java
```diff
@@ -0,0 +1,35 @@
...
+		statements.add( statement );
+	}
+
+	@Override
+	public Iterator<RemoteStatement> iterator() {
+		return Collections.unmodifiableCollection( statements ).iterator();
+	}
+
+	public void addStatement(RemoteStatement statement) {
+		statements.add( statement );
```
gsmet: Why is there a `add` and a `addStatement`? Looks like we should remove one. Btw, maybe we could return `this` so that we could chain the calls?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/http/dialect/impl/HttpNeo4jTupleSnapshot.java
```diff
@@ -27,45 +26,34 @@
...
+	private final Map<String, String> rolesByColumn;
 
-	public RemoteNeo4jTupleSnapshot(RemoteNeo4jClient neo4jClient, Long txId, RemoteNeo4jEntityQueries queries, NodeWithEmbeddedNodes node, EntityKeyMetadata entityKeyMetadata) {
-		this( neo4jClient, txId, queries, node, Collections.<String, AssociatedEntityKeyMetadata>emptyMap(), Collections.<String, String>emptyMap(),
-				entityKeyMetadata );
+	public HttpNeo4jTupleSnapshot(
+			NodeWithEmbeddedNodes node,
+			EntityKeyMetadata entityKeyMetadata) {
+		this( node, entityKeyMetadata, Collections.<String, Node>emptyMap(), null );
 	}
```
gsmet: Looks like this constructor is never called. Maybe we should remove it?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/http/dialect/impl/HttpNeo4jTupleSnapshot.java
```diff
@@ -85,27 +73,12 @@ private Object readPropertyOnOtherNode(String column) {
...
 		if ( associatedEntity == null ) {
-			// Not cached, let's look for it
-			associatedEntity = queries.findAssociatedEntity( neo4jClient, txId, keyValues(), associationrole );
-			if ( associatedEntity == null ) {
-				return null;
-			}
-			else {
-				toOneEntities.put( associationrole, associatedEntity );
-			}
+			return null;
```
gsmet: So, if I understand this change correctly, we now always get the associated entities?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/http/impl/HttpNeo4jClient.java
```diff
@@ -25,7 +30,11 @@
  *
  * @author Davide D'Alto
  */
-public class RemoteNeo4jClient implements AutoCloseable {
+public class HttpNeo4jClient {
```
gsmet: Any particular reason why you removed the AutoCloseable interface? Just curious.

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/utils/EmbeddedNeo4jTestHelperDelgate.java
```diff
@@ -0,0 +1,207 @@
...
+import org.neo4j.graphdb.PropertyContainer;
+import org.neo4j.graphdb.QueryExecutionException;
+import org.neo4j.graphdb.ResourceIterator;
+import org.neo4j.graphdb.Result;
+import org.neo4j.graphdb.Transaction;
+
+/**
+ * @author Davide D'Alto
+ */
+public class EmbeddedNeo4jTestHelperDelgate implements Neo4jTestHelperDelegate {
```
gsmet: s/Delgate/Delegate/

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/utils/Neo4jTestHelper.java
```diff
@@ -99,87 +62,26 @@ private static boolean isNotNull(String neo4jHostName) {
 
 	@Override
 	public long getNumberOfEntities(Session session) {
-		DatastoreProvider provider = getProvider( session.getSessionFactory() );
-		return getNumberOfEntities( session, provider );
+		DatastoreProvider provider = getDatastoreProvider( session.getSessionFactory() );
+		return delegate() .getNumberOfEntities( session, provider );
```
gsmet: Space after `delegate() .`

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/utils/Neo4jTestHelperDelegate.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.hibernate.Session;
+import org.hibernate.ogm.datastore.neo4j.dialect.impl.NodeLabel;
+import org.hibernate.ogm.datastore.neo4j.test.dsl.NodeForGraphAssertions;
+import org.hibernate.ogm.datastore.neo4j.test.dsl.RelationshipsChainForGraphAssertions;
+import org.hibernate.ogm.datastore.spi.DatastoreProvider;
+import org.hibernate.ogm.dialect.spi.GridDialect;
+
+/**
+ * Groups together the operations needed by the {@link Neo4jTestHelper} that are different for each Neo4j back-end.
```
gsmet: s/back-end/backend/

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/bolt/dialect/impl/BoltNeo4jAssociatedNodesHelper.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Map.Entry;
+
+import org.hibernate.ogm.dialect.spi.TupleContext;
+import org.hibernate.ogm.model.key.spi.AssociatedEntityKeyMetadata;
+import org.hibernate.ogm.model.key.spi.EntityKeyMetadata;
+import org.neo4j.driver.v1.Transaction;
+import org.neo4j.driver.v1.types.Node;
+
+/**
+ * Function required to get the associated node when building the tuple snapshot.
```
gsmet: s/node/nodes/

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/bolt/impl/BoltNeo4jSchemaDefiner.java
```diff
@@ -0,0 +1,118 @@
...
+		for ( UniqueConstraintDetails constraint : constraints ) {
+			log.tracef( "Creating unique constraint for nodes labeled as %1$s on property %2$s", constraint.getLabel(), constraint.getProperty() );
+			StringBuilder queryBuilder = new StringBuilder( "CREATE CONSTRAINT ON (n:" );
+			escapeIdentifier( queryBuilder, constraint.getLabel().name() );
+			queryBuilder.append( ") ASSERT n." );
+			escapeIdentifier( queryBuilder, constraint.getProperty() );
+			queryBuilder.append( " IS UNIQUE" );
+			statements.add( new Statement( queryBuilder.toString() ) );
+		}
+
```
gsmet: Bolt and Neo4j could probably share most of this implementation but I suppose you thought about it and preferred this approach?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/bolt/transaction/impl/BoltNeo4jJtaTransactionCoordinator.java
```diff
@@ -0,0 +1,189 @@
...
+	private void closeSession() {
+		try {
+			session.close();
+		}
+		finally {
+			session = null;
+		}
+	}
+
+	private class Neo4jSynchronization implements Synchronization {
```
gsmet: Looks like we could share most of the code here by passing `this` to `Neo4jSynchronization` and `RemoteTransactionDriver`?

integrationtest/src/test/java/org/hibernate/ogm/test/integration/neo4j/remote/bolt/BoltNeo4jJtaModuleMemberRegistrationIT.java
```diff
@@ -0,0 +1,88 @@
...
+	private static PersistenceDescriptor persistenceXml() throws Exception {
+		Properties<PersistenceUnit<PersistenceDescriptor>> propertiesContext = Descriptors.create( PersistenceDescriptor.class )
+				.version( "2.0" )
+				.createPersistenceUnit()
+				.name( "primary" )
+				.transactionType( PersistenceUnitTransactionType._JTA )
+				.provider( HibernateOgmPersistence.class.getName() )
+				.getOrCreateProperties();
+		PersistenceDescriptor persistenceDescriptor = propertiesContext
+				.createProperty().name( Neo4jProperties.DATASTORE_PROVIDER ).value( Neo4j. BOLT_DATASTORE_PROVIDER_NAME ).up()
```
gsmet: There is a space after `Neo4j. `, we should remove it.

gsmet:
Nice patch.

I added a couple of comments here and there but overall I really like the approach and how you managed to keep the quantity of specific code quite low between Bolt and HTTP. Thanks for taking the time to separate the commits, it helped a lot!

Did you reach the Neo4j people about the Integer/Long issue? IIRC, they advertised Bolt as better typed and it looks like this problem should be solved on their part?

gsmet:
Btw, more of a reminder for when I'll merge it (but if you want to fix it right away, be my guest ;)), there are a couple of typos in the first 2 commit comments:
 * 3dbb4c80384b7de1f528b832fc6d75ddd3cbce29 -> "the the" -> "the"
 * 0f3cca0b5605ed2af22047ae97fe83a6f5e03141 -> "ne4j" -> "neo4j"

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/common/request/impl/RemoteStatements.java
DavideD: Are you talking about `statements.add`? That's a java.util.List. RemoteStatements hide the internal representation. I'm not sure I understand the comments, they are two different things.

Ok for the chain calls

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/http/dialect/impl/HttpNeo4jTupleSnapshot.java
DavideD: Yes, maybe this should have been a separate commits.

For the http case, before I was loading this information from the snapshot, in bolt I was doing that outside. I don't think the snapshot should change so when refactoring I opted for the bolt approach

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/http/impl/HttpNeo4jClient.java
DavideD: No particular reason, I guess I didn't notice I removed it because we are not using it anywhere in our code.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/common/request/impl/RemoteStatements.java
gsmet: Nope. You have `add(RemoteStatement statement)` and `addStatement(RemoteStatement statement)` in this very class. I think one should be removed.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/common/dialect/impl/RemoteNeo4jAssociationSnapshot.java
DavideD: As you said, this is more a mistake. I think it's something left from before. I'll keep it as a fall-back

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/common/request/impl/RemoteStatements.java
DavideD: Ok, I'll remove the add method

DavideD:
I've applied the remarks and rebased.

@gsmet I applied all the changes you suggested. I also added a couple of commits to extract some duplicated code

gsmet:
Jenkins, retest this please

gsmet:
Jenkins, retest this please

gsmet:
Merged! Thanks!

PR #772, opened 2016-10-04
Proposal for 5.0.2

gsmet:
Here is my proposal for 5.0.2. The main change is that it adds support for MongoDB 3.2 (and still supports 3.0 which is the version used in the tests). Currently 5.0.x does not support MongoDB 3.2 and doesn't fail gracefully: some features simply does not work while others work correctly (see http://stackoverflow.com/questions/39793733/unable-to-remove-elements-from-mongodb-using-hibernate-ogm/39795532). As the work was already done to support 3.2 (released in Dec 2015) and it doesn't break compatibility with 3.0, I think it might be interesting to do a quick release with this support (and a few other fixes).

(Note: the changes to the release scripts are useless but they were part of a commit so I let them like that, we should be able to use the unified release scripts for this release)


Sanne:
So it's a selection of backports right?

gsmet:
Jenkins, retest this please

gsmet:
@Sanne yes, sometimes stripped to not introduce too many changes.

Sanne:
Good selection, I merged it.

I'm a bit troubled though that all these issues are already marked as "resolved" for the target release; for issues which are already in the "Resolved" state, we typically only mark them for a branch if it was already included: otherwise JIRA's state would be inconsistent with the branch state.

gsmet:
@Sanne yes, I did that yesterday to keep track of the candidates I wanted to backport as the `5.0.2.Final` release was a brand new release without any prior issues.

PR #777, opened 2016-10-13
OGM-1187 Run integration tests on Neo4j using Bolt

DavideD:
https://hibernate.atlassian.net/browse/OGM-1187

At the moment they are not executed and because of this I didn't notice that the port they were using was wrong


gsmet:
@DavideD so this one is merged. Are you aware that these tests are not running on CI though?

gsmet:
Merged! Thanks!

PR #778, opened 2016-10-17
OGM-1157 Check import of `org.hibernate.annotations.common`

DavideD:
https://hibernate.atlassian.net/browse/OGM-1157

  Some classes are using `org.hibernate.annotations.common.AssrtionFailure` instead of
  `org.hibernate.AssertionFailure`, this causes exception in modular environment.

  This change is more restrictive but it seems that we are not using any class
  under org.hibernate.annotations.common.*, so, I prefer to exclude all those classes instead
  of adding a custom checkstyle rule.


gsmet:
Merged! Thanks!

PR #784, opened 2016-10-25
OGM-1194 Add infinispan remote to the distribution

DavideD:
https://hibernate.atlassian.net/browse/OGM-1194


Sanne:
merged. Thanks!

PR #785, opened 2016-10-27
OGM-1173 Configuration for HotRod

DavideD:
https://hibernate.atlassian.net/browse/OGM-1173

### DIFF
--- a/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java
+++ b/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java
@@ -12,6 +12,13 @@ import org.hibernate.ogm.datastore.keyvalue.cfg.KeyValueStoreProperties;
 /**
  * Properties for configuring the Infinispan Remote datastore via {@code persistence.xml} or
  * {@link StandardServiceRegistryBuilder}.
+ * <p>
+ * It is possible to configure the Hot Rod (Infinispan remote) client prefixing them with 'hibernate.ogm.infinispan_remote.client.'.
+ * For example, the property 'infinispan.client.hotrod.tcp_no_delay' can be set in the hibernate.properties file
+ * using 'hibernate.ogm.infinispan_remote.client.infinispan.client.hotrod.tcp_no_delay'.
+ * <p>
+ * Properties expressed with the prefix 'hibernate.ogm.infinispan_remote.client.' will override the same properties
+ * expressed in the configuration file.
  */
 public final class InfinispanRemoteProperties implements KeyValueStoreProperties {
 
@@ -20,6 +27,11 @@ public final class InfinispanRemoteProperties implements KeyValueStoreProperties
 	 * for the Hot Rod (Infinispan remote) client.
 	 */
 	public static final String CONFIGURATION_RESOURCE_NAME = "hibernate.ogm.infinispan_remote.configuration_resource_name";
+
+	/**
+	 * Prefix for the Hot Rod (Infinispan remote) client properties.
+	 */
+	public static final String HOT_ROD_CLIENT_PREFIX = "hibernate.ogm.infinispan_remote.client.";
 
 	/**
 	 * You can inject an instance of {@link org.hibernate.ogm.datastore.infinispanremote.schema.spi.SchemaCapture} into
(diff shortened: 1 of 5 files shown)


Sanne:
Looking good, but different approach than what we decided in chat yesterday? 

I expected that the hotrod prefix wouldn't be necessary, so making properties a bit shorter, but I might have misunderstood. 

@gunnarmorling ?

DavideD:
My understanding was the we wanted both

DavideD:
With this patch you can still use the resource file

DavideD:
@Sanne By the way, would it be possible to provide a default configuraiton for Hot rod or do you think the user should always set it? I'm asking because now we throw an exception if the user does not set the resource file

Sanne:
resuming this review, sorry for the delay.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java
```diff
@@ -12,6 +12,13 @@
 /**
  * Properties for configuring the Infinispan Remote datastore via {@code persistence.xml} or
  * {@link StandardServiceRegistryBuilder}.
+ * <p>
+ * It is possible to configure the Hot Rod (Infinispan remote) client prefixing them with 'hibernate.ogm.infinispan_remote.client.'.
+ * For example, the property 'infinispan.client.hotrod.tcp_no_delay' can be set in the hibernate.properties file
+ * using 'hibernate.ogm.infinispan_remote.client.infinispan.client.hotrod.tcp_no_delay'.
```
Sanne: I was expecting a short-hand form for the Infinispan properties.

So rather than "hibernate.ogm.infinispan_remote.client.infinispan.client.hotrod.tcp_no_delay" I'd prefer one of:

- _hibernate.ogm.infinispan_remote.client.tcp_no_delay_
- _hibernate.ogm.infinispan_remote.hotrod.tcp_no_delay_

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/configuration/impl/InfinispanRemoteConfiguration.java
```diff
@@ -87,4 +106,36 @@ public void initConfiguration(Map<?, ?> configurationMap, ServiceRegistryImpleme
...
+			}
+			catch (IOException e) {
+				throw log.failedLoadingHotRodConfigurationProperties( e );
+			}
+		}
+		// Programmatic properties override the resource file
+		for ( Entry<?, ?> property : configurationMap.entrySet() ) {
+			String key = (String) property.getKey();
+			if ( key.startsWith( HOT_ROD_CLIENT_PREFIX ) ) {
+				String value = propertyReader.property( key, String.class ).withDefault( null ).getValue();
```
Sanne: .withDefault( null )

This is unnecessary right? A not found key would mean something is wrong with our configuration parsing code; maybe let it throw an exception instead?

Sanne:
I think we need to shorten the prefix, to not include the redundant components of Infinispan's own prefix.

This implies we need special care for the properties which relate to the Hot Rod connection pool, as these don't use the prefix.
That's why I gave you a pointer to the list of property names, as I suspect we'll need to list them all in our code:
- https://hibernate.atlassian.net/browse/OGM-1173?focusedCommentId=85324&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-85324

Sanne:
> By the way, would it be possible to provide a default configuration for Hot rod or do you think the user should always set it? I'm asking because now we throw an exception if the user does not set the resource file.

We certainly should not throw an exception if the configuration resource is missing, as long as the alternative required properties are set.

If neither the resource file nor the Hibernate properties define a `infinispan.client.hotrod.server_list` then I think we should throw an exception, and mention exactly which property is missing.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java
DavideD: I agree with you and there are two reasons why I opted for the prefix:
1. I don't have to list all the properties for hot rod
2. the user doesn't have to learn different too many different rules to set the same property.

I will try a different approach to see what happens

DavideD: > I will try a different approach to see what happens

I mean, the approach you suggested

DavideD:
I'll close this one and push the changes on https://github.com/hibernate/hibernate-ogm/pull/787

PR #786, opened 2016-10-28
OGM-1198 Mapping error with multiple associations on the same entity

DavideD:
https://hibernate.atlassian.net/browse/OGM-1198

Check the link on stack Overflow for an example: http://stackoverflow.com/questions/40288481/persist-a-tree-with-hibernate-ogm-and-mongodb-the-collections-that-represent-th

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/persister/impl/BiDirectionalAssociationHelper.java
+++ b/core/src/main/java/org/hibernate/ogm/persister/impl/BiDirectionalAssociationHelper.java
@@ -11,6 +11,7 @@ import java.util.Arrays;
 import org.hibernate.HibernateException;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.ogm.model.key.spi.AssociationKeyMetadata;
+import org.hibernate.ogm.util.impl.StringHelper;
 import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.persister.entity.Joinable;
 import org.hibernate.persister.entity.Loadable;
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/hibernate/ogm/persister/impl/BiDirectionalAssociationHelper.java
```diff
@@ -76,15 +76,30 @@ public static AssociationKeyMetadata getInverseAssociationKeyMetadata(OgmEntityP
...
+						return inverseCollectionPersister.getAssociationKeyMetadata();
+					}
 				}
 			}
 		}
 
 		return null;
 	}
 
+	private static boolean isEmpty(String mappedByProperty) {
```
gsmet: Maybe use `StringHelper`.

core/src/main/java/org/hibernate/ogm/persister/impl/BiDirectionalAssociationHelper.java
```diff
@@ -76,15 +76,30 @@ public static AssociationKeyMetadata getInverseAssociationKeyMetadata(OgmEntityP
 			// candidate is a *-to-many association
 			if ( type.isCollectionType() ) {
 				OgmCollectionPersister inverseCollectionPersister = getPersister( factory, (CollectionType) type );
-				if ( isCollectionMatching( mainSideJoinable, inverseCollectionPersister ) ) {
-					return inverseCollectionPersister.getAssociationKeyMetadata();
+				String mappedByProperty = inverseCollectionPersister.getMappedByProperty();
+				if ( isEmpty( mappedByProperty ) ) {
```
gsmet: You prefer having a if / else if instead of a or in the first if?

gsmet:
@DavideD I added a few comments, the important one being about the case when we don't have a `mappedBy`.

I think this one is bad enough we should consider backporting it to 5.0.x and make a release. WDYT?

DavideD:
I've updated the PR following your remarks. 

> Thinking a bit more about it, if there is no mappedBy, can't we consider it's not the inverse relationship? I'm surprised we returned something in this case.

Good catch, I've tried to return null in that case and everything seems to work after I've updated a test.
You can check in the last commit
Thanks

gsmet:
@DavideD well, it was a good catch but, in fact, I didn't realize the method was working when we passed an inverse side relationship to it.

The javadoc is very clear it's not supposed to and it's only used to manage the navigational information of the inverse side so I think we're safe and it's better to be strict about this.

I was more thinking about the user declaring a relation that could match but isn't declared as an inverse relationship with a `mappedBy` and so shouldn't be considered the inverse relationship of the main side.

gsmet:
Rebased and merged!

PR #787, opened 2016-10-28
OGM-1162 Validate configuration for HotRod

DavideD:
https://hibernate.atlassian.net/browse/OGM-1162

I built this on top of https://github.com/hibernate/hibernate-ogm/pull/785
I think it is OK to provide a way to setup the client using parameters in addition to the resource file.
This allows a user to override default values or write tests more easily.

@Sanne let me know which other options you think we need to require. I didn't add a warning for those options that are harmless (it's easy to add if you feel strong about it).

### DIFF
--- a/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java
+++ b/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java
@@ -7,11 +7,35 @@
 package org.hibernate.ogm.datastore.infinispanremote;
 
 import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
+import org.hibernate.ogm.datastore.infinispanremote.configuration.impl.InfinispanRemoteConfiguration;
 import org.hibernate.ogm.datastore.keyvalue.cfg.KeyValueStoreProperties;
 
 /**
  * Properties for configuring the Infinispan Remote datastore via {@code persistence.xml} or
  * {@link StandardServiceRegistryBuilder}.
+ * <p>
+ * It also contains the properites to configure the Hot Rod client.
+ * There are two ways to configure the client:
+ * <ul>
+ *   <li>Using an external Hot Rod properties file; see {@link InfinispanRemoteProperties#CONFIGURATION_RESOURCE_NAME}
+ *   <li>Defining the properties of the client using {@value InfinispanRemoteProperties#HOT_ROD_CLIENT_PREFIX}
+ * </ul>
+ * <p>
+ * When a property of the Hot Rod client is set in the hibernate configuration, it should be prefixed
+ * with {@value InfinispanRemoteProperties#HOT_ROD_CLIENT_PREFIX}.
+ * For example, the property {@code infinispan.client.hotrod.server_list}
+ * becomes {@code hibernate.ogm.infinispan_remote.client.server_list}.
+ * <p>
+ * Currently, some properties in the Hot Rod client don't have a prefix, in this case it must be added
+ * in the hibernate configuration.
+ * For example, the property {@code maxActive} becomes {@code hibernate.ogm.infinispan_remote.client.maxActive}
+ * <p>
+ * Properties with the Hibernate OGM prefix ({@value InfinispanRemoteProperties#HOT_ROD_CLIENT_PREFIX}) will
+ * override corresponding properties defined in the external Hot Rod configuration file.
+ *
+ * @see InfinispanRemoteConfiguration
+ *
+ * @author Davide D'Alto
  */
 public final class InfinispanRemoteProperties implements KeyValueStoreProperties {
 
@@ -20,6 +44,11 @@ public final class InfinispanRemoteProperties implements KeyValueStoreProperties
 	 * for the Hot Rod (Infinispan remote) client.
 	 */
 	public static final String CONFIGURATION_RESOURCE_NAME = "hibernate.ogm.infinispan_remote.configuration_resource_name";
+
+	/**
+	 * Prefix for the Hot Rod (Infinispan remote) client properties.
+	 */
+	public static final String HOT_ROD_CLIENT_PREFIX = "hibernate.ogm.infinispan_remote.client.";
 
 	/**
 	 * You can inject an instance of {@link org.hibernate.ogm.datastore.infinispanremote.schema.spi.SchemaCapture} into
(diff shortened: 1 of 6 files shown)


DavideD:
I've also added a commit that set the expected values if the user doesn't do it.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/configuration/impl/InfinispanRemoteConfiguration.java
```diff
@@ -31,6 +34,14 @@
 
 	private static final Log log = LoggerFactory.getLogger();
 
+	/*
+	 * The expected configuration value for some properties, an exception is thrown if the value is changed.
+	 */
+	private static final String[][] expectedValuesForHotRod = {
+			{ FORCE_RETURN_VALUES, "true" },
+			{ MARSHALLER, OgmProtoStreamMarshaller.class.getName() }
```
Sanne: ah, sorry I see that's exactly what you do. I got a bit confused as you don't have to set the Marshaller property, but this shouldn't be a problem.

Sanne:
Cool, looks good but it's based on a PR for which I'd rather make some changes (#785).
Feel free to apply this and make changes on top of it to avoid a complex rebase?

Might want to fix on the minor remark, that the marshallers are not set using properties but injected directly as an instance (we need to, as our custom marshaller is initialized with a reference to our generated schema).

DavideD:
> Might want to fix on the minor remark, that the marshallers are not set using properties but injected directly as an instance (we need to, as our custom marshaller is initialized with a reference to our generated schema).

Does this comment come after the following one?

>ah, sorry I see that's exactly what you do. I got a bit confused as you don't have to set the Marshaller property, but this shouldn't be a problem

To summarize (for others as well):
we don't allow the user to change the marshaller but he can still set it in the properties.
The patch makes sure to check that if the value is set, it  is set to the expected value, and if not set it will use the expected values as default.

I guess I could have just throw an error when the property is set (it does not matter to which value), but it seems odd to throw an exception when the value is set to something that it is correct.

Sanne:
> I guess I could have just throw an error when the property is set (it does not matter to which value), but it seems odd to throw an exception when the value is set to something that it is correct.

Yes you can ignore this. My only concern is that just using the right class name is not good enough, it has to be the one instance which OGM initialized, so the user's choice would not be correct even if the name of the class matches. Not a problem though, as we override it..

DavideD:
@Sanne  I've rebased and pushed the changes.

Now the prefix is `hibernate.ogm.infinispan_remote.client.`:
`infinispan.client.hotrod.server_list` becomes `hibernate.ogm.infinispan_remote.client.server_list`;
`maxActive` becomes `hibernate.ogm.infinispan_remote.client.maxActive`..

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteProperties.java
```diff
@@ -12,6 +12,13 @@
 /**
  * Properties for configuring the Infinispan Remote datastore via {@code persistence.xml} or
  * {@link StandardServiceRegistryBuilder}.
+ * <p>
+ * It is possible to configure the Hot Rod (Infinispan remote) client prefixing them with 'hibernate.ogm.infinispan_remote.client.'.
+ * For example, the property 'infinispan.client.hotrod.tcp_no_delay' can be set in the hibernate.properties file
+ * using 'hibernate.ogm.infinispan_remote.client.infinispan.client.hotrod.tcp_no_delay'.
```
Sanne: should be _hibernate.ogm.infinispan_remote.client.tcp_no_delay_ (shorter) right?

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/configuration/impl/InfinispanRemoteConfiguration.java
```diff
@@ -87,4 +167,87 @@ public void initConfiguration(Map<?, ?> configurationMap, ServiceRegistryImpleme
...
+	 * @param hotRodConfiguration the Hot Rod configuration to update
+	 */
+	private void setAdditionalProperties(Map<?, ?> configurationMap, ConfigurationPropertyReader propertyReader, Properties hotRodConfiguration) {
+		// Programmatic properties override the resource file
+		for ( Entry<?, ?> property : configurationMap.entrySet() ) {
+			String key = (String) property.getKey();
+			if ( key.startsWith( HOT_ROD_CLIENT_PREFIX ) ) {
+				String hotRodProperty = key.substring( HOT_ROD_CLIENT_PREFIX.length() );
+				String value = propertyReader.property( key, String.class ).getValue();
+				if ( !ArrayHelper.contains( noPrefixProperties, hotRodProperty ) ) {
```
Sanne: I'd have used a hashmap here: much better "contains" efficiency.
Not important to change though, as it's a very small array (and only happens at boot time).

infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/test/initialize/InfinispanRemoteConfigurationTest.java
```diff
@@ -0,0 +1,153 @@
...
+		// The one I set manually
+		assertThat( clientProperties.getProperty( DEFAULT_EXECUTOR_FACTORY_QUEUE_SIZE ) ).isEqualTo( "99" );
+		assertThat( clientProperties.getProperty( EVICTION_MILLIS_PROPERTY ) ).isEqualTo( "129999" );
+	}
+
+	private Properties extractClientConfiguration(Map<String, Object> settings) {
+		try ( SessionFactory sessionFactory = TestHelper.getDefaultTestSessionFactory( settings ) ) {
+			ServiceRegistryImplementor serviceRegistry = ( (SessionFactoryImplementor) sessionFactory ).getServiceRegistry();
+			InfinispanRemoteConfiguration conf = new InfinispanRemoteConfiguration();
+			conf.initConfiguration( settings, serviceRegistry );
```
Sanne: Nice trick!

Sanne:
Feel free to push it once the javadoc is updated. Great, thanks!

DavideD:
OK, thanks

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/787

Thanks

PR #791, opened 2016-11-04
OGM-1006 Neo4j mass indexer embedded

DavideD:
https://hibernate.atlassian.net/browse/OGM-1006

### DIFF
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/BaseNeo4jEntityQueries.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/BaseNeo4jEntityQueries.java
@@ -467,6 +467,7 @@ public abstract class BaseNeo4jEntityQueries extends BaseNeo4jQueries {
 		queryBuilder.append( ENTITY );
 		queryBuilder.append( ":" );
 		appendLabel( entityKeyMetadata, queryBuilder );
+		queryBuilder.append( ")" );
 		appendOptionalMatchOwnerEmbeddedNodes( queryBuilder );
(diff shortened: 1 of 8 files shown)


DavideD:
@Sanne or @gsmet  can you have a look at this?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/BaseNeo4jEntityQueries.java
```diff
@@ -467,6 +467,7 @@ private static String initFindEntitiesQuery(EntityKeyMetadata entityKeyMetadata,
 		queryBuilder.append( ENTITY );
 		queryBuilder.append( ":" );
 		appendLabel( entityKeyMetadata, queryBuilder );
+		queryBuilder.append( ")" );
```
gsmet: I'm surprised we only find this now. Don't we have a unit test exercising this?

gsmet:
Typos in the commit messages:
s/retrieveal/retrieval/
s/depedendencies/dependencies/

Sanne:
merged

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/BaseNeo4jEntityQueries.java
DavideD: Now that you mention it I think we missed the tests for the foreach method. It's been introduced for the mass indexer so it is only tested if we can run with the dialect.

I will create a JIRA for that

Sanne: Indeed!

PR #794, opened 2016-11-08
OGM-1159 Clarify the Cache names and configurations need to be config…

Sanne:
…ured manually


DavideD:
Merged  https://github.com/hibernate/hibernate-ogm/pull/794

Thanks

PR #796, opened 2016-11-08
OGM-1205 Occasional failures of SequenceIdGeneratorTest for the Hot R…

Sanne:
…od dialect

https://hibernate.atlassian.net/browse/OGM-1205

### DIFF
--- a/infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/test/sequences/SequencedActor.java
+++ b/infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/test/sequences/SequencedActor.java
@@ -1,0 +1,55 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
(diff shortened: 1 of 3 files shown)


infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/test/sequences/SequencedActor.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+@Entity
+public class SequencedActor {
+	static final transient int INITIAL_VALUE = 1;
+	private Long id;
+	private String name;
+	private String totalMovies;
+
+	@Id
+	@GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "sequencerActorGenerator")
```
gsmet: s/sequencer/sequenced/

gsmet:
I'll fix the typos and merge once CI is green.

Sanne:
I really meant "sequencer" but ok it's the same ;)

gsmet:
@Sanne it just feels a bit weird to have the entity named `Sequence**d**Actor` and the generator `sequence**r**ActorGenerator`.

I think we should be consistent. If you agree, pick your choice, I'll do the leg work. If not, just push it :).

Sanne:
ok, applied your suggestions and pushed it.

PR #798, opened 2016-11-11
OGM-1193 Test generation of values for sequences is thread safe

DavideD:
https://hibernate.atlassian.net/browse/OGM-1193
https://hibernate.atlassian.net/browse/OGM-1207
https://hibernate.atlassian.net/browse/OGM-1208

I wanted to move some tests in core to check that we are generating values for sequences in a thread safety way. By doing so I discovered some issues in CouchDB, Redis and MongoDB.

I couldn't make this test pass with Inifinispan embedded. @Sanne can you have a look at it?
I copied the test you added for Hot Rod but I've added a change to make sure that all the values are generated as expected.


Sanne:
Feel free to remove the Hot Rod specific version.

Let's also add some documentation clarifications, e.g. that Redis will store ids with an offset (essentially it ignores the initialvalue option)

Sanne:
You haven't yet documented the fact that Redis stores ids using an offset, i.e. it ignores the _initialvalue_?

DavideD:
I'm writing it now

DavideD:
I'll ping you when I think the documentation is ready

DavideD:
@Sanne I think I've done with the documentation and this PR

documentation/manual/src/main/asciidoc/en-US/modules/architecture.asciidoc
```diff
@@ -234,6 +234,31 @@ when the schema is changed as the schema is right in front of the developers' ey
...
+* [classname]`@TableGenerator`:
+  it will emulate sequences storing the value in the most appropriate data structure;
+  for example a document in MongoDB or a node in Neo4j.
+
+Here's some things to keep in mind when dealing with sequence generation:
+
+* [classname]`@TableGenerator` is the fallback approach used when the underlying datastore does not
+  support native sequences generation.
+* If the datastore does not support atomic operations and does not support native sequences,
+  Hibernate OGM will throw an exception.
```
Sanne: This sounds worse than the reality. Could you clarify that the expcertion is thrown at bootstrap, and that it will suggest to use a different ID generation strategy for this dialect?

documentation/manual/src/main/asciidoc/en-US/modules/architecture.asciidoc
```diff
@@ -234,6 +234,31 @@ when the schema is changed as the schema is right in front of the developers' ey
...
+* The mapping of the sequence might change based on the annotation used, you should check the
+  mapping paragraph in the documentation related to the dialect you are using.
+* The value saved in the the datastore might not be the next value in the sequence. The reason
+  is that some datastores don't provide all the operations we need to update and create sequences
+  reliably. In this cases, the value should be considered as a seed that Hibernate OGM
```
Sanne: I wouldn not try to explain why we decided this, it sounds like we're making excuses.

DavideD:
@Sanne updated

DavideD:
Thanks, it's nice to have this fixed and tested now

DavideD:
I'll apply this to the 5.0 branch as well, is that all right?

Sanne:
sounds good!

PR #801, opened 2016-11-21
OGM-1210 Add discriminator value as label for a node in Neo4j

DavideD:
https://hibernate.atlassian.net/browse/OGM-1210

THis patch will add a new label the the node when it's created but won't remove the discriminator property. THis way the mapping is still backward compatible but a user can still execute native queries using the more natural approach.

This would be an intermediate step before removing the property in the next major release.

@Sanne @gunnarmorling WDYT?

### DIFF
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BaseNeo4jDialect.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BaseNeo4jDialect.java
@@ -10,11 +10,17 @@ import static org.hibernate.ogm.datastore.neo4j.query.parsing.cypherdsl.impl.Cyp
 import static org.hibernate.ogm.datastore.neo4j.query.parsing.cypherdsl.impl.CypherDSL.skip;
 import static org.hibernate.ogm.util.impl.EmbeddedHelper.isPartOfEmbedded;
 
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 
+import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.ogm.datastore.neo4j.dialect.impl.BaseNeo4jAssociationQueries;
+import org.hibernate.ogm.datastore.neo4j.dialect.impl.BaseNeo4jEntityQueries;
 import org.hibernate.ogm.datastore.neo4j.dialect.impl.BaseNeo4jTypeConverter;
+import org.hibernate.ogm.datastore.neo4j.dialect.impl.EntityKeyMetadataWithDiscriminator;
 import org.hibernate.ogm.datastore.neo4j.query.impl.Neo4jParameterMetadataBuilder;
 import org.hibernate.ogm.dialect.multiget.spi.MultigetGridDialect;
 import org.hibernate.ogm.dialect.query.spi.BackendQuery;
@@ -29,6 +35,7 @@ import org.hibernate.ogm.dialect.spi.DuplicateInsertPreventionStrategy;
 import org.hibernate.ogm.dialect.spi.OperationContext;
 import org.hibernate.ogm.dialect.spi.SessionFactoryLifecycleAwareDialect;
 import org.hibernate.ogm.dialect.spi.TupleContext;
+import org.hibernate.ogm.dialect.spi.TupleTypeContext;
 import org.hibernate.ogm.model.key.spi.AssociatedEntityKeyMetadata;
 import org.hibernate.ogm.model.key.spi.AssociationKey;
 import org.hibernate.ogm.model.key.spi.AssociationKeyMetadata;
@@ -38,8 +45,12 @@ import org.hibernate.ogm.model.key.spi.RowKey;
 import org.hibernate.ogm.model.spi.Association;
 import org.hibernate.ogm.model.spi.Tuple;
 import org.hibernate.ogm.model.spi.TupleSnapshot;
+import org.hibernate.ogm.persister.impl.OgmCollectionPersister;
+import org.hibernate.ogm.persister.impl.OgmEntityPersister;
 import org.hibernate.ogm.type.spi.GridType;
 import org.hibernate.ogm.util.impl.ArrayHelper;
+import org.hibernate.persister.collection.CollectionPersister;
+import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.service.spi.ServiceRegistryAwareService;
 import org.hibernate.service.spi.ServiceRegistryImplementor;
 import org.hibernate.type.Type;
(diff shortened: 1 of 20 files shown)


neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/EntityKeyMetadataWithDiscriminator.java
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.datastore.neo4j.dialect.impl;
+
+import org.hibernate.ogm.model.key.spi.EntityKeyMetadata;
+
+/**
+ * This class is an {@link EntityKeyMetadata} that kepp tracks of the discriminator value required to keep track of
```
gsmet: s/kepp tracks/keep track/

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/EntityKeyMetadataWithDiscriminator.java
```diff
@@ -0,0 +1,92 @@
...
+ * entities whne dealing with single table per class inheritance.
```
gsmet: s/whne/when/

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/EntityKeyMetadataWithDiscriminator.java
```diff
@@ -0,0 +1,92 @@
...
+ * <p>
+ * Normally, Hibernate ORM add a discriminator column and use the value in it to select the right tuples. Neo4j can
```
gsmet: `adds` and `uses`

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/EntityKeyMetadataWithDiscriminator.java
```diff
@@ -0,0 +1,92 @@
...
+ * keepo track of the different entities using labels, therfore the discriminator value becomes another metdata
```
gsmet: s/keepo/keep/
s/therfore/therefore/
s/metdata/metadata/

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/EntityKeyMetadataWithDiscriminator.java
```diff
@@ -0,0 +1,92 @@
+/*
```
gsmet: In the commit message:
s/this queries needs now/these queries now need/
s/therefor/therefore/

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BaseNeo4jDialect.java
```diff
@@ -10,11 +10,17 @@
 import static org.hibernate.ogm.datastore.neo4j.query.parsing.cypherdsl.impl.CypherDSL.skip;
```
gsmet: In the commit message:
s/discriminatoir/discriminator/
s/Enitty/Entity/
s/is is/is/
s/remnove/remove/

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/BaseNeo4jQueries.java
```diff
@@ -64,31 +65,51 @@ protected static void appendProperties(StringBuilder queryBuilder, String[] colu
...
 		queryBuilder.append( ":" );
 		appendLabel( entityKeyMetadata, queryBuilder );
+		if ( addDiscriminatorLabel ) {
+			appendDiscriminatorLabels( tupleTypeContext, queryBuilder );
+		}
 		appendProperties( queryBuilder, entityKeyMetadata.getColumnNames(), offset );
 		queryBuilder.append( ")" );
 	}
 
+	public static void appendDiscriminatorLabels(TupleTypeContext tupleTypeContext, StringBuilder queryBuilder) {
```
gsmet: s/appendDiscriminatorLabels/appendDiscriminatorLabel/
(there is only one label)

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/BaseNeo4jQueries.java
```diff
@@ -64,31 +65,51 @@ protected static void appendProperties(StringBuilder queryBuilder, String[] colu
...
-	protected static void appendEntityNode(String alias, EntityKeyMetadata entityKeyMetadata, StringBuilder queryBuilder) {
-		appendEntityNode( alias, entityKeyMetadata, queryBuilder, 0 );
+	protected static void appendEntityNode(String alias, EntityKeyMetadata entityKeyMetadata, TupleTypeContext tupleTypeContext, StringBuilder queryBuilder) {
+		appendEntityNode( alias, entityKeyMetadata, tupleTypeContext, queryBuilder, 0, false );
 	}
 
-	protected static void appendEntityNode(String alias, EntityKeyMetadata entityKeyMetadata, StringBuilder queryBuilder, int offset) {
+	/*
+	 * At the moment, we want to add the discriminator value as a label in Neo4j, but we still want to keep compatibility with the old mapping.
+	 * For this reason we don't to consider the discriminator label when executing a select.
```
gsmet: s/don't to/don't/

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/EmbeddedNeo4jDialect.java
```diff
@@ -83,72 +77,34 @@
...
-			if ( collectionPersister instanceof OgmCollectionPersister ) {
-				OgmCollectionPersister ogmCollectionPersister = (OgmCollectionPersister) collectionPersister;
-				EntityKeyMetadata ownerEntityKeyMetadata = ( (OgmEntityPersister) ( ogmCollectionPersister.getOwnerEntityPersister() ) ).getEntityKeyMetadata();
-				AssociationKeyMetadata associationKeyMetadata = ogmCollectionPersister.getAssociationKeyMetadata();
-				queryMap.put( associationKeyMetadata, new EmbeddedNeo4jAssociationQueries( ownerEntityKeyMetadata, associationKeyMetadata ) );
-			}
-		}
-		return queryMap;
+	@Override
+	protected EmbeddedNeo4jAssociationQueries createAssociationQueries(EntityKeyMetadata ownerEntityKeyMetadata, TupleTypeContext ownerTupleTypeContext,
```
gsmet: You will probably hate me for that but I would have added the `TupleTypeContext` parameter consistently at the beginning as it's more general than the `EntityKeyMetadata`.

DavideD:
@gsmet You are right, it's missing the intermediate labels. I did some re-work on the part related to hierarchies so I will fix this pull request after the other one is integrated.

DavideD:
I'll close it for now

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/EmbeddedNeo4jDialect.java
DavideD: As far as I can see, the signature of basically every method we have put the key and/or key metadata before a context.
The only one that it is different is forEachTuple and the reason is that initially it was supposed to have multiple EntityKeys.

The way I see it is that the first value tells you which entities are considered and the context gives you some additional information to get them.

PR #802, opened 2016-11-25
OGM-732 Queries with entities having InheritanceType.SINGLE_TABLE

DavideD:
https://hibernate.atlassian.net/browse/OGM-732

This patch make it possible to run queries on entity hierarchies when mapped with InheritanceType.SINGLE_CLASS. This wasn't working before for Neo4j and MongoDB.

I haven't implemented the TABLE_PER_CLASS approach, so , for now, it throws an exception

See also https://github.com/hibernate/hibernate-ogm/pull/803 for branch 5.0

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/util/impl/Log.java
+++ b/core/src/main/java/org/hibernate/ogm/util/impl/Log.java
@@ -13,6 +13,7 @@ import static org.jboss.logging.Logger.Level.WARN;
 
 import java.io.Serializable;
 import java.lang.annotation.ElementType;
+import java.util.Collection;
 
 import javax.persistence.PersistenceException;
 import javax.transaction.SystemException;
@@ -301,4 +302,6 @@ public interface Log extends BasicLogger {
 	@Message(id = 88, value = "Configuration is referring to deprecated datastore provider name '%1$s'. Please use the new form '%2$s' instead.")
 	void usingDeprecatedDatastoreProviderName(String deprecatedName, String newName);
 
+	@Message(id = 89, value = "%1$s does not support queries on polymorphic entities using TABLE_PER_CLASS inheritance strategy. You should try using SINGLE_TABLE instead. Entities: %2$s")
+	HibernateException queriesOnPolymorphicEntitiesAreNotSupportedWithTablePerClass( String datastore, Collection<String> subclassEntityNames );
 }
--- a/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/inheritance/SingleTableInheritanceTest.java
+++ b/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/inheritance/SingleTableInheritanceTest.java
@@ -1,0 +1,154 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.datastore.mongodb.test.inheritance;
+
+import static org.fest.assertions.Assertions.assertThat;
+
+import java.util.List;
+
+import javax.persistence.EntityManager;
+
+import org.hibernate.ogm.utils.jpa.OgmJpaTestCase;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * @author Davide D'Alto
+ */
+public class SingleTableInheritanceTest extends OgmJpaTestCase {
+
+	private Node parent;
+	private Node simpleChild;
+	private TextNode textChild;
+	private NodeLink linkToSimple;
+	private NodeLink linkToText;
+
+	@Before
+	public void setUp() {
+		EntityManager em = getFactory().createEntityManager();
+		try {
+			em.getTransaction().begin();
+
+			parent = new SimpleNode();
+			parent.setName( "root" );
+
+			simpleChild = new SimpleNode();
+			simpleChild.setName( "children 1" );
+
+			textChild = new TextNode();
+			textChild.setName( "children 2" );
+			textChild.setText( "a text" );
+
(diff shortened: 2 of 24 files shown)


neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/query/parsing/CypherQueryParsingTest.java
```diff
@@ -0,0 +1,273 @@
...
+import org.hibernate.ogm.datastore.neo4j.test.query.parsing.model.inheritance.CommunityMemberST;
+import org.hibernate.ogm.datastore.neo4j.test.query.parsing.model.inheritance.EmployeeST;
+import org.hibernate.ogm.datastore.neo4j.test.query.parsing.model.inheritance.PersonST;
+import org.hibernate.ogm.utils.OgmTestCase;
+import org.hibernate.ogm.utils.parser.MapBasedEntityNamesResolver;
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * Integration test for {@link org.hibernate.ogm.datastore.mongodb.query.parsing.impl.MongoDBQueryResolverDelegate} and
```
gsmet: It's not related to MongoDB!

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/query/parsing/CypherQueryParsingTest.java
```diff
@@ -0,0 +1,273 @@
...
+ * {@link org.hibernate.ogm.datastore.mongodb.query.parsing.impl.MongoDBQueryRendererDelegate}.
+ *
+ * @author Gunnar Morling
```
gsmet: s/Gunnar Morling/Davide D'Alto/?
(looks like you rewrote everything)

core/src/main/java/org/hibernate/ogm/util/impl/Log.java
```diff
@@ -301,4 +302,6 @@
 	@Message(id = 88, value = "Configuration is referring to deprecated datastore provider name '%1$s'. Please use the new form '%2$s' instead.")
 	void usingDeprecatedDatastoreProviderName(String deprecatedName, String newName);
 
+	@Message(id = 89, value = "%1$s does not support queries on polymorfic entities using TABLE_PER_CLASS inheritance strategy. You should try using SINGLE_TABLE instead. Entities: %2$s")
```
gsmet: s/polymorfic/polymorphic/

core/src/main/java/org/hibernate/ogm/persister/impl/SingleTableOgmEntityPersister.java
```diff
@@ -37,6 +38,10 @@ public SingleTableOgmEntityPersister(
 	private static EntityDiscriminator resolveDiscriminator(final PersistentClass persistentClass, final SessionFactoryImplementor factory) {
 		if ( persistentClass.isPolymorphic() ) {
 			Value discrimValue = persistentClass.getDiscriminator();
+			if ( discrimValue == null ) {
+				throw new MappingException( "discriminator mapping required for single table polymorphic persistence" );
```
DavideD: I copied this from Hibernate ORM, not sure how much we want to they similar to what they do

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/inheritance/SingleTableInheritanceTest.java
```diff
@@ -0,0 +1,130 @@
...
+		}
+	}
+
+	@Test
+	public void testSubclassReturnedByNativeQuery() throws Exception {
+		EntityManager em = getFactory().createEntityManager();
+		try {
+			em.getTransaction().begin();
+			@SuppressWarnings("unchecked")
+			List<TextNode> resultList = em.createNativeQuery( "db.Node.find({'DTYPE': 'TextNode'})", TextNode.class ).getResultList();
```
DavideD: Well, the expected return type is a parameter of the createNativeQuery method, so yes.

I'm not sure what happens if one doesn't specified the returned type. I think it returns an array of objects

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/query/parsing/CypherQueryParsingTest.java
DavideD: I should stop using  copy and paste :)

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/inheritance/SingleTableInheritanceTest.java
gsmet: Yes, my remark was more to outline that when using the DTYPE query directly, we don't return the possible subclasses. I think we can't do better but I wanted to be sure I understood the limitations right.

DavideD:
I push-forced all the changes and removed the confusing part in OgmLoader.
It wasn't needed and the main goal was to align the behaviour to ORM. I will keep it for another pull request.

I think now, the patch is much simpler and we can actually apply it to the 5.0 branch.
It seems odd that people cannot execute queries for entities using SINGLE_TABLE inheritance strategy.

I think we have implemented all the features needed for a final release. My plan now is to release a Beta2, align class names and improve the documentation if needed and then make another release before a CR1 and a Final.

Let's see how it goes, we received quite some feedback recently and I hope the trend will continue.

gsmet:
Nice work! Merged!

I'm wondering if we should add some documentation on what's supported or not depending on the backend. I remember you updating the doc on inheritance recently. Maybe a table with what's supported by each backend would be in order? WDYT?

DavideD:
Yes, I need to update the documentation on several front. I was waiting for this PR to be integrated.
I will update the doc working on https://hibernate.atlassian.net/browse/OGM-1216

gsmet:
OK cool, glad we're on the same page! Closing then.

PR #803, opened 2016-11-25
OGM-732 For branch 5.0

DavideD:
Same as https://github.com/hibernate/hibernate-ogm/pull/802
but for the 5.0 branch


gsmet:
Let's validate it for master first. We'll see about 5.0 then.

It's a bit of a massive change and I'm not sure it should be backported. Are we far from the next major release?

DavideD:
Update this pul request as well

DavideD:
> Let's validate it for master first. We'll see about 5.0 then.
> 
> It's a bit of a massive change and I'm not sure it should be backported. Are we far from the next major release?

I've answered on the other PR, but in short for others:
1. the PR is now much smaller and it contains mainly tests
2. I think it's not nice that people see errors when using queries with inheritance considering that we can fix it

gsmet:
Applied to 5.0 as well. Thanks for working on the backport!

PR #804, opened 2016-12-13
OGM Dialect for Apache Ignite

Z-z-z-z:
Here is dialect for Apache Ignite (http://ignite.apache.org/features/datagrid.html)


DavideD:
How come when I run a test in Eclipse, it opens an H" console on the browser? Any Idea?

DavideD:
This PR does not pass the tests on CI:
http://ci.hibernate.org/view/Pull%20Requests/job/hibernate-ogm-PR/1122/

Is this expected or there is something we need to setup?

Z-z-z-z:
What is it *H" console* ? Or you mean Eclipse opens new browser window and
put console output there?

2016-12-13 13:35 GMT+03:00 DavideD <notifications@github.com>:

Z-z-z-z:
http://ci.hibernate.org/view/Pull%20Requests/job/hibernate-
ogm-PR/1122/console
Tests run: 235, Failures: 58, Errors: 31, Skipped: 11

It means that my module passed 62% of tests.

Should I mark tests that I've not yet implemented with @Ignore annotation?


2016-12-13 14:35 GMT+03:00 Victor Z <vikad111@gmail.com>:

DavideD:
> Should I mark tests that I've not yet implemented with @Ignore annotation?

No, some of those tests should pass. In particular the one related to the queries.

DavideD:
Ill complete my review in a bit and I'll add some details about it

DavideD:
> What is it *H" console* ? Or you mean Eclipse opens new browser window and
put console output there?

I mean the H2 console in the browser window: http://www.h2database.com/html/images/quickstart-4.png

DavideD:
I finished my review, here a summary of the things that should change:

1. Figure out why when I run the test in eclipse H2 open a web console on the browser. This happens because H2 is a transitive dependency of ignite-indexing but I'm not sure if it's needed to begin with and it 's certainly doing something wrong

2. Make the build work on CI (or Travis), You might want to create an account on [Travis](https://travis-   ci.org). This way the build will run every time you push something on your remote personal repository.
It's not enough to ginore all the not working tests because, based on your implementation, some are expected to work;  These three definitely should:
    * SimpleQueriesTest
    * EmbeddableTest 
    * ManyToManyExtraTest

3. Remove these classes:
    * IgniteHibernateException
    * IgniteSingleTableEntityPersister
    * IgniteOgmCollectionPersister
    * IgniteOgmPersisterClassResolver
    * IgniteServiceRegistryInitializer

    These are not needed you only need because the OgmEntityPersister and similar already interact with    the GridDialect and the DatastoreProvider

4. There's something wrong with the query implementation, I already explained it in another comment so I won't repeat it here.  

5. This is actually a question: Is it possible to configure the database programmatically and remove the ignite-default.xml?

6 Exceptions should use the Log class and the same approach we use for the other dialects. Some of the reasons behind this are:
   * better performance
   * Internationalization
   * all the exception will have a code that help us dealing with user requests

Let me know if you have other questions and thank for the help, I think it is going in the right direction.

Z-z-z-z:
1.

   *Figure out why when I run the test in eclipse H2 open a web console on
   the browser. This happens because H2 is a transitive dependency of
   ignite-indexing but I'm not sure if it's needed to begin with and it 's
   certainly doing something wrong*

As far as I know Ignite use H2 engine to implement SQL queries. I can't
remove it.

I search source code of Ignite and found this in class
org.apache.ignite.internal.processors.query.h2.IgniteH2Indexing:

            if (getString(IGNITE_H2_DEBUG_CONSOLE) != null) {
                Connection c = DriverManager.getConnection(dbUrl);

                WebServer webSrv = new WebServer();
                Server web = new Server(webSrv, "-webPort", "0");
                web.start();
                String url = webSrv.addSession(c);

                try {
                   * Server.openBrowser(url);*
                }
                catch (Exception e) {
                    U.warn(log, "Failed to open browser: " +
e.getMessage());
                }
            }

I think it is your problem is here or may be you have some database plugins
installed.

2016-12-13 18:10 GMT+03:00 DavideD <notifications@github.com>:

bom/pom.xml
```diff
@@ -634,13 +643,47 @@
...
+                <version>${igniteVersion}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.ignite</groupId>
+                <artifactId>ignite-jta</artifactId>
+                <version>${igniteVersion}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.ignite</groupId>
+                <artifactId>ignite-indexing</artifactId>
```
Z-z-z-z: They use H2 to implement SQL queries

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/IgniteDialect.java
```diff
@@ -0,0 +1,780 @@
...
+		if ( po != null ) {
+			return new Tuple( new IgniteTupleSnapshot( id, po, key.getMetadata() ), SnapshotType.UPDATE );
+		}
+		else {
+			return null;
+		}
+	}
+
+//	private Tuple createTuple(EntityKey key, OperationContext operationContext, Object id, BinaryObject found) {
+//		if ( found != null ) {
```
Z-z-z-z: done

ignite/src/test/resources/ignite-config.xml
```diff
@@ -0,0 +1,749 @@
...
+  limitations under the License.
+-->
+
+<beans xmlns="http://www.springframework.org/schema/beans"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns:util="http://www.springframework.org/schema/util"
+       xsi:schemaLocation="
+        http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
+        http://www.springframework.org/schema/util http://www.springframework.org/schema/util/spring-util.xsd">
+    <bean class="org.apache.ignite.configuration.IgniteConfiguration">
```
Z-z-z-z: I've already done it. Look at IgniteTestConfigurationBuilder.java
I forget to delete xml config file. Done

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/IgnitePessimisticReadLockingStrategy.java
```diff
@@ -0,0 +1,83 @@
...
+		TypeTranslator typeTranslator = lockable.getFactory().getServiceRegistry().getService( TypeTranslator.class );
+		GridType idGridType = typeTranslator.getType( lockable.getIdentifierType() );
+		EntityKey key = EntityKeyBuilder.fromData( ( (OgmEntityPersister) lockable ).getRootEntityKeyMetadata(),
+				idGridType,
+				id,
+				session );
+
+		IgniteCache<Object, BinaryObject> cache = provider.getEntityCache( key.getMetadata() );
+		if ( cache == null ) {
+			throw new IgniteHibernateException( "Cache " + key.getMetadata().getTable() + " is not found" );
```
Z-z-z-z: This check is unnecessary. I removed it

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/exception/IgniteHibernateException.java
```diff
@@ -0,0 +1,25 @@
...
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.datastore.ignite.exception;
+
+import org.hibernate.HibernateException;
+
+public class IgniteHibernateException extends HibernateException {
```
Z-z-z-z: done

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/IgniteDialect.java
```diff
@@ -0,0 +1,780 @@
...
+		else {
+			if ( backendQuery.getSingleEntityMetadataInformationOrNull() != null ) {
+				return new IgnitePortableFromProjectionResultCursor(
+								result,
+								queryParameters.getRowSelection(),
+								backendQuery.getSingleEntityMetadataInformationOrNull().getEntityKeyMetadata()
+							);
+			}
+			else {
+				throw new UnsupportedOperationException( "Not implemented yet" );
```
Z-z-z-z: David  method executeBackendQuery works with native queries. There is no HQL here.
I use static method IgniteHqlQueryParser.createParameterList

DavideD: OK, my bad, but there is still something wrong in the whole queries are implemented.

You shouldn't have an `IgniteHqlQueryParser`.
The method `IgniteQueryParserService#parseQuery` should create an instance should look something like:
```
		org.hibernate.hql.QueryParser queryParser = new QueryParser();
		IgniteProcessingChain processingChain = createProcessingChain( sessionFactory );
		IgniteQueryParsingResult result = queryParser.parseQuery( queryString, processingChain );

		return result;
```

where `IgniteProcessingChain extends AstProcessingChain<IgniteQueryParsingResult>`
and `IgniteQueryParsingResult extends QueryParsingResult`.

DavideD: I think we might actually have a SQL parser already implemented, I need to check the project as I'm not sure and I cannot find it right now. I'll have a look tomorrow

Z-z-z-z: David, I'll look again at the query implementation. We made it a long time ago. May be some things should be made in different way.
By the way, we don't use own implementation of SQL or HQL parser. The aim is to make a little changes in default SQL query.

Z-z-z-z:
Hi!
I have a question about ManyToManyExtraTest.
Now in fails on this line
	assertThat( getNumberOfAssociations( sessionFactory ) ).isEqualTo( expectedAssociationNumber() );
With error: "org.junit.ComparisonFailure: expected:<[2]> but was:<[3]>"
My implementation of many to many association is classic SQL implementation, and I think that 3 is correct. 
I looked at Ehcache implementation. They store list of Student ids in one value for each ClassRoom id. 
Do you have some kind of requirement about this sort of asociation?

Z-z-z-z:
You asked me to remove these classes:
- IgniteSingleTableEntityPersister
- IgniteOgmPersisterClassResolver
- IgniteServiceRegistryInitializer

I can't do this because I have to override method
`public String filterFragment(String alias, Map enabledFilters, Set<String> treatAsDeclarations)`
it have to return an empty string instead of null.
The reason is here: AbstractEntityPersister.java
`return sessionFilterFragment.append( filterFragment( alias, treatAsDeclarations ) ).toString();`
now it returns string "null"

schernolyas:
Привет @Z-z-z-z  !
В апреле будет конференция JPoint и я хочу рассказать про Hibernate OGM. Если тебе это интересно то свяжись со мной пожалуйста.

DavideD:
> You asked me to remove these classes:
> 
> IgniteSingleTableEntityPersister
> IgniteOgmPersisterClassResolver
> IgniteServiceRegistryInitializer
> I can't do this because I have to override method
> public String filterFragment(String alias, Map enabledFilters, Set<String> treatAsDeclarations)
> it have to return an empty string instead of null.
> The reason is here: AbstractEntityPersister.java
> return sessionFilterFragment.append( filterFragment( alias, treatAsDeclarations ) ).toString();
> now it returns string "null"

I'll have another look at this

DavideD:
> My implementation of many to many association is classic SQL implementation, and I think that 3 is correct.
> I looked at Ehcache implementation. They store list of Student ids in one value for each ClassRoom id.
> Do you have some kind of requirement about this sort of asociation?

The getNumberOfAssociations(...) should return the number of associations created, in this test it should be one association with role `students` for "math" and one association with role  `english`. Note that we are not counting the number of elements in the association. At least, I think that the reason the expected result is 2 and not 3.

I notice now that there 2 errors in this test:
- Neo4j should not have a different expected returned value
- One student is added twice i the same classroom

Z-z-z-z:
> The getNumberOfAssociations(...) should return the number of associations created, in this test it should be one association with role students for "math" and one association with role english. Note that we are not counting the number of elements in the association. At least, I think that the reason the expected result is 2 and not 3.

Are you sure about "we are not counting the number of elements in the association" ?
Here is Ehcache:
`	public long getNumberOfAssociations(SessionFactory sessionFactory) {
		int asscociationCount = 0;
		Set<Cache<?>> processedCaches = Collections.newSetFromMap( new IdentityHashMap<Cache<?>, Boolean>() );

		for ( CollectionPersister collectionPersister : ( (SessionFactoryImplementor) sessionFactory ).getCollectionPersisters().values() ) {
			Cache<?> associationCache = getAssociationCache( sessionFactory, ( (OgmCollectionPersister) collectionPersister ).getAssociationKeyMetadata() );
			if ( !processedCaches.contains( associationCache ) ) {
				asscociationCount += associationCache.getSize();
				processedCaches.add( associationCache );
			}
		}

		return asscociationCount;
	}
`
It seems to me that it is exactly number of elements. MongoDB dialect the same

DavideD:
If I remember correctly EhCache stores associations using associationKey as key and a map as value, the map contains the elements of  the association but when you use the size you have the number of associations. I'll double check and let you know

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/IgniteDialect.java
Z-z-z-z: Hi!
I didn't find an implementation of AstProcessingChain that make SQL. Hibernate Core use another way to do this. I can't implement all SQL queries by myself. It's a huge work. 
May be my code is slightly dirty but it's easy to change it when hibernate-core use new HQL parser

DavideD:
> If I remember correctly EhCache stores associations using associationKey as key and a map as value, the map contains the elements of the association but when you use the size you have the number of associations. I'll double check and let you know

I gave a look at the code and it seems I was right. I think it's clear when one looks at EhcacheDialect#createAssociation.
It first gets the association cache and then inserts a new element using the AssociationKey as key and a map as value. 
```
final Cache<AK> associationCache = ...
Map<SerializableRowKey, Map<String, Object>> association = ...
AK cacheKey = getKeyProvider().getAssociationCacheKey( key )
associationCache.put( new Element(  cacheKey, association ) );
```

`association` is the map containing all the elements of the association using the `RowKey` to identify them.
There are also some diagrams in the documentation: https://docs.jboss.org/hibernate/ogm/5.1/reference/en-US/html_single/#associations_3

ignite/src/test/java/org/hibernate/ogm/datastore/ignite/utils/IgniteTestConfigurationBuilder.java
```diff
@@ -0,0 +1,416 @@
...
+					.build()
+		);
+		cacheConfig.add(
+				createCacheConfig( "Car_Tire" )
+					.appendIndex( "cars_carId_maker", String.class )
+					.appendIndex( "cars_carId_model", String.class )
+					.appendIndex( "tires_tireId_maker", String.class )
+					.appendIndex( "tires_tireId_model", String.class )
+					.build()
+		);
```
DavideD: Isn't it possible to move this logic in the IgniteCacheInitializer?

DavideD: Great job removing the XML by the way

ignite/src/test/java/org/hibernate/ogm/datastore/ignite/utils/IgniteTestConfigurationBuilder.java
```diff
@@ -0,0 +1,416 @@
...
+	private IgniteConfiguration loadFromResource(String resourceName) throws IgniteCheckedException {
+		URL url = getClass().getClassLoader().getResource( resourceName );
+		return IgnitionEx.loadConfiguration( url ).get1();
+	}
+
+	private IgniteConfiguration createConfig() {
+		IgniteConfiguration config = new IgniteConfiguration();
+
+		config.setGridName( "OgmTestGrid" );
+		config.setLocalHost( "127.0.0.1" );
```
DavideD: You should read this properties from the configuration file, the host and the port might be different when running on CI

ignite/src/test/java/org/hibernate/ogm/datastore/ignite/utils/IgniteTestConfigurationBuilder.java
DavideD: @Z-z-z-z I have a question, is it possible to create a cache an a server that it is already started?

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/IgniteDialect.java
```diff
@@ -0,0 +1,738 @@
...
+				}
+			}
+
+			BinaryObjectBuilder binaryObjectBuilder = provider.createBinaryObjectBuilder( binaryObject );
+			binaryObjectBuilder.setField( column, associationObjects.toArray( new BinaryObject[ associationObjects.size() ] ) );
+			binaryObject = binaryObjectBuilder.build();
+			associationCache.put( id, binaryObject );
+		}
+
+		association.reset();
```
DavideD: how come you need this reset?

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/impl/IgniteDatastoreProvider.java
```diff
@@ -0,0 +1,398 @@
...
+			BinaryObjectBuilder builder = createBinaryObjectBuilder( findKeyType( key.getMetadata() ) );
+			for ( int i = 0; i < key.getColumnNames().length; i++ ) {
+				builder.setField( StringHelper.stringAfterPoint( key.getColumnNames()[i] ), key.getColumnValues()[i] );
+			}
+			result = builder.build();
+		}
+		return result;
+	}
+
+	public Object createAssociationKeyObject( RowKey rowKey, AssociationKeyMetadata keyMetadata ) {
```
DavideD: I think there is something wrong in the creation of the association key, it seems that you are not considering the entitykey and you risk to insert an element once even if it connected to two different entities.

I think this is the reason ManyToManyExtraTest fails, the student `mario` is inserted in the class math and english classes but when you count the element in the associations the end result is 3 because the association is only recorded for one of the classes. @Z-z-z-z Can you investaigate this?

DavideD: I Suspect this causes other errors as well

ignite/src/test/java/org/hibernate/ogm/datastore/ignite/utils/IgniteTestConfigurationBuilder.java
Z-z-z-z: > I have a question, is it possible to create a cache an a server that it is already started?
Yes. It is possible. A cache appears when it fistly requested. The only problem is that you can't create cache after you begin transaction. That's why a prefer to have a static cache configuration.

Z-z-z-z: >  Isn't it possible to move this logic in the IgniteCacheInitializer?

No. I don't think so. This class is for real work not for tests.
And why? What's wrong with IgniteTestConfigurationBuilder?

ignite/src/test/java/org/hibernate/ogm/datastore/ignite/utils/IgniteTestConfigurationBuilder.java
Z-z-z-z: I don't think it's a good idea. 
I made configuration specially to run tests locally.
In case with remote server we need two configurations client and server, not a property file. 
IMHO it's too complicated.

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/IgniteDialect.java
Z-z-z-z: I remove it.

ignite/src/test/java/org/hibernate/ogm/datastore/ignite/utils/IgniteTestConfigurationBuilder.java
DavideD: It's not really about IgniteTestConfigurationBuilder, I'm wondering if the configuration of the cache can be created at start up by OGM using the information from the existing annotation. This is the approach we use in OGM, unless it is no feasable for some reasons. And that's also the idea behind the SchemaDefiner interface. For example, we already know that those columns are identifeirs of an association and therefore we could create an index when we create the cache.

This way the user does not have to provide a configuration for the caches unless he needs some more customization that we don't provide out of the box or at elast he nedds to provide less information.

For example. in Neo4j we add unique constraints when a property is mapped using `@Column(unique=true)`. Not all datastores allow this and I'm wondering if it's feasable with Ignite.

I'm not saying that you have to do everuthing now, though

Z-z-z-z:
You were right about query implementation. Your version is match more clear and elegant,
I rewrote my queries. Take a look please.
There is no support for projections and assosiations but I hope that it is enough for the first version.

ignite/src/test/java/org/hibernate/ogm/datastore/ignite/utils/IgniteTestConfigurationBuilder.java
DavideD: OK, but sometimes we need to connect to different ports because we run several different things on CI, would it be possible to get the LocalHost from the property `hibernate.ogm.datastore.host`?

DavideD:
Thanks a lot, that's great.
I don't have the energy to look at it right now, I'll check it tomorrow.

> There is no support for projections and assosiations but I hope that it is enough for the first version.

Yes, that's OK for a ffirst version

ignite/src/test/java/org/hibernate/ogm/datastore/ignite/utils/IgniteTestConfigurationBuilder.java
Z-z-z-z: Of course it is possible, but property LocalHost allows you to select one local IP address to listen to. It's not for connecting to remote host. For real configurations I made interface IgniteConfigurationBuilder.
Let's fix problems when we have them.

ignite/src/test/java/org/hibernate/ogm/datastore/ignite/utils/IgniteTestConfigurationBuilder.java
Z-z-z-z: Hmmm... Now I understand you point. 
IMHO it's not very useful in the real word. I've always had storage structure at first and then made classes to work with it. I think that you able to create database from classes only inside HelloWorld-programs or unit tests
But I'll think how to do this.

Z-z-z-z:
Hi, David!
I really want to include Ignite dialect to next OGM release.
Now I have only one issue ManyToManyExtraTest. Is it critical or I can fix it later?
What else I have to do to achieve the purpose?

DavideD:
> I really want to include Ignite dialect to next OGM release.
> Now I have only one issue ManyToManyExtraTest. Is it critical or I can fix it later?
> What else I have to do to achieve the purpose?

If I'm right about why the test is failing it seems quite a critical issue. I explained it in a previous comment but the dialect is basically loosing association entries. Also I have many other tests failing right now, that might have similar issue when counting the associations.

ignite/src/main/java/org/hibernate/ogm/datastore/ignite/query/parsing/impl/IgniteQueryRendererDelegate.java
```diff
@@ -0,0 +1,242 @@
...
+
+	private void select(StringBuilder queryBuilder) {
+		queryBuilder.append( "SELECT _KEY, _VAL " );
+	}
+
+	private void from(StringBuilder queryBuilder) {
+		String tableAlias = propertyHelper.findAliasForType( targetTypeName );
+		OgmEntityPersister persister = (OgmEntityPersister) ( sessionFactory ).getEntityPersister( targetType.getName() );
+		String tableName = propertyHelper.getKeyMetaData( targetType.getName() ).getTable();
+		queryBuilder.append( " FROM " ).append( tableName ).append( ' ' ).append( tableAlias ).append( ' ' );
```
DavideD: Is that `tableName` also the name of the cache? If this is the case, you can  keep track of the value in the IgniteQueryDescriptor and use it in the dialect when `EntityMetadataInformation` is null (I think, I haven't test it).

DavideD: This way you might be able to support projections as well

Z-z-z-z:
> If I'm right about why the test is failing it seems quite a critical issue. I explained it in a previous comment but the dialect is basically loosing association entries. Also I have many other tests failing right now, that might have similar issue when counting the associations.

Ok. I'll check it again. But you don't answer my main question:
What else I have to do to include Ignite dialect to next OGM release?
Documentaion, WildFly module, anything else?

DavideD:
Documentation for sure. It also needs to pass our checks on CI, this means
that the tests that are now failing need to be fixed or skipped. You can
see some examples on how we do that for MongoDB in SimpleQueriesTest
(annotation @SkipbyGridDialect(...)).

We can add later the module for WildFly.

These are sort of the minimum requirements. The mapping should be fine but
I'm having difficulties checking the content of the caches, during the
tests; is there any tool that I can use to do it (apart from ignitevisor)?

Z-z-z-z:
> I'm having difficulties checking the content of the caches, during the tests; is there any tool that I can use to do it (apart from ignitevisor)?

I'm using visorui from GridGain Enterprise Edition (https://www.gridgain.com/resources/download) or my own web console.
There is also  Ignite Web Console (https://apacheignite.readme.io/docs/web-console) but I don't know is it works or not.

Z-z-z-z:
> Documentation for sure. It also needs to pass our checks on CI, this means that the tests that are now failing need to be fixed or skipped. You can see some examples on how we do that for MongoDB in SimpleQueriesTest (annotation @SkipbyGridDialect(...)).

Than I have some kind of roadmap:

- fix ManyToManyExtraTest 
- mark all failed test with annotation @SkipbyGridDialect
- make documentation 

Correct?

DavideD:
Yes, It would also be nice if can also rebase it to the latest master.

Z-z-z-z:
There are some problems in Jenkins that I can't solve.  I need a help, please.

Z-z-z-z:
Now I have another problem in Jankins. 
Radis tests fail with error:
`Caused by: com.lambdaworks.redis.RedisConnectionException: Unable to connect to localhost/127.0.0.1:6379
`
May be I changed something, but I don't understand what. 

Could anybody take I look what's wrong with Radis?

Thank you

DavideD:
I'll have a look next week

Z-z-z-z:
Hm... Now it's ok.

**Done:**
- fix ManyToManyExtraTest                  
- mark all failed test with annotation @SkipbyGridDialect 

Get another code review if it's necessary

**In progress**
- make documentation

Z-z-z-z:
I added chapter into manual. Now my roadmap is complete.

Could you please make a review.

Z-z-z-z:
Hi, David!
Problem with radis again
`Caused by: java.net.ConnectException: Connection refused: localhost/127.0.0.1:6379`

I need your help.

DavideD:
Jenkins, retest this please

Z-z-z-z:
Hi, David!
Will you approve this pull request?

DavideD:
Hi @Z-z-z-z, I'll be away for a week at a work-related meeting where we will discuss about OGM. I'll try to wrap-up this PR when I'm back. 

Thanks and sorry for my late replies.

Z-z-z-z:
David?

Z-z-z-z:
Hi @DavideD ! 
Does label "Next release" mean that you are going to merge this PR to release 5.2.0?

DavideD:
Hi @Z-z-z-z,
sorry for the radio silence. Yes, I talked with the team and we decided to create a separate repository where to put most of the dialects and where rules for the integration of dialcets is more relaxed compared to the core project. We want to give contributors more freedom about how to work an a dialect without necessarily waiting for us  to make a review (as yo uare already noticed, we are very slow when it comes to that). 

OrientDB, Apache Ignite and other existing dialects will end in the same repository. I'm working on it right now.

schernolyas:
HI @Z-z-z-z and @DavideD !

I working with the dialect. Firstly, I will up to date the dialect. Later, I will prepare new PR (I can send commit to the PR). Also, current version not supports dynamic schema generation. I will fix it in next releases

Z-z-z-z:
Hi @schernolyas!
This dialect supports dynamic schema generation. But of couse you can try to improve it.

schernolyas:
Hi @Z-z-z-z !

I mean the schema configuration for tests. https://github.com/schernolyas/hibernate-ogm/blob/OGM-983-ApacheIgnite-DEV1/ignite/src/test/java/org/hibernate/ogm/datastore/ignite/utils/IgniteTestConfigurationBuilder.java#L87:L173

Z-z-z-z:
@schernolyas me too

DavideD:
I've created a new repsiroty for the Apache Ignite Dialect on my account: https://github.com/DavideD/hibernate-ogm-ignite

It will required a snpashot of Hibernate OGM built using the following branch: https://github.com/hibernate/hibernate-ogm/pull/850

I've created the repository adding a squashed version of this pr because it was too hard to move the commits and keep the history. It would be nice if you can have a l;ook and make sure I didn't forget anything.

I will move the repository in the Hibernate account once my pull request has been integrated

Z-z-z-z:
I made the pull request to your branch https://github.com/DavideD/hibernate-ogm-ignite/pull/1

Z-z-z-z:
Hi!

Is there any news about the new repository for Ignite dialect?

DavideD:
Hi @Z-z-z-z,
sorry, I didn't have much time to work on this but the situation is better now.

You should see some progress soon.

DavideD:
I moved everything in the http://github.com/hibernate/hibernate-ogm-ignite repository

PR #810, opened 2017-01-03
OGM-1191 Remove dead code warning in OgmLoader

DavideD:
With the current code the resultSet can never be null, if that happens,
  you have NullPointerException before calling the method.

  I prefer to remove the code for now and eventually put the check back if we
  have a use case where it makes sense.


DavideD:
https://hibernate.atlassian.net/browse/OGM-1191

PR #817, opened 2017-01-10
OGM-1221 Null pointer in Neo4j with SINGLE_TABLE inheritance strategy

DavideD:
https://hibernate.atlassian.net/browse/OGM-1221

  Remote dialects create queries at start-up and use the entity key metadata as key of
  a map to find the right set of queries for an entity.

  Problem is that when we deal with single table per class inheritance the entity key metadata is
  the same for all the entities in the hierarchy. This was causing an error, because the queries
  created for an entities were replaced with the queries generated for another one.

  The code has also been refactored because it was starting to get too complicated to copy and
  paste fixes around the 3 different dialects.

  Note that the problem occured only with remote dialects because the embedded one execute some
  operations with the programmatic API.

  This bug was found by the user Mariusz Florek

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/dialect/impl/TupleTypeContextImpl.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/impl/TupleTypeContextImpl.java
@@ -26,6 +26,9 @@ public class TupleTypeContextImpl implements TupleTypeContext {
 	private final List<String> selectableColumns;
 	private final OptionsContext optionsContext;
 
+	private final String discriminatorColumn;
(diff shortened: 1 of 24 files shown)


gsmet:
Jenkins, retest this please

gsmet:
@DavideD looks good. The only question I have is if we should consider changing EntityKeyMetadata to include the discriminator value instead of having another class.

I think it would make sense. It's probably not something we want to do now but maybe we should create a new issue with a fixVersion 6? WDYT?

gsmet:
Merged, thanks!

PR #819, opened 2017-01-27
OGM-1238 Keep order of collections of elements in MongoDB

DavideD:
https://hibernate.atlassian.net/browse/OGM-1238

The mongodb driver returns collection of elements in the same order they appear in the DB, we should do the same.

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/datastore/document/association/spi/AssociationRows.java
+++ b/core/src/main/java/org/hibernate/ogm/datastore/document/association/spi/AssociationRows.java
@@ -6,8 +6,6 @@
  */
 package org.hibernate.ogm.datastore.document.association.spi;
 
-import static org.hibernate.ogm.util.impl.CollectionHelper.newHashMap;
-
 import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
@@ -17,6 +15,7 @@ import org.hibernate.ogm.model.key.spi.RowKey;
 import org.hibernate.ogm.model.spi.AssociationSnapshot;
 import org.hibernate.ogm.model.spi.Tuple;
 import org.hibernate.ogm.model.spi.Tuple.SnapshotType;
+import org.hibernate.ogm.util.impl.CollectionHelper;
 
 /**
  * Represents the rows of an association in form of {@link AssociationRow}s.
@@ -28,7 +27,7 @@ public class AssociationRows implements AssociationSnapshot {
 	private final Map<RowKey, AssociationRow<?>> rows;
 
 	public AssociationRows(AssociationKey associationKey, Collection<?> wrapped, AssociationRowFactory associationRowFactory) {
-		this.rows = newHashMap( wrapped.size() );
+		this.rows = CollectionHelper.newHashMap( wrapped.size() );
 
 		for ( Object object : wrapped ) {
 			AssociationRow<?> row = associationRowFactory.createAssociationRow( associationKey, object );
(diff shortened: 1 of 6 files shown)


DavideD:
Jenkins, retest this please

core/src/main/java/org/hibernate/ogm/datastore/document/association/spi/AssociationRows.java
```diff
@@ -28,7 +27,7 @@
 	private final Map<RowKey, AssociationRow<?>> rows;
 
 	public AssociationRows(AssociationKey associationKey, Collection<?> wrapped, AssociationRowFactory associationRowFactory) {
-		this.rows = newHashMap( wrapped.size() );
+		this.rows = new LinkedHashMap<>( wrapped.size() );
```
gsmet: Please add a new `newLinkedHashMap` method to our CollectionHelper.

I'll work on pushing the changes we made in HV regarding the sizing.

(With the current code, you always undersize the map)

core/src/main/java/org/hibernate/ogm/model/spi/Association.java
```diff
@@ -184,7 +184,7 @@ public int size() {
 		}
 		else {
 			// It may be a bit too large in case of removals, but that's fine for now
-			Set<RowKey> keys = new HashSet<RowKey>( cleared ? currentState.size() : snapshot.size() + currentState.size() );
+			Set<RowKey> keys = new LinkedHashSet<RowKey>( cleared ? currentState.size() : snapshot.size() + currentState.size() );
```
gsmet: Same here. Let's create a `newLinkedHashSet` method in `CollectionHelper` and I'll optimize it.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/embeddable/ElementCollectionOrderWithSessionTest.java
```diff
@@ -0,0 +1,139 @@
...
+import org.hibernate.ogm.utils.TestForIssue;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * @author Davide D'Alto
+ */
+@TestForIssue(jiraKey = "OGM-1238")
+public class ElementCollectionOrderWithSessionTest extends OgmTestCase {
```
gsmet: Is it really worth it to add 2 tests for that?

DavideD: I think so, we are testing two different interfaces and the test classes are simple. I don't see where's the problem. Potentially, comsething could happen between EntityManager and Session that causes a test to fail.

core/src/main/java/org/hibernate/ogm/datastore/document/association/spi/AssociationRows.java
DavideD: Perhaps, but this is preexisting code, why stopping this PR that tries to solve a different issue?

DavideD:
Closing this PR until I fixed a couple of things

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/embeddable/ElementCollectionOrderWithSessionTest.java
gsmet: OK, it's just that I wasn't sure it was useful.

DavideD:
jenkins, retest this please

PR #820, opened 2017-01-31
OGM-1240, OGM-1241 MongoDB authentication fails

DavideD:
https://hibernate.atlassian.net/browse/OGM-1240
https://hibernate.atlassian.net/browse/OGM-1241

MongoDB uses an authentication database to verify the credentials, this might differ fomr the database we want to connect. I've added a new property with default value to admin.

THe only problem that I have with this issue is that it is complicated to test with a job.
I've created [OGM-1242](https://hibernate.atlassian.net/browse/OGM-1242) to keep track of it.

I've also upgraded the driver in the process.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -57,7 +57,7 @@
         <ehcacheVersion>2.6.11</ehcacheVersion>
 
         <!-- MongoDB -->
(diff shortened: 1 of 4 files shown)


gsmet:
Merged, thanks!

PR #821, opened 2017-02-01
Fix master

DavideD:
Currently master is not building because of the documentation.

I don't see anything odd in the latest changes I committed but for some reasons the build fails.
It seems that if I update the reosurces plugin everything works fine.


Sanne:
I had the same problem, this fixes it... merged

PR #822, opened 2017-02-01
Search upgrade

Sanne:
- https://hibernate.atlassian.net/browse/OGM-1243
 - https://hibernate.atlassian.net/browse/OGM-1244
 - https://hibernate.atlassian.net/browse/OGM-1245

The Arquillian idea to introduce proper isolation failed; maybe it could work but I gave up as it wasn't giving me enough control, and it was extremely slow to boot (20+ seconds per test).

I wish I didn't have to, but ultimately I ended up learning DMR and boostrapping the container "manually" ..


integrationtest/src/test/resources/hotrod-client-testingconfiguration-offset100.properties
```diff
@@ -0,0 +1,25 @@
...
+infinispan.client.hotrod.default_executor_factory.pool_size = 1
+infinispan.client.hotrod.default_executor_factory.queue_size = 100
+infinispan.client.hotrod.tcp_no_delay = true
+infinispan.client.hotrod.tcp_keep_alive = false
+
+#TODO could we estimate these in advance from our knowledge of the data model?
+infinispan.client.hotrod.key_size_estimate = 64
+infinispan.client.hotrod.value_size_estimate = 512
+#This is critical: [FIXME check for this to be set?]
+infinispan.client.hotrod.force_return_values = true
```
DavideD: Is this necessary? I think we already validate the expected values in `InfinispanRemoteConfiguration`. Just asking because, if that's the case, the comment is misleading

infinispan-remote/pom.xml
```diff
@@ -168,10 +145,88 @@
...
+            <artifactId>infinispan-server-build</artifactId>
+            <version>${infinispanVersion}</version>
+            <type>zip</type>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
```
DavideD: What is this? :)

Sanne: I think it's an error of the Maven pom of that artifact.. it attempts to transitively bring in all its dependencies. Including the Hibernate Search version we needed to get rid of..

integrationtest/src/test/resources/hotrod-client-testingconfiguration-offset100.properties
Sanne: I don't remember. It's a copy of the other configuration file we use for testing - I just changed the port to match the port offset (+100).

If the comment is wrong, then the existing configuration files should be fixed as well.

PR #823, opened 2017-02-02
OGM-1024 Support aggregate operation for MongoDB CLI syntax

DavideD:
https://hibernate.atlassian.net/browse/OGM-1024

This is something that has been required  a lot and that I had almost ready on one of my local branches.

### DIFF
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/impl/MongoDBQueryDescriptor.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/impl/MongoDBQueryDescriptor.java
@@ -12,6 +12,7 @@ import static org.hibernate.ogm.datastore.mongodb.query.impl.MongoDBQueryDescrip
 import static org.hibernate.ogm.datastore.mongodb.query.impl.MongoDBQueryDescriptor.Operation.UPDATE;
 
 import java.io.Serializable;
+import java.util.Collections;
 import java.util.List;
 
 import com.mongodb.DBObject;
@@ -33,7 +34,14 @@ public class MongoDBQueryDescriptor implements Serializable {
 		REMOVE,
 		UPDATE,
 		COUNT,
-		AGGREGATE;
+		/**
+		 * This is used by the query parser when the parsed query requires an aggregation, usually for embedded collections.
+		 */
+		AGGREGATE,
+		/**
+		 * This is used for native queries, when the user wants to execute a generic aggregation query.
+		 */
+		AGGREGATE_PIPELINE;
 	}
 
 	private final String collectionName;
@@ -57,6 +65,19 @@ public class MongoDBQueryDescriptor implements Serializable {
 	 */
 	private final DBObject options;
 	private final List<String> unwinds;
+	private final List<DBObject> pipeline;
+
+	public MongoDBQueryDescriptor(String collectionName, Operation operation, List<DBObject> pipeline) {
+		this.collectionName = collectionName;
+		this.operation = operation;
+		this.criteria = null;
+		this.projection = null;
+		this.orderBy = null;
+		this.options = null;
+		this.updateOrInsert = null;
+		this.unwinds = null;
+		this.pipeline = pipeline == null ? Collections.<DBObject>emptyList() : pipeline;
+	}
 
 	public MongoDBQueryDescriptor(String collectionName, Operation operation, DBObject criteria, DBObject projection, DBObject orderBy, DBObject options, DBObject updateOrInsert, List<String> unwinds) {
 		this.collectionName = collectionName;
@@ -67,6 +88,11 @@ public class MongoDBQueryDescriptor implements Serializable {
 		this.options = options;
 		this.updateOrInsert = updateOrInsert;
 		this.unwinds = unwinds;
+		this.pipeline = Collections.<DBObject>emptyList();
+	}
+
+	public List<DBObject> getPipeline() {
+		return pipeline;
 	}
 
 	/**
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
@@ -6,6 +6,13 @@
  */
 package org.hibernate.ogm.datastore.mongodb.query.parsing.nativequery.impl;
 
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Deque;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
 import org.hibernate.ogm.datastore.mongodb.query.impl.MongoDBQueryDescriptor;
 import org.hibernate.ogm.datastore.mongodb.query.impl.MongoDBQueryDescriptor.Operation;
 import org.hibernate.ogm.util.impl.StringHelper;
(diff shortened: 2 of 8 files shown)


gsmet:
Jenkins, retest this please

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
```diff
@@ -100,4 +152,37 @@ public DBObject parse(String json) {
...
+	private static DBObject stage(String key, Object value) {
+		DBObject stage = new BasicDBObject();
+		stage.put( key, value );
+		return stage;
+	}
+
+	public boolean addPipeline(StackElement element, String value) {
+		if ( !parsed.contains( element.getIndex() ) ) {
+			parsed.add( element.getIndex() );
+			boolean isJsonObject = value.trim().matches( "^\\{.*:.*\\}$" );
```
gsmet: Why don't you use the `parse` method directly for this? It was added exactly to deal with this and avoid this kind of tests.

It would avoid the regexp (which would have to be statically compiled anyway but let's get rid of it).

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
```diff
@@ -100,4 +152,37 @@ public DBObject parse(String json) {
...
+			}
+			else {
+				BasicDBObject object = BasicDBObject.parse( "{ 'json': " + value + "}" );
+				pipeline.add( stage( normalized( element ), object.get( "json" ) ) );
+			}
+		}
+		return true;
+	}
+
+	private String normalized(StackElement element) {
```
gsmet: Maybe rename it `normalizeOperation`?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
```diff
@@ -92,12 +94,13 @@ public Rule Separator() {
 	}
 
 	public Rule Reserved() {
-		return FirstOf( Find(), FindOne(), FindAndModify(), Insert(), Remove(), Update(), Count() );
-		//TODO There are many more query types than what we support.
+		return FirstOf( Aggregate(), Find(), FindOne(), FindAndModify(), Insert(), Remove(), Update(), Count() );
```
gsmet: Let's add it at the end instead of at the beginning. It's less common than the other operations.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBSessionCLIQueryTest.java
```diff
@@ -305,6 +323,114 @@ public void testFindWithAnd() throws Exception {
...
+			assertThat( result ).onProperty( "id" ).containsExactly( imperatrix.getId(), athanasia.getId() );
+
+			transaction.commit();
+			session.clear();
+		}
+	}
+
+	@Test
+	@TestForIssue(jiraKey = "OGM-1024")
+	public void testAggregateWithMatchSortAndRegexWithFirstResultAndMAxResults() {
```
gsmet: s/M**A**xResults/MaxResults/

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/impl/MongoDBQueryDescriptor.java
```diff
@@ -57,6 +65,20 @@
...
+		this.collectionName = collectionName;
+		this.operation = operation;
+		this.criteria = null;
+		this.projection = null;
+		this.orderBy = null;
+		this.options = null;
+		this.updateOrInsert = null;
+		this.unwinds = null;
+		this.pipeline = pipeline == null ? Collections.<DBObject>emptyList() : pipeline;
+
```
gsmet: Useless new line.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
```diff
@@ -92,12 +94,13 @@ public Rule Separator() {
...
+		// TODO There are many more query types than what we support.
 	}
 
 	public Rule Operation() {
 		return FirstOf(
+				Sequence( Aggregate(), builder.setOperation( Operation.AGGREGATE_PIPELINE ) ),
```
gsmet: Same here, let's move it at the end.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
```diff
@@ -100,4 +152,37 @@ public DBObject parse(String json) {
...
+		return element.getOperation().replaceAll( "'", "" ).replaceAll( "\"", "" ).trim();
+	}
+
+	public boolean push(int index, java.lang.String match) {
```
gsmet: We don't need `java.lang.`.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBSessionCLIQueryTest.java
```diff
@@ -305,6 +323,114 @@ public void testFindWithAnd() throws Exception {
...
+			List<OscarWildePoem> result = query.list();
+
+			assertThat( result ).onProperty( "id" ).containsExactly( portia.getId() );
+
+			transaction.commit();
+			session.clear();
+		}
+	}
+
+	@Test
```
gsmet: You added a parser test with a nested aggregation but I think it would be safer if we also add a test here, really executing the query on the MongoDB server.

gsmet:
Btw, the test are failing because of Cassandra probably not responding on slave2.

DavideD:
> I wonder if we could try to support distinct and mapReduce at the same time?

I created an issue for mapReduce already as somebody asked for it on the forum. I would wait for 5.1.1 unless we can find an asy way to implement it. I will add an issue for distinct, it seems relatively easier to add, I'll see if I have time

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
DavideD: The parse method doesn't work for some cases because it expects the return value to be a DBObject but now sometimes it returns a String. I will change the approach

gsmet: Missed that.

I think we should rename the existing one `parseDBObject` and have the `parse` method returning an object. `parseDBObject` would call `parse` and cast the result.

WDYT?

gsmet:
OK to keep the issues separate!

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
DavideD: Yep, that's what I started to do

DavideD:
Remarks applied

gsmet:
Merged, thanks!

PR #825, opened 2017-02-07
OGM-1248 Switch to Asciidoctor for the final documentation output

gsmet:
* https://hibernate.atlassian.net/browse/OGM-1248


gsmet:
@DavideD I just force pushed an update following a comment of Yoann in the Search PR.

gsmet:
@DavideD any issue for reviewing this? It would be nice to have it in before any documentation update or we will have conflicts!

gsmet:
OK, thanks!

distribution/src/main/assembly/dist.xml
```diff
@@ -246,7 +246,7 @@
             </excludes>
         </fileSet>
         <fileSet>
-            <directory>../documentation/manual/target/docbook/publish</directory>
+            <directory>../documentation/manual/target/asciidoctor</directory>
```
DavideD: Instead of  `asciidoctor`, can we call it `output` or `asciidoctor-output` or something similar?

Not important obviously, I just find it harder to figure out what the folder contains from the name.

gsmet: Yeah, I must admit I'm a bit reluctant to do last minute changes that can break something when they are not strictly necessary as I tested the whole chain for all the projects and I'd like to keep them consistent.

gsmet:
@DavideD it's ready to go, I force pushed the last change I wanted to do.

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/825

Thanks

PR #826, opened 2017-02-09
OGM-1247 : Support distinct native queries

raghav-jha:


Sanne:
ok to test

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -900,6 +903,25 @@ private static DBObject stage(String key, Object value) {
 		return stage;
 	}
 
+	private static ClosableIterator<Tuple> doDistinct(
+			final MongoDBQueryDescriptor queryDescriptor, final DBCollection collection) {
+		List distinctFieldValues = collection.distinct( queryDescriptor.getFieldName(), queryDescriptor.getCriteria() );
```
DavideD: I see that the distinct method also has a ReadPreference option, could you check if it make sense to pass the argument returned by `tupleContext.getTupleTypeContext().getOptionsContext().getUnique( ReadPreferenceOption.class );`

raghav-jha:
@DavideD 
How do I squash your commits into mine ???

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBSessionCLIQueryTest.java
```diff
@@ -598,4 +606,24 @@ public void testInsertMultipleWithNumberLongThenRemove() throws Exception {
...
+			Transaction transaction = session.beginTransaction();
+
+			String nativeQuery = "db." + OscarWildePoem.TABLE_NAME + ".distinct(\"name\",{\"author\":\"Oscar Wilde\"})";
+
+			Query query = session.createNativeQuery( nativeQuery );
+
+			List result = (List) session.createNativeQuery( nativeQuery ).uniqueResult();
+
+
+			assertThat( result.size() ).isEqualTo( 3 );
```
raghav-jha: Separate test means ??

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
```diff
@@ -37,7 +37,7 @@
 	private String criteria;
 	private String projection;
 	private String orderBy;
-
+	private String fieldName;
```
raghav-jha: Wouldn't it be better to rename it as distinctFieldName???

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBSessionCLIQueryTest.java
DavideD: A separate method, now you have `testDistinctQuery()`, you need to add `testDistinctQueryWithoutCriteria()` for example. See what we have done for Aggregate in the same class

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
DavideD: That works as well

DavideD:
> @DavideD
> How do I squash your commits into mine ???

Probably the simplest way is to download the diff: https://github.com/DavideD/hibernate-ogm/commit/0a3a8392c805d5579471af62a9433e6a52333c49.diff

apply it with `git apply 0a3a8392c805d5579471af62a9433e6a52333c49.diff` and then `git amend`

Alternativly you can cherry-pick the commit on my repository and then rebase it as explained here: http://stackoverflow.com/questions/2563632/how-can-i-merge-two-commits-into-one

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
raghav-jha: The tuple context is not available, probably because we are dealing with more than a single entity type.

DavideD: OK

raghav-jha:
`collation: {
   locale: <string>,
   caseLevel: <boolean>,
   caseFirst: <string>,
   strength: <int>,
   numericOrdering: <boolean>,
   alternate: <string>,
   maxVariable: <string>,
   backwards: <boolean>
}`

So for collation, I am thinking it to just them parse as options, and leave it to mongo db to throw syntax exception if something is wrong.  What do you think @DavideD, this approach is correct ??

DavideD:
Sounds good, it similar to what I did for the aggregation pipeline.

raghav-jha:
@DavideD : I have added collation support as well. Please review!!!

Thanks

DavideD:
See https://github.com/hibernate/hibernate-ogm/pull/829

I've squashed the commits in this PR and created a new one.

Thanks a lot

PR #827, opened 2017-02-11
OGM-913 Allows join on collection of elements

DavideD:
https://hibernate.atlassian.net/browse/OGM-913

This only works for Neo4j and MongoDB at the moment, I'll create a new issue for the lucene parser since it requires some changes in the HQL parser project

### DIFF
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/impl/MongoDBPropertyHelper.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/impl/MongoDBPropertyHelper.java
@@ -14,7 +14,11 @@ import org.hibernate.hql.ast.spi.PropertyHelper;
 import org.hibernate.ogm.datastore.mongodb.MongoDBDialect;
 import org.hibernate.ogm.persister.impl.OgmEntityPersister;
 import org.hibernate.ogm.query.parsing.impl.ParserPropertyHelper;
+import org.hibernate.ogm.type.spi.GridType;
+import org.hibernate.ogm.type.spi.TypeTranslator;
 import org.hibernate.ogm.util.impl.StringHelper;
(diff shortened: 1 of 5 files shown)


mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/impl/MongoDBPropertyHelper.java
```diff
@@ -36,6 +43,27 @@ public String getColumnName(String entityType, List<String> propertyPath) {
...
+			return ( (CollectionType) propertyType ).getElementType( sessionFactory );
+		}
+		return propertyType;
+	}
+
+	@Override
+	public Object convertToBackendType(String entityType, List<String> propertyPath, Object value) {
+		Type propertyType = getPropertyType( entityType, propertyPath );
+		if ( isElementCollection( propertyType ) ) {
+			// For colleciton of elements we return the type of the collection
```
gsmet: s/colleciton/collection/

DavideD:
@gsmet I'll reorder the commits. I usually do that but I guess I forgot this time

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/impl/MongoDBPropertyHelper.java
DavideD: Fixed

DavideD:
I fixed the commits and pushed-forced.

The commits should be in the right order, first the fix and then the test.

gsmet:
@DavideD ok, thanks. I'm not very confident in the order GitHub displays the commits, especially after a force push, so I prefer if you push it directly when you're done!

Thanks!

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/827

PR #834, opened 2017-02-24
OGM-1253 MapWithEmbedded

gsmet:
@Davide here is your PR with a couple of typos fixed.

I was about to merge it but the Travis build failed and it looks like there is an issue with remote Neo4j:
```
Running org.hibernate.ogm.backendtck.embeddable.ElementCollectionWithMapAndEmbeddedTest
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.548 sec <<< FAILURE! - in org.hibernate.ogm.backendtck.embeddable.ElementCollectionWithMapAndEmbeddedTest
testName(org.hibernate.ogm.backendtck.embeddable.ElementCollectionWithMapAndEmbeddedTest)  Time elapsed: 0.374 sec  <<< FAILURE!
java.lang.AssertionError: the map:<{}> does not contain the entries:<['issue1'=OGM-1, 'issue2'=OGM-2]>
	at org.hibernate.ogm.backendtck.embeddable.ElementCollectionWithMapAndEmbeddedTest.testName(ElementCollectionWithMapAndEmbeddedTest.java:54)
```
both with HTTP and Bolt.

Could you take a look and base your work on these fixed commits? Note that you can push commits to my branch.

Thanks!

PR #835, opened 2017-02-27
OGM-1253 Test Map<String, Embedded> and update documentation

DavideD:
Follows up for https://github.com/hibernate/hibernate-ogm/pull/834

@gsmet  This should fix the problem and avoid the creation of an embedded node without properties.

I've also created a common class and removed HttpNeo4jTupleAssociationSnapshot and BoltNeo4jTupleAssociationSnpashot


gsmet:
@DavideD hmm, there was something weird with the last 3 commits: they weren't ordered as in the GitHub interface and the commits using the common class were coming before the one creating it (at least locally when I checked out your branch). To be on the safe side, I squashed the 3 commits related to the addition of this new class and the removal of the old ones. It's the same logical unit so I think it made sense anyway.

Tests are running on Travis, I'll merge it as soon as it's green.

gsmet:
@DavideD so I pushed it to master.

I'm a bit worried because I have weird failures on Travis with Bolt:
```
testSetElementCollectionStorageAndRemoval(org.hibernate.ogm.backendtck.associations.collection.types.MapTest)  Time elapsed: 0.14 sec  <<< ERROR!
org.hibernate.AssertionFailure: Unrecognized associationKind: EMBEDDED_COLLECTION
	at org.hibernate.ogm.backendtck.associations.collection.types.MapTest.testSetElementCollectionStorageAndRemoval(MapTest.java:86)
```

but, after looking at the code, I really can't see how these failures are possible... we'll see how it goes on our CI.

DavideD:
That's weird.

Thanks, we will see from here what happens

PR #838, opened 2017-02-28
Fixes

DavideD:
@gsmet This should fix the problems with the Bolt job.

I made a silly mistake with the latest PR and that's the reason of the weird error on Travis, I also noticed that something was off with the authentication that leads me to refactor the test a bit.


neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BoltNeo4jDialect.java
```diff
@@ -488,8 +488,10 @@ private void putAssociationOperation(AssociationKey associationKey, AssociationO
 		switch ( associationKey.getMetadata().getAssociationKind() ) {
 			case EMBEDDED_COLLECTION:
 				createRelationshipWithEmbeddedNode( associationKey, associationContext, action );
+				break;
```
gsmet: Raaah, I read this code 2 times without seeing the issue...

DavideD: Same for me. Weird, isn't it?
It's good to have tests :)

gsmet:
Merged, thanks!

PR #842, opened 2017-03-03
[PREVIEW] working on migration to 3.4 api

schernolyas:
HI @DavideD !

I change Eclipse to Idea and prepare first part of changes. 
I just changed DBObject to Document. look my PR please.
Need do:
1) continue migration
2) correct tests
3) check and fix tests

to be continued

### DIFF
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
@@ -24,7 +24,6 @@ import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 
-import com.mongodb.AggregationOutput;
 import com.mongodb.bulk.BulkWriteResult;
 import com.mongodb.client.AggregateIterable;
 import com.mongodb.client.FindIterable;
@@ -116,7 +115,6 @@ import org.hibernate.ogm.type.impl.CharacterStringType;
 import org.hibernate.ogm.type.impl.StringCalendarDateType;
 import org.hibernate.ogm.type.spi.GridType;
 import org.hibernate.ogm.util.impl.CollectionHelper;
-import org.hibernate.ogm.util.impl.StringHelper;
 import org.hibernate.type.StandardBasicTypes;
 import org.hibernate.type.Type;
 import org.parboiled.Parboiled;
@@ -127,7 +125,6 @@ import org.parboiled.support.ParsingResult;
 import com.mongodb.DuplicateKeyException;
 import com.mongodb.ReadPreference;
 import com.mongodb.WriteConcern;
-import com.mongodb.WriteResult;
 
 /**
  * Each Tuple entry is stored as a property in a MongoDB document.
@@ -376,10 +373,10 @@ public class MongoDBDialect extends BaseGridDialect implements QueryableGridDial
 	}
 	private MongoCollection<Document> getCollection(String table,ReadPreference readPreference,WriteConcern writeConcern) {
 		MongoCollection<Document> collection = currentDB.getCollection( table );
-		if (readPreference!=null) {
+		if ( readPreference != null ) {
 			collection.withReadPreference( readPreference );
 		}
-		if (writeConcern!=null) {
+		if ( writeConcern != null ) {
 			collection.withWriteConcern( writeConcern );
 		}
 		return collection;
@@ -685,7 +682,7 @@ public class MongoDBDialect extends BaseGridDialect implements QueryableGridDial
 
 				// if there is a single column on the value side left, unwrap it
 				if ( row.keySet().size() == 1 ) {
-					rows.put( rowKeyValue, DocumentUtil.toMap(row).values().iterator().next() );
+					rows.put( rowKeyValue, DocumentUtil.toMap( row ).values().iterator().next() );
 				}
 				else {
 					rows.put( rowKeyValue, row );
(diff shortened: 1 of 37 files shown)


schernolyas:
Hi @DavideD !

I have refactored main classes. 
I see that the PR is better that previus and ... I hope You can make review.

schernolyas:
Hi @DavideD !
Code in the branch is compile only. but ... not works.
Please, review the changes and mark them. I will fix tests and your remarks.

Also ... not close the branch  please

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -257,46 +265,49 @@ private DBObject getEmbeddingEntity(AssociationKey key, AssociationContext assoc
...
 
-		BasicDBObject projection = getProjection( tupleContext );
+		Document projection = getProjection( tupleContext );
 
-		DBObject query = new BasicDBObject();
-		query.put( ID_FIELDNAME, new BasicDBObject( "$in", searchObjects ) );
-		return collection.find( query, projection );
+		Document query = new Document();
+		query.put( ID_FIELDNAME, new Document( "$in", searchObjects ) );
+		//return collection.find( query, projection );
```
DavideD: Just remove the old code, don't comment it, please

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1199,9 +1239,9 @@ private static AssociationStorageStrategy getAssociationStorageStrategy(Associat
 	public void executeBatch(OperationsQueue queue) {
 		if ( !queue.isClosed() ) {
 			Operation operation = queue.poll();
-			Map<DBCollection, BatchInsertionTask> inserts = new HashMap<DBCollection, BatchInsertionTask>();
+			Map<MongoCollection<Document>, BatchInsertionTask> inserts = new HashMap<MongoCollection<Document>, BatchInsertionTask>();
 
-			List<Tuple> insertTuples = new ArrayList<Tuple>();
+			List<Tuple> insertTuples = new LinkedList<Tuple>();
```
DavideD: Why did you replace an ArrayList with a LinkedList?

DavideD:
> Also ... not close the branch please

@schernolyas The way we work is that we usually send a PR when we think it's ready to be merged, we don't send partial pull requests that don't compile or build. The exception to this rule is when the committer needs some kind of feedback. Once he has the feedback it will close the PR. There is no reason to keep a PR that cannot be merged open while you are working on it.

I cannot review your PR every time you do a single or some changes, it takes too much time.
The overall approach looks good, just reopen it or send a different one from a different branch when everything works (it compiles, builds and pass the tests). It takes 2 seconds for you to create a new branch so I don't see the problem. You can use [Travis](https://travis-ci.org/) to test your work on an external CI.

Also, it's not alright to add 3 commits to a PR and call all of them `migration to new API`, the way we describe a commit an a PR is `JIRACODE - Description`. Please, try to stick to our conventions and use a more descriptive approach. For example:

```
OGM-768 Migrate MongoDB dialect to new 3.4 API
  
  Replacing the deprecated method from the mongo driver in preparation for the GridFS support:

  * Replace DBObject with Document
  * Replace DBCollection with MongoCollection
```

Remember to delete the commented code before sending the PR, as @Sanne already mentioned before, we don't usually keep commented code around unless there is a good reason to (and you should explain the reason in the comment).

Feel free to ask questions if you have some problems with making the tests pass but don't send a PR each time, remember that if we need to check something we can always take a look at your branch.

Thanks for you help.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
schernolyas: HI @DavideD 
Because count of elements not defined. In this way LinkedList more preferable.

DavideD: > Because count of elements not defined. In this way LinkedList more preferable.

So this change has nothing to do with migrating the API, keep it in a separate commit, please

Sanne: It's a great example of why unrelated changes should be sent in a different pull request:

it's not true that a LinkedList would be more suited in this case, or at least that's my opinion so - if I were to review this PR - I would start debating about this change.
Such a debate would delay the integration of a pull request which might otherwise contain nice changes. In extreme cases, having many such "distractions" would get to endless delays, and possibly conflicts which get harder to resolve when things aren't merged quickly. It's possible that large PRs with many off-topic changes never make it to be integrated..

schernolyas: Dear @DavideD  and @Sanne !

Ok. I will restore it.

Sanne: thanks @schernolyas ! appreciate it. I hope you don't consider me being too harsh, such practices are necessary in large remote / OSS teams.. I hope you'll understand.

Personally when I had once the wrong formatting template, I simply thrown the branch away and started over.
I am not sure how large & complex your changes have been, but maybe that's easier..

schernolyas: hi @Sanne !

I will restore ArrayList. it is not a problem. 
Broken formatting .... yes You right.  Easy to restart more carefully.

About migration to 3.4 api ... It is large and complex change. And ... of course I can analyse and do self. But ... I hope to help of members because they are authors.

PR #847, opened 2017-03-14
OGM-1272 Infinispan configuration handling is making caches transacti…

Sanne:
…onal as side-effect

 - https://hibernate.atlassian.net/browse/OGM-1272


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/847

Thanks

PR #848, opened 2017-03-14
OGM-588 Revert work-around required due to ORM version update

Sanne:
should be trivial, but I couldn't test all DBs. Relying on CI for that.

https://hibernate.atlassian.net/browse/OGM-588


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/848

PR #849, opened 2017-03-15
OGM-786 migration to MongoDB Java Driver 3.x API

schernolyas:
The PR contains changes for migration from 2.x API to 3.x API.


DavideD:
There is the following error on CI (
http://ci.hibernate.org/job/hibernate-ogm-PR/1259/console):
```
[ERROR] -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.5.1:compile (default-compile) on project hibernate-ogm-performancetest: Compilation failure
/mnt/jenkins-workdir/workspace/hibernate-ogm-PR/performance/src/main/java/org/hibernate/ogm/perftest/mongodb/ogm/EntityManagerFactoryHolder.java:[72,47] cannot find symbol
  symbol:   method dropDatabase()
  location: interface com.mongodb.client.MongoDatabase
```

Can you have a quick look? It seems everything else is building, well done!

schernolyas:
Hi @DavideD !

Problem have fixed. Recheck please

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
```diff
@@ -105,6 +104,8 @@ public Rule Operation() {
 				Sequence( FindOne(), builder.setOperation( Operation.FINDONE ) ),
 				Sequence( FindAndModify(), builder.setOperation( Operation.FINDANDMODIFY ) ),
 				Sequence( Insert(), builder.setOperation( Operation.INSERT ) ),
+				Sequence( InsertOne(), builder.setOperation( Operation.INSERTONE ) ),
+				Sequence( InsertMany(), builder.setOperation( Operation.INSERTMANY ) ),
 				Sequence( Remove(), builder.setOperation( Operation.REMOVE ) ),
```
schernolyas: @DavideD  please see to [MongoCollection](http://api.mongodb.com/java/current/com/mongodb/client/MongoCollection.html) .
Breafly, MongoCollection has not method "insert", but mongodb shell has. In this way, I must to **support all methods "insert" in the PR**.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/type/impl/MongoDBTypeContributor.java
```diff
@@ -20,5 +20,6 @@
 	@Override
 	public void contribute(TypeContributions typeContributions, ServiceRegistry serviceRegistry) {
 		typeContributions.contributeType( new StringAsObjectIdType() );
+		typeContributions.contributeType( new BytesAsBsonBinaryType() );
```
schernolyas: @DavideD . The change links with the PR. You can see to test BuiltinTypesTest. And ... You can remove my change and run the test.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -874,195 +905,200 @@ public DuplicateInsertPreventionStrategy getDuplicateInsertPreventionStrategy(En
...
 		Boolean upsert = (Boolean) queryDesc.getCriteria().get( "upsert" );
 		Boolean bypass = (Boolean) queryDesc.getCriteria().get( "bypassDocumentValidation" );
-		DBObject o = (DBObject) queryDesc.getCriteria().get( "writeConcern" );
+		Document o = (Document) queryDesc.getCriteria().get( "writeConcern" );
 		WriteConcern wc = getWriteConcern( o );
-		final DBObject theOne = collection.findAndModify( query, fields, sort, ( remove != null ? remove : false ),
-				update, (nevv != null ? nevv : false), (upsert != null ? upsert : false), (bypass != null ? bypass : false),
+		//@TODO Analyse it! New API have methods findAndRemove and findAndUpdate
+		/*final Document theOne = collection.findAndModify( query, fields, sort, ( remove != null ? remove : false ),
+				update, (returnNewDocument != null ? returnNewDocument : false), (upsert != null ? upsert : false), (bypass != null ? bypass : false),
```
schernolyas: @DavideD  A lot of thanks. I must to remove comments.  I will prepare new commit.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/performance/MongoDBPerformanceTest.java
```diff
@@ -29,37 +29,40 @@
...
+					action = "countInvocation(\"load\")",
+					name = "countFind1"),
+			@BMRule(targetClass = MONGO_COLLECTION,
+					isInterface = true,
+					targetMethod = "find()",
+					helper = HELPER,
 					action = "countInvocation(\"load\")",
-					name = "countFindOne")
+					name = "countFind2")
 	})
```
schernolyas: @DavideD  ... You see list of actual rules.  About name of the rule ....May be  You right.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
```diff
@@ -146,19 +145,74 @@ public boolean setCollation(String collation) {
...
-			return new MongoDBQueryDescriptor(
-				collection,
-				operation,
-				parse( criteria ),
-				parse( projection ),
-				parse( orderBy ),
-				parse( options ),
-				parse( updateOrInsert ),
-				null );
+		//@todo redactor the spagetti!
```
schernolyas: @DavideD  Okey ... I refactor the part of code as I think. Thanks!!!!

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/impl/MongoDBBasedQueryParserService.java
```diff
@@ -32,9 +31,16 @@
 
 	private volatile SessionFactoryEntityNamesResolver entityNamesResolver;
 
+	private ThreadLocal<QueryParser> parser = new ThreadLocal<QueryParser>() {
```
schernolyas: @DavideD  I changed "new QueryParser()" to "parser.get()". I think that "new QueryParser()" operation is expensive. I offer to optimize it. Are You agree?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/impl/MongoDBQueryDescriptor.java
```diff
@@ -142,45 +161,44 @@ public Operation getOperation() {
...
 
 	/**
 	 * Returns the update (new values to apply) in case this is an UPDATE query or values to insert in case this is an
 	 * INSERT query.
 	 */
-	public DBObject getUpdateOrInsert() {
-		return updateOrInsert;
-	}
+	public Document getUpdateOrInsertOne() { return updateOrInsertOne; }
+	public List<Document> getUpdateOrInsertMany() { return updateOrInsertMany; }
```
schernolyas: @DavideD  A lot of thanks!!!! I will restore format!

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
```diff
@@ -144,39 +145,51 @@ public void stop() {
...
 					throw log.databaseDoesNotExistException( config.getDatabaseName() );
 				}
 			}
-			DB db = mongo.getDB( databaseName );
+			MongoDatabase db = mongo.getDatabase( databaseName );
 			if ( containsDatabase == null ) {
 				// force a connection to make sure we do have read access
 				// otherwise the connection failure happens during the first flush
-				db.collectionExists( "WeDoNotCareWhatItIsWeNeedToConnect" );
+				MongoCursor<String> it = db.listCollectionNames().iterator();
```
schernolyas: @DavideD  Hm .... Yes we can. I will fix it

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/type/impl/BytesAsBinaryTypeDescriptor.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.hibernate.type.descriptor.WrapperOptions;
+import org.hibernate.type.descriptor.java.AbstractTypeDescriptor;
+
+import org.bson.types.Binary;
+
+/**
+ * @author Sergey Chernolyas &amp;sergey_chernolyas@gmail.com&amp;
+ */
+public class BytesAsBinaryTypeDescriptor extends AbstractTypeDescriptor<byte[]> {
```
schernolyas: @DavideD . Thanks. Fixed.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/impl/MongoDBQueryDescriptor.java
```diff
@@ -26,30 +26,47 @@
...
-		DISTINCT;
+		AGGREGATE_PIPELINE( false ),
+		DISTINCT ( false );
+		private boolean deprecated;
+
+		Operation(boolean deprecated) {
+			this.deprecated = deprecated;
+		}
+
+		public boolean isDeprecated() {
```
schernolyas: @DavideD May be You right. Fistly, I wants to mark operations that deprecaded in mongodb java 3.x driver. For marking, I upgrade the enum.  You right .... I remove the unused method.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
schernolyas: @DavideD  I can do in OGM-1265 other new operations. Lke 'deleteOne' and etc. (Operation 'remove' is deprecaded in java driver)

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
DavideD: Make sur eto refactor only if you need to. Otherwise, only remove the comment for now

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
DavideD: > Breafly, MongoCollection has not method "insert", but mongodb shell has. In this way, I must to support all methods "insert" in the PR

You can certainly do that in a separate PR like I asked you. In reality, you are adding a new feature to our parser that consist in understanding the inertOne and insertMany construct. It would deserve at least to be included in a separate commit. The reason I'm saying this is a new feature is because you didn't remove the old doInsert method. In the end you are not replacing the deprecated API.

Also, I'm not sure what's the criteria for adding insertOne and insertMany and not the new update and delete methods.

DavideD: I guess you can keep it like this for this time.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/performance/MongoDBPerformanceTest.java
DavideD: > @DavideD ... You see list of actual rules. About name of the rule ....May be You right.

What do you mean?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
schernolyas: @DavideD The code is hard for understanding :( I try to refactor. After it, I will remove the comment.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -811,19 +840,21 @@ public int executeBackendUpdateQuery(final BackendQuery<MongoDBQueryDescriptor>
...
 		}
 
 		switch ( queryDescriptor.getOperation() ) {
+			case INSERTMANY:
+			case INSERTONE:
 			case INSERT:
 				return doInsert( queryDescriptor, collection );
 			case REMOVE:
-				return doRemove( queryDescriptor, collection );
+				return (int) doRemove( queryDescriptor, collection );
```
DavideD: It would be better if you return an int. There's no reason for doRemove to return a Long

DavideD: Well, what I mean is that this method is only used here, returning an int make more sense

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -811,19 +840,21 @@ public int executeBackendUpdateQuery(final BackendQuery<MongoDBQueryDescriptor>
...
 			case UPDATE:
-				return doUpdate( queryDescriptor, collection );
+				return (int) doUpdate( queryDescriptor, collection );
```
DavideD: same as here as in doRemove

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1176,7 +1220,7 @@ private static DBObject associationKeyToObject(AssociationKey key, AssociationSt
 			MongoHelpers.setValue( columns, name.substring( prefix.length() ), columnValues[i++] );
 		}
 
-		BasicDBObject idObject = new BasicDBObject( 1 );
+		Document idObject = new Document(  );
```
DavideD: Double spaces in the constructor

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1159,13 +1203,13 @@ else if ( entityKeyMetadata != null ) {
...
 
-	private static DBObject associationKeyToObject(AssociationKey key, AssociationStorageStrategy storageStrategy) {
+	private static Document associationKeyToObject(AssociationKey key, AssociationStorageStrategy storageStrategy) {
 		if ( storageStrategy == AssociationStorageStrategy.IN_ENTITY ) {
 			throw new AssertionFailure( MongoHelpers.class.getName()
 					+ ".associationKeyToObject should not be called for associations embedded in entity documents" );
 		}
 		Object[] columnValues = key.getColumnValues();
-		DBObject columns = new BasicDBObject( columnValues.length );
+		Document columns = new Document(  );
```
DavideD: Double spaces in the constructor

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1302,10 +1347,11 @@ else if ( operation instanceof InsertOrUpdateAssociationOperation ) {
 				}
 				else {
 					MongoDBAssociationSnapshot associationSnapshot = (MongoDBAssociationSnapshot) association.getSnapshot();
-					DBCollection associationCollection = getAssociationCollection( associationKey, storageStrategy );
-					DBObject query = associationSnapshot.getQueryObject();
-					DBObject update = new BasicDBObject( "$set", new BasicDBObject( ROWS_FIELDNAME, toStore ) );
-					associationCollection.update( query, update, true, false, getWriteConcern( associationContext ) );
+					MongoCollection<Document> associationCollection = getAssociationCollection( associationKey,
+																								storageStrategy );
```
DavideD: Why this new line?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/type/impl/BytesAsBsonBinaryType.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+package org.hibernate.ogm.datastore.mongodb.type.impl;
+
+import org.hibernate.type.BinaryType;
+
+/**
+ * @author Sergey Chernolyas &amp;sergey_chernolyas@gmail.com&amp;
+ */
+public class BytesAsBsonBinaryType extends BinaryType {
+	public static final String TYPE_NAME = "bson_binary_bytes";
```
DavideD: Why did you create a new class for this? Use Binary type directly.

We already have many classes to map type and I would prefer to not add new ones unless needed

DavideD: Also you are only using this for the constant. In this case you can simply replace the constant and write the string directly (like we do everything else) and remove this class.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/type/impl/LongAsBinaryType.java
```diff
@@ -0,0 +1,18 @@
...
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.datastore.mongodb.type.impl;
+
+/**
+ * @author Sergey Chernolyas &amp;sergey_chernolyas@gmail.com&amp;
+ */
+public class LongAsBinaryType extends org.hibernate.type.LongType {
+	public static final String TYPE_NAME = "binary_long";
```
DavideD: See comment on BytesAsBsonBinaryType

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/type/impl/LongAsBinaryTypeDescriptor.java
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	@Override
+	public Long fromString( String string ) {
+		return Long.parseLong( string );
+	}
+
+	@Override
+	public <X> X unwrap(Long value, Class<X> type, WrapperOptions options) {
+		log.debug( " unwrap !" );
```
DavideD: This debug message is not very useful

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/type/impl/MongoDBTypeContributor.java
DavideD: I think you can avoid this change and in the dialect refers directly to BinaryType instead.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -753,13 +773,22 @@ else if ( type.getReturnedClass() == ObjectId.class ) {
 		else if ( type instanceof StringAsObjectIdType ) {
 			return StringAsObjectIdGridType.INSTANCE;
 		}
+		else if ( type instanceof BytesAsBsonBinaryType ) {
```
DavideD: I think you can use BinaryType here without the need of this additional class. Unless I'm missing something

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -753,13 +773,22 @@ else if ( type.getReturnedClass() == ObjectId.class ) {
...
+			return BytesAsBsonBinaryGridType.INSTANCE;
+		}
+		else if ( type.getReturnedClass() == Long.class && type.getClass() == SerializableToBlobType.class  ) {
+			return LongAsBinaryGridType.INSTANCE;
+		}
```
DavideD: I think what you have to do here is something similar at what I've done for Neo4j with Bolt, check the classes:
`ByteArrayMappedGridTypeDescriptor`
`SerializableAsByteArrayType`

I don't think you should check the returned class and only applied this for Long, I'm pretty sure a user could store other types as well using Lob, it needs to be a  more generic.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -753,13 +773,22 @@ else if ( type.getReturnedClass() == ObjectId.class ) {
...
+		else if ( type.getReturnedClass() == byte[].class && type.getClass() == MaterializedBlobType.class  ) {
+			return BytesAsBinaryGridType.INSTANCE;
+		}
```
DavideD: In this case I don't think you have to check the returned class, you can just use:
```
		else if ( type instanceof MaterializedBlobType ) {
			return BytesAsBsonBinaryGridType.INSTANCE;
		}
```

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
DavideD: I'm doing something similar in `BoltNeo4jTypeConverter`:

```
 	@Override
	public GridType convert(Type type) {
		if ( type instanceof SerializableToBlobType ) {
			SerializableToBlobType<?> exposedType = (SerializableToBlobType<?>) type;
			return new SerializableAsStringType<>( exposedType.getJavaTypeDescriptor() );
		}
		return conversionMap.get( type );
```

DavideD:
Ok, I gave a second look and we actually need the type converter (except for a couple of classes).

There are still some things to change though that I've commented it in line.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/impl/MongoDBBasedQueryParserService.java
DavideD: Let's avoid this optimization for now. We will think about it later when we have some performance test

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
schernolyas: @DavideD  I pushed my offer for refactor it.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
schernolyas: @DavideD . API changed. See [getDeletedCount](http://mongodb.github.io/mongo-java-driver/3.4/javadoc/com/mongodb/client/result/DeleteResult.html#getDeletedCount--)

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
schernolyas: @DavideD  API changed. See [UpdateResult](http://mongodb.github.io/mongo-java-driver/3.4/javadoc/com/mongodb/client/result/UpdateResult.html)

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
schernolyas: fixed

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
schernolyas: fixed

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
schernolyas: fixed

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/impl/MongoDBBasedQueryParserService.java
schernolyas: restored

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
schernolyas: restored and migrated to 3.x .  fixed.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/impl/MongoDBQueryDescriptor.java
schernolyas: Unused method removed.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
schernolyas: revert the commit. it broke DatastoreInitializationTest

schernolyas:
Hi @DavideD !
Please, review last changes.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
DavideD: I got it,just cast the value to the expected type inside the method.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
DavideD: I got it,just cast the value to the expected type inside the method. Not in the switch, this way you don't have to change the signature of the method

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
DavideD: Did you check this?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
schernolyas: Hi @DavideD . Okey. I move casting to inside the methods.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
schernolyas: fixed

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/type/impl/MongoDBTypeContributor.java
schernolyas: @DavideD I rechecked.
______________________
java.lang.ClassCastException: org.bson.types.Binary cannot be cast to [B
_____________________________________

DavideD:
@schernolyas you should send separate PR for separate issues, if you continue to add unrelated stuff to this PR I will never be able to integrate it. I also explain you before that we cannot include SNAPSHOT libraries in the project, it's OK if you want to experiment with it but we will only use an external  library when there is at least a Final  release.

schernolyas:
Hi @DavideD !

You right! I see that I something goes wrong. Git offers me commits from other branch. I will recreate branch for You can continue review.

PR #850, opened 2017-03-16
OGM-650 Contrib repository

DavideD:
https://hibernate.atlassian.net/browse/OGM-650

I've removed the part related to Ehcache, Cassandra, CouchDB and Redis

At the moment they are in a new project on my repository: https://github.com/DavideD/hibernate-ogm-contrib

I'll move them on the hibernate repository after the review.


bom/pom.xml
```diff
@@ -620,20 +591,6 @@
...
-            <!-- Redis -->
-            <dependency>
-                <groupId>biz.paluch.redis</groupId>
-                <artifactId>lettuce</artifactId>
-                <version>${lettuceVersion}</version>
-            </dependency>
-
             <!-- RESTEasy for accessing CouchDB -->
             <dependency>
                 <groupId>org.jboss.resteasy</groupId>
```
gsmet: If we use it only for CouchDB, we should remove it. If not, we should change the comment.

distribution/src/main/assembly/dist.xml
```diff
@@ -151,52 +136,7 @@
...
-            <useProjectArtifact>false</useProjectArtifact>
-            <outputDirectory>dist/lib/redis</outputDirectory>
-            <useTransitiveFiltering>true</useTransitiveFiltering>
-            <unpack>false</unpack>
-            <includes>
-                <include>org.hibernate.ogm:hibernate-ogm-redis</include>
-            </includes>
-        </dependencySet>
-    </dependencySets>
+   </dependencySets>
```
gsmet: Looks like there is a space missing here.

modules/src/main/assembly/dist.xml
```diff
@@ -334,5 +225,5 @@
                 <include>org.ow2.asm:asm-util</include>
             </includes>
        </dependencySet>
-    </dependencySets>
+   </dependencySets>
```
gsmet: Same here, looks like there is a space missing now.

core/src/test/java/org/hibernate/ogm/backendtck/associations/collection/manytomany/ManyToManyExtraTest.java
```diff
@@ -21,7 +21,7 @@
  * @author Emmanuel Bernard &lt;emmanuel@hibernate.org&gt;
  */
 @SkipByGridDialect(
-		value = { GridDialectType.CASSANDRA, GridDialectType.INFINISPAN_REMOTE },
+		value = { GridDialectType.INFINISPAN_REMOTE },
```
gsmet: ~~This bugs me a little. We are losing information and I would have expected the TCK to be used for the contrib dialects. It looks like you're reintroducing the ability to do that in a further commit. What's your strategy exactly for that?~~

We agree that you have moved this information in the skipTests file of each contrib module?

bom/pom.xml
DavideD: It's used for Neo4j as well

DavideD: I missed the comment, I'll update it

distribution/src/main/assembly/dist.xml
DavideD: ok

modules/src/main/assembly/dist.xml
DavideD: ok

core/src/test/java/org/hibernate/ogm/backendtck/associations/collection/manytomany/ManyToManyExtraTest.java
DavideD: Yes, I moved the tests in the skipTests files

DavideD:
> AFAICS you lost all the history of these modules. I'm wondering if you couldn't keep it but applying the changes on top of a clone of the original repository. I know the history is often very useful and it might be even more if we expect new people to step in to maintain this repository. I don't know how many changes you applied on top of the existing dialects and if they are easily appliable on top of the existing repo. I wouldn't do it in the details (I mean no history for the modules, the integrationtests, the doc... is OK but I think it's important for the various dialects)

OK, It makes sense. I see what I can do.

> about the test strategy, I dug a bit further and it looks like you are running the TCK tests and add the excludes in a skipTests file. Am I right? It looks interesting. I'm a bit worried about just using the simple name and not the FQCN though.

I'm open to other solutions but I think I prefer to have a way for each dialect to skip it's own tests.
I guess I can use the whole name. It's probably safier

> as for Travis, I think we should try to make it a Matrix build and not use the parent to build all the  modules. This way, I think Travis makes the info easier to find when the build fails and it will build all modules. This means that a person working on a specific module would have the results even if another module does not build.

it should fail only if the person changed something affecting the other modules. It's a good idea anyway.

DavideD:
I'll let you know when I applied the changes

gsmet:
> it should fail only if the person changed something affecting the other modules. It's a good idea anyway.

Yeah, my guess is that when we will make an infrastructure change to the core, the contrib dialects won't be adapted right away so some may fail for some time. Having Travis for the one you work on even in this case looks like a good idea to me.

DavideD:
@gsmet I've applied the remarks and updated the other repository as well, History included

gsmet:
@DavideD 

Looks good. Logging things we should not forget about: update the README.md, clean up the .travis.yml file (there are Redis and Cassandra specific stuff in it).

I'm wondering if we should consider changing the groupId in the contrib repo to org.hibernate.ogm.contrib (not sure it's recommended to use a subgroup or if we have to use something like org.hibernate.ogm-contrib). WDYT? It's probably something we should discuss with the others.

I think it's probably worth a general email on the list to check that we're all on the same page before merging this.

Nice work.

DavideD:
I've added a commit to move some properties in the BOM so that they are consistent in the new repositories. The repository for CouchDB, Cassandra, EhCache and Redis are currently on my account but I'll create them in the hibernate one if they are OK:

https://github.com/DavideD/hibernate-ogm-redis
https://github.com/DavideD/hibernate-ogm-couchdb
https://github.com/DavideD/hibernate-ogm-ehcache
https://github.com/DavideD/hibernate-ogm-cassandra

They will need some additional work in regard to the documentation and possibly something else but I think they are a good starting point.

gsmet:
Merged and added a couple of commits on top of that.

PR #852, opened 2017-03-28
OGM-786 migration to MongoDB Java Driver 3.x API

schernolyas:
Hi @DavideD !
I recreated branch. You can continue review

PR #856, opened 2017-04-10
OGM-1279 Migration to mongo 3 api

DavideD:
https://hibernate.atlassian.net/browse/OGM-1279

I've created a sub-task on JIRA for this PR as it is much bigger than I expected initially.

@gsmet I decided to apply some changes to  @schernolyas [ initial pull request](https://github.com/hibernate/hibernate-ogm/pull/852), if you could have a look to the additional commits that would be appreciated.

@schernolyas  I've applied the following changes;
* Revert check for the existence of a database in MongoDB because it does not work with Fongo
* Remove additional classes fo rthe conversion of type in MongoDB. In the end we really need a new class to store the binary as BSON binary
* Remove the commit withthe refactoring. Honestly. this PR is already big enough and I prefer not to add additional stuff. I'll have a second look at it later
* I added some code to make sure that the user could still run native `find` queries with modifiers. These are deprecated in MongoDB but our parser is not capable to understand the new syntax I think.

Thanks


DavideD:
@gsmet I've rebased this PR on top of https://github.com/hibernate/hibernate-ogm/issues/857

As far as I understand  the values are mapped in the same way on the db, what's changed is the way they are stored in the driver once they are retrieved. This should not have an impact on the user. I've decided [to remove the randomness in BuiltInTypeTest](https://github.com/hibernate/hibernate-ogm/pull/856/commits/662154ed9e615254810ccc630e7bdf78e856700a) to compare the data in the db using before/after the patch and it seems to contain the same results.

Thanks

gsmet:
Merged, thanks!

PR #857, opened 2017-04-27
OGM-1281 Test native mapping of binary values in MongoDB

DavideD:
https://hibernate.atlassian.net/browse/OGM-1281

@gsmet This PR add some check to make sure that the mapping of the binary value is not changed when we are going to apply https://github.com/hibernate/hibernate-ogm/pull/856


gsmet:
Merged, thanks!

PR #858, opened 2017-04-28
OGM-1282 Remove FongoDB

DavideD:
https://hibernate.atlassian.net/browse/OGM-1282

I would remove it now that we are moving dialects in different places.
We can always create a separate repository for it if the need arise.


gsmet:
@DavideD merged and added 2 commits, one to remove it from the Travis build and one to simplify some logic:
https://github.com/hibernate/hibernate-ogm/commit/9294b96b1ce0fdbe8bbb96a997308c864ccf9e76

Any good reason for the rather complicated logic that existed before? IIRC it came from the original Sergey's patch.

PR #859, opened 2017-05-03
OGM-1284 wild fly configuration

DavideD:
https://hibernate.atlassian.net/browse/OGM-1284

This should improve the paragraph related ot the configuration of WildFly.

@gsmet Can you have a look and let me know if it makes sense to you?

Thanks

### DIFF
--- a/documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
+++ b/documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
@@ -358,43 +358,67 @@ as you can include your favourite version of Hibernate OGM directly to the colle
 
 ==== Packaging Hibernate OGM applications for WildFly 10
 
+When using WildFly several of the technologies it includes are automatically enabled.
+For example Hibernate ORM is made available to your applications if your `persistence.xml`
+defines a persistence unit using Hibernate as persistence provider
+(or is not specifying any provider, as Hibernate is the default one).
+
+Similarly, Hibernate Search is automatically activated and made available on the user's application
+classpath if and when the application server detects the need for it.
+This is the default behaviour, but you are in control and can override this all;
+see the https://docs.jboss.org/author/display/WFLY10/JPA+Reference+Guide[WildFly JPA Reference Guide]
+for a full list of properties you can explicitly set.
+
+WildFly 10 however does not include Hibernate OGM and the compatible Hibernate ORM and Hibernate Search
+versions and it will require some configuration to make everything works.
+
+Set up WildFly after downloading the modules::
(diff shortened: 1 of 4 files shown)


documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
```diff
@@ -358,43 +358,67 @@ as you can include your favourite version of Hibernate OGM directly to the colle
...
+(and not the one shipped with WildFLy).
 
-.Property to set a different Hibernate ORM version
+.Property for enabling a Hibernate Search and Hibernate ORM version compatible with Hibernate OGM
 ====
 [source, XML]
 [subs="verbatim,attributes"]
 ----
-<property name="jboss.as.jpa.providerModule" value="org.hibernate:{hibernate-orm-major-minor-version}"/>
+<property name="wildfly.jpa.hibernate.search.module" value="org.hibernate.search.module:{hibernate-search-module-slot}"/>
```
roguexz: Should the value read `org.hibernate.search.orm` or `org.hibernate.search.module`. The documentation above says one while the example uses another. Which one is right?

roguexz: Strangely, I had to change the value to `org.hibernate.search.engine` and that seems to have loaded the application. I understand modules, but probably not as much as I would have liked w.r.t. WildFly.

DavideD: > Should the value read org.hibernate.search.orm or org.hibernate.search.module. The documentation above says one while the example uses another. Which one is right?

Damn, writing documentation is hard :/

The complete value should be `org.hibernate.search.orm:5.6.1.Final-orm51` and it's the the complete identifier of the module containing Hibernate Search ORM. You can find an explanation of the property on the WildFLy documentation: https://docs.jboss.org/author/display/WFLY10/JPA+Reference+Guide#JPAReferenceGuide-UsingHibernateSearch

Note that the -orm51 is important because it make sure that you are going to use a version compatible with ORM 5.1

Some details are also included in the Hibernate Search doc: https://docs.jboss.org/hibernate/search/5.6/reference/en-US/html_single/#_update_and_activate_latest_hibernate_search_version_in_wildfly

Sorry for the inconvinience. 

Please, let me know if you have other issues.

DavideD: I've fix the error in the value of the property

gsmet:
Merged with some additional fixes. I agree it's much better this way. Thanks!

PR #861, opened 2017-05-31
OGM-1288 Constrain the edge direction when querying associations

hholierhoek:
Modifying BaseNeo4jEntityQueries.findAssociationPartialQuery to include
the edge direction between parent and owner.

### DIFF
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BoltNeo4jDialect.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BoltNeo4jDialect.java
@@ -421,7 +421,7 @@ public class BoltNeo4jDialect extends BaseNeo4jDialect<BoltNeo4jEntityQueries, B
 
 		Transaction tx = transaction( associationContext );
 		ClosableIterator<RemoteNeo4jAssociationPropertiesRow> relationships = getEntityQueries( entityKey.getMetadata(), associationContext )
-				.findAssociation( tx, entityKey.getColumnValues(), relationshipType );
+				.findAssociation( tx, entityKey.getColumnValues(), relationshipType, associationKey.getMetadata() );
 		while ( relationships.hasNext() ) {
 			RemoteNeo4jAssociationPropertiesRow row = relationships.next();
 			AssociatedEntityKeyMetadata associatedEntityKeyMetadata = associationContext.getAssociationTypeContext().getAssociatedEntityKeyMetadata();
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/EmbeddedNeo4jDialect.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/EmbeddedNeo4jDialect.java
@@ -302,7 +302,7 @@ public class EmbeddedNeo4jDialect extends BaseNeo4jDialect<EmbeddedNeo4jEntityQu
 	private Map<RowKey, Tuple> createAssociationMap(AssociationKey associationKey, AssociationContext associationContext, EntityKey entityKey) {
 		String relationshipType = associationContext.getAssociationTypeContext().getRoleOnMainSide();
 		ResourceIterator<Relationship> relationships = getEntityQueries( entityKey.getMetadata(), associationContext )
-				.findAssociation( dataBase, entityKey.getColumnValues(), relationshipType );
+				.findAssociation( dataBase, entityKey.getColumnValues(), relationshipType, associationKey.getMetadata() );
 
 		Map<RowKey, Tuple> tuples = new HashMap<RowKey, Tuple>();
 		try {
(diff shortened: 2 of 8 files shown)


DavideD:
ok to test

DavideD:
The patch looks good, I would like to add an additional test case though.

May I ask you how you tried to get the data from the datastore when you had the problem?

hholierhoek:
Not sure this is what you're looking for, but the fetch was based on querying a secondary key, effectively in an @ElementCollection List<String>, which of course were stored in associated nodes. (In reality the list contained an object which is converted to string via @Converter(autoApply = true), but I suspect that doesn't matter.)

The parent/child relationship details are in the forum posting.

And we noticed the problem when a delete was done in the middle of a tree, and it removed the entire tree from its root.

If I missed your meaning let me know though.

hholierhoek:
Still waiting for CLA approval it seems. I'll see if I can figure out where the email might have gone.

hholierhoek:
Any chance you can find out from the inside what email address the CLA approval went to? I can't find the answer here. The FOSS and legal folks don't know. I've tried resending too.

DavideD:
@hholierhoek Don't worry, I'll check and let you know

DavideD:
@hholierhoek What name di you use to sign the CLA?

hholierhoek:
Hubert Holierhoek
My login id on JBoss Developer is huberth.

DavideD:
Did you select the Hibernate project when you signed the CLA? It seems we cannot find your user name in the queue. Could you eventually try to sign it again please? Just in case

hholierhoek:
Most definitely it was Hibernate. See attachment.

cla

hholierhoek:
Trying to sign it again gives the following error:

You already are a contributor, or have applied to be a contributor to Hibernate.

Sanne:
hi @hholierhoek I'm sorry for the delay. I've raised a support ticket with the team running that service.

hholierhoek:
Any update on the support ticket?

hholierhoek:
I just noticed a reference to OGM-1290 in this thread, which I created and was working on a fix for... does it mess with this pull request if I've pushed that fix to my fork already?

emmanuelbernard:
Hello @hholierhoek sorry for the long wait on this one. We had to track down where in the process your agreement got stuck. The system was expecting a click on the verification email you should have received. If you have not received it, can you click on the `Resend verification email` and check any spam filter you might have. Note if that's not your email address that was used, the appropriate person would have to check its mailbox.

Let me know how that goes.

emmanuelbernard:
@hholierhoek about the reference to ` OGM-1290`, yes that's not the workflow we prefer.
What you should do is create a dedicated git branch (we call them topic branch) per issue you are working on (unless they are very closely related). That way you can push these branches to your GitHub repo and create dedicated pull requests for each.

For this time, we can deal with it. It would require Git kung fu on your side to fix it.

hholierhoek:
I didn't receive a verification email. I just confirmed my email address is correct in my JBoss Developer account. And I've clicked the resend button close to half a dozen times. Nothing in my junk folder either.

But I presumed that email was going to someone with authority here. Is that not the case?

emmanuelbernard:
OK let's make it a bit different. @hholierhoek, can you reply to this pull request by confirming that Nokia and you agreed to the Hibernate CLA? If you agree here, we will fix the technical hurdle later.

hholierhoek:
Re: OGM-1290. Okay, thanks. Note that I do have another commit to come for that one before I can create a pull request.  Will I need to move this latest commit to another branch first?

emmanuelbernard:
@hholierhoek if you know how to do that ( move this latest commit to another branch first), then go ahead and move all OGM-1290 in a dedicated branch. We will ignore the one that shows up here.

hholierhoek:
Sure. I can confirm that I, and my manager (and his manager), at Nokia agree to the Hibernate CLA.

hholierhoek:
I'll see what git 'kung fu' I can find to do that.

hholierhoek:
Hmm... not sure I know what git-fu to do here. If I create a branch off my current master I don't know whether a pull request would pick up the first 1290 commit. If I branch it before 1288 I think we'll end up with merge conflicts - there are a reasonable number of overlapping lines changed.

Or maybe it's easier for you, too, if I just wait until this one is pulled, then commit to master and create a new pull request from my master? (And promise never to complicate things like this again!)

DavideD:
@hholierhoek sorry for the late response, I'm not sure I understand the problem related to OGM-1290 if what you have works, send a PR so that I can have a look at it. Thanks a lot.

hholierhoek:
Okay, there will be one more commit to my master (I've not pushed the latest one up), so the new pull request will contain a total of 3 commits - one for 1288 and two for 1290.

hholierhoek:
I've done the last commit. It seems I can't create a new PR against the same fork/branch.

DavideD:
Yes, there's no need to send a new pr I will pull your commit. By the way your previous  PR looks good but by working on a testcase before merging I've found some problems with some dialects. I'm working on it.

DavideD:
I'm going to close this one and send a new PR for OGM-1290 that will include the commit you submitted.

The reason is that I need to add a test (I think) and fix checkstyle issues.

Check https://github.com/hibernate/hibernate-ogm/issues/863 for OGM-1288

PR #864, opened 2017-07-24
OGM-1290 'target' node in Neo4j association queries needs to be qualified with a label

DavideD:
https://hibernate.atlassian.net/browse/OGM-1290

### DIFF
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
@@ -1333,12 +1333,16 @@ public class MongoDBDialect extends BaseGridDialect implements QueryableGridDial
 				if ( storageStrategy == AssociationStorageStrategy.IN_ENTITY ) {
 					writeConcern = mergeWriteConcern( writeConcern, getWriteConcern( associationContext ) );
 					if ( insertStatement != null ) {
+						// The association is updated in a new document
 						MongoHelpers.setValue( insertStatement, collectionRole, rows );
 					}
 					else {
-						MongoHelpers.setValue( ( (MongoDBTupleSnapshot) associationContext.getEntityTuplePointer().getTuple().getSnapshot() ).getDbObject(),
-								associationKey.getMetadata().getCollectionRole(), toStore );
-						addSetToQuery( updateStatement, collectionRole, toStore );
+						// The association is updated on an existing document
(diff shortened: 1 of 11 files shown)


DavideD:
Jenkins, retest this please

DavideD:
@gsmet Jenkins seems to be happy about this PR

DavideD:
@gsmet Did you check this PR?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/dialect/impl/BaseNeo4jEntityQueries.java
```diff
@@ -588,11 +588,11 @@ public String getUpdateEmbeddedColumnQuery(Object[] keyValues, String embeddedCo
 		return query;
 	}
 
-	public String getFindAssociationQuery(String role, AssociationKeyMetadata associationKeyMetadata) {
-		String query = findAssociationQueryCache.get( role );
+	public String getFindAssociationQuery(String relationshipType, AssociationKeyMetadata associationKeyMetadata) {
```
gsmet: @DavideD I'm going to merge the PR but it would be nice if you could add a comment to the method explaining the difference between the 2 and why we have to use one in one case and the other in the other case.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1333,12 +1333,16 @@ else if ( operation instanceof InsertOrUpdateAssociationOperation ) {
...
+						// The association is updated in a new document
 						MongoHelpers.setValue( insertStatement, collectionRole, rows );
 					}
 					else {
-						MongoHelpers.setValue( ( (MongoDBTupleSnapshot) associationContext.getEntityTuplePointer().getTuple().getSnapshot() ).getDbObject(),
-								associationKey.getMetadata().getCollectionRole(), toStore );
-						addSetToQuery( updateStatement, collectionRole, toStore );
+						// The association is updated on an existing document
+						Document document = getDocument( association, associationContext );
+						if ( document != null ) {
```
gsmet: Do you know when this situation happens? Don't we lose some data in the case the document is null?

gsmet:
@DavideD I merged the PR, thanks. Still interested in you adding some comments to the `getFindAssociationQuery()` method you modified in `BaseNeo4jEntityQueries`.

PR #866, opened 2017-08-02
OGM-1294 Properties in subclasses

DavideD:
https://hibernate.atlassian.net/browse/OGM-1294

All the tests I've added are working, I think we should add them to master but I couldn't replicate the issue.

@gsmet Maybe you can find what's missing

Cheers

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/persister/impl/EntityAssociationUpdater.java
+++ b/core/src/main/java/org/hibernate/ogm/persister/impl/EntityAssociationUpdater.java
@@ -151,7 +151,6 @@ class EntityAssociationUpdater {
 	}
 
 	private void addNavigationalInformationForInverseSide(int propertyIndex, AssociationKeyMetadata associationKeyMetadata, Object[] newColumnValue) {
-		AssociationPersister associationPersister = createAssociationPersister( propertyIndex, associationKeyMetadata, newColumnValue );
 
 		RowKey rowKey = getInverseRowKey( associationKeyMetadata, newColumnValue );
 
@@ -159,6 +158,8 @@ class EntityAssociationUpdater {
 		for ( String column : rowKey.getColumnNames() ) {
 			associationRow.put( column, rowKey.getColumnValue( column ) );
 		}
+
+		AssociationPersister associationPersister = createInverseAssociationPersister( propertyIndex, associationKeyMetadata, newColumnValue );
 		associationPersister.getAssociation().put( rowKey, associationRow );
 
 		if ( associationPersister.getAssociationContext().getEntityTuplePointer().getTuple() == null ) {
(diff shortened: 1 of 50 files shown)


gsmet:
@DavideD I think what you missed is that the issue is not with the main class but with a relation.

The OP test is getting a `Depositor`:
```java
class Depositor {
    [...]

    @OneToMany(mappedBy = "depositor", cascade = CascadeType.ALL, orphanRemoval = true, fetch = FetchType.EAGER)
    private Set<Account> accounts = new HashSet<>();

    [...]
}
```
`Account` is abstract and can be either a `GiroAccount` or an `InstantAccessAccount`.

The `Depositor` is called `entity` below and it does:
```java
final Set<Account> accounts = entity.getAccounts();
final Account account = accounts.iterator().next();
assertThat(account, instanceOf(GiroAccount.class));
final GiroAccount giroAccount = (GiroAccount) account;
assertThat(giroAccount.getBalance(), equalTo(100000.0));

// TODO: enable as soon as https://hibernate.atlassian.net/browse/OGM-1294 is fixed
assertThat(giroAccount.getCreditLimit(), equalTo(100000.0));
```

So the issue seems to be with accessing the specific fields in a polymorphic relation.

DavideD:
Opss...thanks, I'll have fix it tomorrow

gsmet:
@dadrus Thanks for taking the time to do it!

DavideD:
Jenkins, retest this please

DavideD:
I've updated the patch and it seems to work now.

DavideD:
Jenkins, retest this please

DavideD:
Jenkins, retest this please

DavideD:
Jenkins, retest this please

core/src/main/java/org/hibernate/ogm/persister/impl/EntityAssociationUpdater.java
```diff
@@ -191,11 +191,13 @@ private AssociationPersister createAssociationPersister(int propertyIndex, Assoc
...
 
 		Class<?> entityType = persister.getPropertyTypes()[propertyIndex].getReturnedClass();
+		String entityName = persister.getFactory().getClassMetadata( entityType ).getEntityName();
+		OgmEntityPersister entityPersister = (OgmEntityPersister) persister.getFactory().getEntityPersister( entityName );
 
 		AssociationTypeContext associationTypeContext = new AssociationTypeContextImpl(
 				serviceContext.getPropertyOptions( entityType, associationKeyMetadata.getCollectionRole() ),
 				serviceContext.getEntityOptions( entityType ),
-				persister.getTupleTypeContext(),
+				entityPersister.getTupleTypeContext(),
```
gsmet: So as far as I understand it, you pass the persister of the association entity, not the one of the owner? We probably should make it clear that the `persister` field is about the owner of the association.

The field in `AssociationTypeContext` is called `ownerEntityTupleTypeContext` so it should probably be changed. What also bugs me is that `AssociationTypeContextImpl` is instantiated in quite a lot of places and, currently, it's sometimes instantiated with the context of the owner and sometimes with the context of the current association entity.

I'm wondering if we should maybe instantiate it each time with both the current entity context and the owner context but I'm not sure it will not break anything.

But I'm not sure which one should be returned from `AssociationContextImpl#getTupleTypeContext()`...

DavideD:
Jenkins, retest this please

DavideD:
@gsmet I've added a couple of commit that should make it clearer what's going on.

I think initially we develop the method to keep track of the owner of the association but because in some situation we need the opposite side we use it  abit differently. This last two commits add some javadoc, rename and refactor some methods to (hopefully) clarify this

gsmet:
@DavideD thanks, it's much better now. I fixed a few typos in the comments and merged everything to master.

Nice work!

PR #873, opened 2017-08-31
OGM-786. Investigate usage of MongoDB's GridFS for storing large binary data

schernolyas:


schernolyas:
wait open of 5.4 release

PR #874, opened 2017-09-04
Changes that needs for Ignite and OrientDB dialects

schernolyas:
the commit contains changes that needs for Ignite and OrientDB dialects. (queryHints needs for Ignite, prepareDatabase needs for OrientDB)


DavideD:
Thanks @schernolyas, the patch is nice but I will send a separate PR with some changes

PR #877, opened 2017-09-08
OGM-983. make the test skipable

schernolyas:
Now the test is unskipable.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -10,7 +10,7 @@
 
     <groupId>org.hibernate.ogm</groupId>
     <artifactId>hibernate-ogm-bom</artifactId>
-    <version>5.2.0-SNAPSHOT</version>
+    <version>5.2.0.Alpha1</version>
 
     <packaging>pom</packaging>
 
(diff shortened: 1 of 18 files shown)


Sanne:
Nice suggestion, thanks!

Could you just change the commit message to be consistent with the project conventions? And while at it fix the english typo.

"OGM-983. make the test skipable"

should be

"OGM-983 Make it possible to skip test JPAResourceLocalTest"

- we don't use punctuation or other symbols to separate the JIRA code from the comment, just a single whitespace
- The comment shall start with a capital letter
- The comment should be correct English
- The comment should be descriptive enough to diferentiate this change from similar ones

So for example "update version of Hibernate ORM" is not good as we'll have many of those. "Update version of Hibernate ORM to v. 5.2.1.Final" is much better.
"Fixing a test", "Making a test skippable", etc.. are not useful comments.

schernolyas:
Hi @Sanne 

Thanks. I make correction that you said. I send new PR. Please see https://github.com/hibernate/hibernate-ogm/pull/878

PR #880, opened 2017-09-19
OGM-1318 Pass the tenant id to PurgeAllLuceneWork

gsmet:
I don't think it changes anything for now but as the rest of the code takes the tenant id into account, we should be consistent.

https://hibernate.atlassian.net/browse/OGM-1318


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/880

DavideD:
Thanks

PR #881, opened 2017-09-20
OGM-1269 Infinispan 9.1

DavideD:
https://hibernate.atlassian.net/browse/OGM-1269

A collective effort among me, @Sanne and @fax4ever

I've rebased the branch and change a bit the commit description by adding the JIRA issue where was missing.

Thanks

### DIFF
--- a/infinispan/src/main/java/org/hibernate/ogm/datastore/infinispan/logging/impl/Log.java
+++ b/infinispan/src/main/java/org/hibernate/ogm/datastore/infinispan/logging/impl/Log.java
@@ -8,9 +8,10 @@ package org.hibernate.ogm.datastore.infinispan.logging.impl;
 
 import org.hibernate.HibernateException;
 import org.hibernate.service.spi.ServiceException;
-
 import org.infinispan.commons.marshall.AdvancedExternalizer;
+import org.jboss.logging.Logger.Level;
 import org.jboss.logging.annotations.Cause;
+import org.jboss.logging.annotations.LogMessage;
 import org.jboss.logging.annotations.Message;
 import org.jboss.logging.annotations.MessageLogger;
 
@@ -43,5 +44,10 @@ public interface Log extends org.hibernate.ogm.util.impl.Log {
 	HibernateException externalizerIdNotMatchingType(Integer externalizerId, AdvancedExternalizer<?> registeredExternalizer, AdvancedExternalizer expectedExternalizer);
 
 	@Message(id = 1106, value = "Missing configuration for cache '%1$s' and a default configuration was not provided")
-	Exception missingCacheConfiguration(String cacheName);
+	HibernateException missingCacheConfiguration(String cacheName);
+
+	@LogMessage(level = Level.WARN)
+	@Message(id = 1107, value = "The Infinispan dialect uses fine grained maps and clustering.hash.groups must be enabled. clustering.hash.groups is now enable for cache `%1$s`."
+			+ "Enable it in your configuration to make this message disappear.")
+	void clusteringHashGroupsMustBeEnabled(String cacheName);
 }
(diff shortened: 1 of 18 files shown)


Sanne:
Awesome! thanks @DavideD @fax4ever ! 
Need to leave soon, will have a quick check now and review it more properly tonight.

Sanne:
this PR is obsolete :D

http://blog.infinispan.org/2017/09/infinispan-911final-is-out.html

infinispan/src/main/java/org/hibernate/ogm/datastore/infinispan/logging/impl/Log.java
```diff
@@ -41,4 +42,12 @@
 	@Message(id = 1105, value = "Infinispan Externalizer mistmatch: id [%1$d] was registered but taken " +
 			"by implementation '%2$s'. Expected externalizer: '%3$s' ")
 	HibernateException externalizerIdNotMatchingType(Integer externalizerId, AdvancedExternalizer<?> registeredExternalizer, AdvancedExternalizer expectedExternalizer);
+
+	@Message(id = 1106, value = "Missing configuration for cache '%1$s' and a default configuration was not provided")
+	HibernateException missingCacheConfiguration(String cacheName);
+
+	@LogMessage(level = Level.WARN)
+	@Message(id = 1107, value = "The Infinispan dialect uses fine grained maps and clustering.hash.groups must be enabled. clustering.hash.groups is now enable for cache `%1$s`."
```
gsmet: enable**d**

infinispan/src/test/resources/infinispan-local.xml
```diff
@@ -23,33 +22,29 @@
         <!-- *************************** -->
         <!--   Default cache settings    -->
         <!-- *************************** -->
-        <local-cache name="DEFAULT">
-            <transaction mode="NON_XA" />
+        <local-cache-configuration name="default-template">
+            <locking isolation="READ_COMMITTED" />
+            <transaction mode="NON_DURABLE_XA" transaction-manager-lookup="org.infinispan.transaction.lookup.JBossStandaloneJTAManagerLookup"/>
```
Sanne: Why did we change the transaction mode? I'd expect `NON_XA` to still be a reasonable default?

Sanne: Also wondering if `transaction-manager-lookup` is necessary?

infinispan/src/test/resources/log4j.properties
```diff
@@ -14,6 +14,6 @@ log4j.appender.stdout.layout.ConversionPattern=%d{ABSOLUTE} %5p %c{1}:%L - %m%n
 log4j.rootLogger=trace, stdout
 
 log4j.logger.org.hibernate=info
-log4j.logger.org.infinispan=info
+log4j.logger.org.infinispan=trace
```
Sanne: could we undo this? Infinispan at trace writes a hell of a lot of stuff :)

modules/pom.xml
```diff
@@ -50,6 +50,12 @@
             <optional>true</optional>
             <exclusions>
                 <exclusion>
+                <!-- Infinispan might use a different version than Neo4j. It is 
+                    safe to exclude it as we are going to use a module already present on WildFly -->
```
Sanne: Are you sure you mean "already present on WildFly" ?
It is true that WildFly contains a copy of Infinispan (together with its dependencies) but we should be using the updated module from the `infinispan-wildfly-modules` artifacts. Let's make sure we're binding to the right version and clarify the comment.

infinispan/src/test/resources/infinispan-local.xml
Sanne: I tried it and neither seems to be necessary according to our tests. I'm not sure though :)

Sanne:
@fax4ever the commit comments for  7b59675 looks a bit unconventional. We try to summarize the reason for the change in one single line; when it's not enough and you feel you need to explain more we still summarize it as one line, then followed by a blank line and then you can write a longer text. (The blank line is important as various tools will make it all look nice).

See also https://chris.beams.io/posts/git-commit/

modules/pom.xml
DavideD: When we launch the tests we donwload and unzip the infinispan-wildfly-modules first so, technically, they are already in the WIldFly we are using :)

Sanne: I'm replacing it with:

                

ok?

DavideD: Sounds good.

For consistency, the same comment appear underneath this one in the class but instead of caffeine refers to lucene, could you update that as well?

Sanne:
I've applied the improvements mentioned here but now realized we didn't updated documentation and example configuration files.

Closing this, I'll fix docs and send a new PR soon.

PR #882, opened 2017-09-21
OGM-1323 base class for grid dialect test helper

DavideD:
https://hibernate.atlassian.net/browse/OGM-1323

Not critical, but it makes our life easier if we need to add some methods.


gsmet:
Merged, thanks!

PR #884, opened 2017-09-26
OGM-1170 Create remote caches if they don't exist

DavideD:
https://hibernate.atlassian.net/browse/OGM-1170

This patch allows the user to enable the cache creation via the property: `hibernate.ogm.create_database`

As default cache it will use the one defined on the server.
At the moment is not possible to define a template to use different than the one on the server but I think it's already pretty good for a release.

It's working but I need to update the documentation.

Thanks


DavideD:
Added a small paragraph related to the new configuration option

Sanne:
I'm not sure I like this `create_database` property. It doesn't fit the Infinispan case very well.

But that's a topic for the mailing lists and/or meeting, it's good to have some flag now.

Sanne:
I noticed this causes quite some ERRORs to be logged by Infinispan, as it's not liking our attempts to see if the cache already exists by doing getCache(name)!=null .

Opened https://issues.jboss.org/browse/ISPN-8347

documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
```diff
@@ -210,6 +210,11 @@ The following two strategies exist (values of the `org.hibernate.ogm.datastore.k
 
 +
 Defaults to `CACHE_PER_TABLE`. It is the recommended strategy as it makes it easier to target a specific cache for a given entity.
+`hibernate.ogm.datastore.create_database`::
+Accept `true` or `false`. If `true`, Hibernate OGM will create a new cache on an Infinispan Server
+when the required one is not defined, this works only if a default configuration is defined on
+the server. If set to `false` an exception is thrown when the cache is not defined.
+Defaults to `false`.
```
Sanne: May I suggest:

`hibernate.ogm.datastore.create_database`::
If set to `true` Hibernate OGM will create any missing Cache definitions on the Infinispan Server.
This requires the Infinispan Server configuration to have a default configuration defined, as this will be copied to the newly defined caches.
If set to `false` an exception is thrown when a Cache is expected but was not explicitly configured on the server.
Defaults to `false`.

Sanne:
Looks good and I'd merge it!
I suggested a change in the doc, feel free to take it and merge, or ignore me and merge if you don't like it.

Sanne:
merged! Thanks

PR #889, opened 2017-10-17
[PREVIEW] OGM-359 Support stored procedures

schernolyas:
Hi!
The PR is for review. I need comments and offers.  

Please look to interface org.hibernate.ogm.dialect.storedprocedure.spi.StoredProcedureAwareGridDialect

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/jpa/impl/OgmJpaStoredProcedureQuery.java
+++ b/core/src/main/java/org/hibernate/ogm/jpa/impl/OgmJpaStoredProcedureQuery.java
@@ -1,0 +1,171 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.jpa.impl;
+
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+
+import javax.persistence.EntityManager;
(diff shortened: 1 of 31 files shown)


core/src/main/java/org/hibernate/ogm/jpa/impl/OgmJpaStoredProcedureQuery.java
```diff
@@ -132,7 +122,8 @@ public int getUpdateCount() {
 		return super.getUpdateCount();
 	}
 
-	/* (non-Javadoc)
+	/*
+	 * (non-Javadoc)
 	 * @see org.hibernate.jpa.internal.StoredProcedureQueryImpl#getResultList()
 	 */
```
DavideD: Can you simply remove these comments? As far as I know there is no benefit in keeping this type of comments

DavideD: Same for the other places where it appears

DavideD:
Thanks @schernolyas, I'll the PR tomorrow but I already have some comments already:

1. Please, remove the dot after the JIRA issue from the commit message, We simply write `OGM-385` (and not `OGM-385.`)
2. The commit with the checkstyle fixes seems to refer to the new classes you added so you should squash it with the initial commit I guess.
3. Even if this is a preview you should take care of the history log properly:
   1. The only reasonable time to use `initial commit` in a commit message is when you are creating a new repository.
   2. If you split the tests in a separate commit, it is much easier for me to check them and understand what you are trying to achieve.

Keep in mind that even if this is a PREVIEW I still need to review it. It might not work but I still expect a relative clean PR.

schernolyas:
See https://github.com/hibernate/hibernate-ogm/pull/890

PR #890, opened 2017-10-18
[PREVIEW] OGM-359 Support stored procedures

schernolyas:
Hi!
The PR is for review. I need comments and offers.

Please look to interface org.hibernate.ogm.dialect.storedprocedure.spi.StoredProcedureAwareGridDialect

Also, see comments from PR  https://github.com/hibernate/hibernate-ogm/pull/889

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/dialect/storedprocedure/spi/StoredProcedureAwareGridDialect.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/storedprocedure/spi/StoredProcedureAwareGridDialect.java
@@ -1,0 +1,65 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.dialect.storedprocedure.spi;
+
+import java.util.Map;
+
+import org.hibernate.ogm.dialect.query.spi.ClosableIterator;
+import org.hibernate.ogm.dialect.spi.GridDialect;
+import org.hibernate.ogm.dialect.spi.TupleContext;
+
+/**
+ * A facet for {@link GridDialect} implementations which support the execution of stored procedures.
+ *
+ * Cases of stored procedures are :
+ * <ou>
+ *     <li>procedure without any input or output parameters</li>
+ *	   <li>function with many input parameters and one returned value</li>
+ * </ou>
+ *
+ *
+ * @see <a href="https://stackoverflow.com/questions/32480060/call-mongodb-function-from-java">Example for MongoDB</a>
+ * @see <a href="http://orientdb.com/docs/2.0/orientdb.wiki/Functions.html">Example for OrientDB</a>
+ * @author Sergey Chernolyas &amp;sergey_chernolyas@gmail.com&amp;
+ */
+public interface StoredProcedureAwareGridDialect extends GridDialect {
+
+	/**
+	 * Is data storage supports parameter position by name (true) or by index (false)
+	 *
+	 * @return
+	 */
+	boolean supportsNamedPosition();
+
+
+	/**
+	 * Returns the result of a stored procedure executed on the backend.
+	 * Tne method uses for storages that supports name position
+	 *
+	 * @param storedProcedureName name of stored procedure.
+	 * @param params - array with values
+	 * @param tupleContext the tuple context
+	 *
+	 * @return an {@link ClosableIterator} with the result of the query
+	 */
+
+	Object callStoredProcedure( String storedProcedureName, Object[] params, TupleContext tupleContext);
+
+	/**
+	 * Returns the result of a stored procedure executed on the backend.
+	 * Tne method uses for storages that supports name position
+	 *
+	 * @param storedProcedureName name of stored procedure.
+	 * @param params - map with pair "name"-"value"
+	 * @param tupleContext the tuple context
+	 *
+	 * @return an {@link ClosableIterator} with the result of the query
+	 */
+
+	Object callStoredProcedure(	String storedProcedureName, Map<String,Object> params, TupleContext tupleContext);
+
+}
--- a/core/src/main/java/org/hibernate/ogm/hibernatecore/impl/OgmSessionImpl.java
+++ b/core/src/main/java/org/hibernate/ogm/hibernatecore/impl/OgmSessionImpl.java
@@ -50,6 +50,7 @@ import org.hibernate.ogm.loader.nativeloader.impl.BackendCustomQuery;
 import org.hibernate.ogm.options.navigation.GlobalContext;
 import org.hibernate.ogm.query.NoSQLQuery;
 import org.hibernate.ogm.query.impl.NoSQLQueryImpl;
+import org.hibernate.ogm.storedprocedure.impl.NoSQLProcedureCallImpl;
 import org.hibernate.ogm.util.impl.Log;
 import org.hibernate.ogm.util.impl.LoggerFactory;
 import org.hibernate.persister.entity.EntityPersister;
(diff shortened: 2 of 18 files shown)


core/src/main/java/org/hibernate/ogm/hibernatecore/impl/OgmSessionImpl.java
```diff
@@ -177,22 +178,60 @@ public void doWork(Work work) throws HibernateException {
...
 	}
 
 	@Override
 	public ProcedureCall createStoredProcedureCall(String procedureName, Class... resultClasses) {
-		throw new NotSupportedException( "OGM-359", "Stored procedures are not supported yet" );
+		log.infof( "2. createStoredProcedureCall with name %s",procedureName );
+		errorIfClosed();
+		checkTransactionSynchStatus();
+
+		if ( log.isTraceEnabled() ) {
```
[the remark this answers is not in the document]
DavideD: `log.tracev(...)` will already skip the log if the trace level is disabled.

It only makes sense if you need to print a complex log message

core/src/main/java/org/hibernate/ogm/jpa/impl/OgmJpaStoredProcedureQuery.java
```diff
@@ -0,0 +1,152 @@
...
+import org.hibernate.ogm.storedprocedure.impl.NoSQLProcedureOutputImpl;
+import org.hibernate.ogm.util.impl.Log;
+import org.hibernate.ogm.util.impl.LoggerFactory;
+import org.hibernate.procedure.ParameterRegistration;
+import org.hibernate.procedure.ProcedureCallMemento;
+
+/**
+ * @author Sergey Chernolyas &amp;sergey_chernolyas@gmail.com&amp;
+ */
+public class OgmJpaStoredProcedureQuery extends StoredProcedureQueryImpl {
```
DavideD: I would remove the Jpa from the name, it's already in the package

core/src/main/java/org/hibernate/ogm/jpa/impl/OgmJpaStoredProcedureQuery.java
```diff
@@ -0,0 +1,152 @@
...
+		entityManager().checkOpen( true );
+		parameterRegistrations.add( procedureCall.registerParameter( position, type, mode ) );
+		return this;
+	}
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public StoredProcedureQuery registerStoredProcedureParameter(String parameterName, Class type, ParameterMode mode) {
+		entityManager().checkOpen( true );
+		parameterRegistrations.add( procedureCall.registerParameter( parameterName, type, mode ) );
```
DavideD: I'm surprised that you need to override these methods, in the end you are registering parameters couldn't you reuse the existing ones?

core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureCallImpl.java
```diff
@@ -0,0 +1,133 @@
...
+	public <T> ParameterRegistration<T> registerParameter(int position, Class<T> type, ParameterMode mode) {
+		NoSQLProcedureParameterRegistration<T> parameterRegistration =
+				new NoSQLProcedureParameterRegistration<T>( this, position, null, mode, type );
+		registerParameter( parameterRegistration );
+		return parameterRegistration;
+	}
+
+	@Override
+	public ProcedureOutputs getOutputs() {
+		log.info( "I am here!" );
```
DavideD: Terrible log

core/src/main/java/org/hibernate/ogm/dialect/storedprocedure/spi/StoredProcedureAwareGridDialect.java
```diff
@@ -0,0 +1,65 @@
...
+	 * Tne method uses for storages that supports name position
+	 *
+	 * @param storedProcedureName name of stored procedure.
+	 * @param params - array with values
+	 * @param tupleContext the tuple context
+	 *
+	 * @return an {@link ClosableIterator} with the result of the query
+	 */
+
+	Object callStoredProcedure( String storedProcedureName, Object[] params, TupleContext tupleContext);
```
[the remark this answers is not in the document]
schernolyas: In this way, stored procedure have to return a collection of Entity.  But ... What will we do with case then stored procedure returns simple type?  For example, Integer

core/src/main/java/org/hibernate/ogm/jpa/impl/OgmJpaStoredProcedureQuery.java
schernolyas: I tried. Existing implementation coupled with JDBC. I need remove these links.

core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureCallImpl.java
schernolyas: yes. I needed these logs for make travel in lot of classes :). Now ... logs like this  must be removed. I will do it.

core/src/main/java/org/hibernate/ogm/jpa/impl/OgmJpaStoredProcedureQuery.java
schernolyas: ok

core/src/main/java/org/hibernate/ogm/dialect/storedprocedure/spi/StoredProcedureAwareGridDialect.java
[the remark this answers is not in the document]
DavideD: A tuple is not necessary an entity, it can contain anything.
 
You can check the implementations of `QueryableGridDialect#executeBackendQuery` to see what we already do. You will find that what we return is an iterator of map containing many different values or an iterator of the underlying object that represents the entity in the db, for example nodes in Neo4j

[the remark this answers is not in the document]
DavideD: If you need to return only one value it will be a tuple with only one column.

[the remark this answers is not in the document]
schernolyas: ok. I will investigate

core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureCallImpl.java
DavideD: I usually work with a debugger

core/src/main/java/org/hibernate/ogm/jpa/impl/OgmJpaStoredProcedureQuery.java
DavideD: Well see if you can update the other things, we can come back to this later

schernolyas:
"
I would also like to see a test with Neo4j (http://neo4j.com/docs/developer-manual/current/extending-neo4j/procedures/) or MongoDB (http://dirolf.com/2010/04/05/stored-javascript-in-mongodb-and-pymongo.html) to see that the feature actually works.
"
Do you mean real test? .... Okey ... I will do for mongodb.
But ... What cases do you need?
I offer :
1) procedure returns single simple value (For example, Integer)
2) procedure returns one document
3) procedure returns several documents
Is it good script?

DavideD:
> Is it good script?

In core (you can skipt this tests for the dialect not supporting this feature),
 we should check that the procedure returns:
 1. single simple value (For example: Integer)
 2. multiple simple values
 3. single managed entity
 4. list of entities

I MonogDB I guess you can check that we can return documents or DBObject, what makes more sense.

core/src/main/java/org/hibernate/ogm/jpa/impl/OgmJpaStoredProcedureQuery.java
schernolyas: fixed

DavideD:
I'm closing this for now, please, reopen it or send a PR when done. Thanks

PR #895, opened 2017-10-26
[PREVIEW] OGM-359 support stored procedures

schernolyas:
The PR contains improvements of the feature in 'core' module. Also, the PR contains implementation of the feature for MondoDB.

Comments are welcome!


DavideD:
Is this a preview or is it ready to be merged?

schernolyas:
Hi @DavideD !

I think that it is "preview" PR. Because I didn't all cases that appropriate for stored procedures.
I mean out parametes and etc. 
What is borders of the task?

I offer discuss about it. And ...   I prepare PR with status "ready for commit".
I offer the road map.
What do you think?

core/src/main/java/org/hibernate/ogm/dialect/storedprocedure/spi/StoredProcedureAwareGridDialect.java
```diff
@@ -0,0 +1,66 @@
...
+ *     <li>procedure without any input or output parameters</li>
+ *	   <li>function with many input parameters and one returned value</li>
+ * </ou>
+ *
+ *
+ * @see <a href="https://stackoverflow.com/questions/32480060/call-mongodb-function-from-java">Example for MongoDB</a>
+ * @see <a href="http://orientdb.com/docs/2.0/orientdb.wiki/Functions.html">Example for OrientDB</a>
+ * @author Sergey Chernolyas &amp;sergey_chernolyas@gmail.com&amp;
+ */
+public interface StoredProcedureAwareGridDialect extends GridDialect {
```
DavideD: I would prefer not to have a new Interface for this feature, calling a stored procedure is not much different than running a native query. Can't we add the `callStoredProcedure` method in the `QueryableGridDialect` interface?

core/src/main/java/org/hibernate/ogm/dialect/storedprocedure/spi/StoredProcedureAwareGridDialect.java
```diff
@@ -0,0 +1,66 @@
...
+	 * Tne method uses for storages that supports name position
+	 *
+	 * @param storedProcedureName name of stored procedure.
+	 * @param params - map with pair "name"-"value"
+	 * @param tupleContext the tuple context
+	 *
+	 * @return an {@link ClosableIterator} with the result of the query
+	 */
+
+	ClosableIterator<Tuple> callStoredProcedure( String storedProcedureName, Map<String,Object> params, TupleContext tupleContext);
```
DavideD: Instead of maps or arrays I think a class would be better for the parameters. Can't we reuse QueryParameters? Or an extension of it?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1636,6 +1639,84 @@ private static ReadPreference getReadPreference(AssociationContext associationCo
...
+	public boolean supportsNamedPosition() {
+		return false;
+	}
+
+	/**
+	 * call stored procedure
+	 * @param storedProcedureName name of stored procedure.
+	 * @param params - array with values
+	 * @param tupleContext the tuple context
+	 *@see <a href="https://stackoverflow.com/questions/32480060/call-mongodb-function-from-java"> example</a>
```
DavideD: We usually use `@see` to link to another  class or method in the project. We usually put links in the description.

In this case I would simply remove the link, I don't think it's needed.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1636,6 +1639,84 @@ private static ReadPreference getReadPreference(AssociationContext associationCo
...
+	public ClosableIterator<Tuple> callStoredProcedure(String storedProcedureName, Object[] params,
+			TupleContext tupleContext) {
+		StringBuffer commandLine = new StringBuffer( storedProcedureName ).append( "(" );
+		for ( Object param : params ) {
+			if ( param instanceof String ) {
+				commandLine.append( '\'' ).append( param.toString() ).append( '\'' );
+			}
+			else {
+				commandLine.append( param.toString() );
+			}
```
DavideD: I don't understand this code, in both cases you are converting the parameter to a String but only  in the first case you are adding quotes.

I would expect something like:

    if( hasToBeEscaped( param )) { ... }

Where` hasToBeEscaped` checks if the value has some invalid characters.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1636,6 +1639,84 @@ private static ReadPreference getReadPreference(AssociationContext associationCo
...
+	 * @param params - array with values
+	 * @param tupleContext the tuple context
+	 *@see <a href="https://stackoverflow.com/questions/32480060/call-mongodb-function-from-java"> example</a>
+	 * @return
+	 */
+	@SuppressWarnings("unchecked")
+	@Override
+	public ClosableIterator<Tuple> callStoredProcedure(String storedProcedureName, Object[] params,
+			TupleContext tupleContext) {
+		StringBuffer commandLine = new StringBuffer( storedProcedureName ).append( "(" );
```
DavideD: Why not` StringBuilder`?

core/src/main/java/org/hibernate/ogm/dialect/storedprocedure/spi/StoredProcedureAwareGridDialect.java
DavideD: Removing this interface will also make it possible to remove Initiator

core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureOutputImpl.java
```diff
@@ -0,0 +1,58 @@
...
+		return !resultList.isEmpty() && !isTuple();
+	}
+
+	private boolean isTuple() {
+		return resultList.get( 0 ) instanceof Tuple;
+	}
+
+	@Override
+	public List getResultList() {
+		log.info( "I am called!" );
```
DavideD: Remember to remove this

core/src/test/java/org/hibernate/ogm/backendtck/storedprocedures/indexed/IndexedStoredProcedureCallTest.java
```diff
@@ -0,0 +1,200 @@
...
+		assertThat( call4 ).isInstanceOfAny( OgmStoredProcedureQuery.class );
+		call4.registerStoredProcedureParameter( 0,Void.class, ParameterMode.REF_CURSOR );
+		call4.registerStoredProcedureParameter( 1, String.class, ParameterMode.IN );
+		call4.registerStoredProcedureParameter( 2, String.class, ParameterMode.IN );
+		call4.setParameter( 1, "id1" );
+		call4.setParameter( 2, "title1" );
+		listResult = call4.getResultList();
+		assertThat( listResult ).isNotNull();
+		assertThat( listResult.size() ).isEqualTo( 1 );
+		assertThat( listResult.get( 0 ) ).isEqualTo( new Car( "id1","title1" ) );
```
DavideD: These should be 2 separate tests

core/src/test/java/org/hibernate/ogm/backendtck/storedprocedures/indexed/IndexedStoredProcedureCallTest.java
```diff
@@ -0,0 +1,200 @@
...
+		assertThat( listResult.get( 0 ) ).isEqualTo( new Car( "id", "title" ) );
+
+		StoredProcedureQuery call4 = em.createNamedStoredProcedureQuery( "testproc4_2" );
+		assertThat( call4 ).isInstanceOfAny( OgmStoredProcedureQuery.class );
+		call4.setParameter( 1, "id1" );
+		call4.setParameter( 2, "title1" );
+		listResult = call4.getResultList();
+		assertThat( listResult ).isNotNull();
+		assertThat( listResult.size() ).isEqualTo( 1 );
+		assertThat( listResult.get( 0 ) ).isEqualTo( new Car( "id1", "title1" ) );
```
DavideD: I think I would prefer to have these as separate tests, maybe with different parameters. It seems to me that you are testing the same case (both parameters are strings, they return the same single object). You shoudl also test what happens when you pass parameters with weird chacters like `D'Alto`. Will it work?

core/src/test/java/org/hibernate/ogm/backendtck/storedprocedures/indexed/IndexedStoredProcedureCallTest.java
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	@Override
+	protected void configure(GetterPersistenceUnitInfo info) {
+
+		Properties properties = info.getProperties();
+
+		if ( TestHelper.getCurrentDialectType().equals( GridDialectType.HASHMAP ) ) {
```
DavideD: Let's avoid this, please. Let's simply skip this tests for the dialects that don't support storedProcedures

DavideD: You can simply create a separate test in the package `org.hibernate.ogm.backendtck` that uses IndexedStoredProcProvider

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1636,6 +1639,84 @@ private static ReadPreference getReadPreference(AssociationContext associationCo
...
+			if ( firstRetVal instanceof List ) {
+				//it is collection of entities
+				List<Document> documents = (List<Document>) firstRetVal;
+				resultTuples = new ArrayList<>( documents.size() );
+				for ( Document doc : documents ) {
+					resultTuples.add( convert( doc ) );
+				}
+			}
+			else {
+				throw new NotYetImplementedException( "Not supported type " + firstRetVal.getClass() );
```
DavideD: What other value could this return?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1636,6 +1639,84 @@ private static ReadPreference getReadPreference(AssociationContext associationCo
...
+	 * @return
+	 */
+	@SuppressWarnings("unchecked")
+	@Override
+	public ClosableIterator<Tuple> callStoredProcedure(String storedProcedureName, Object[] params,
+			TupleContext tupleContext) {
+		StringBuffer commandLine = new StringBuffer( storedProcedureName ).append( "(" );
+		for ( Object param : params ) {
+			if ( param instanceof String ) {
+				commandLine.append( '\'' ).append( param.toString() ).append( '\'' );
```
DavideD: If it's a String you can simply cast the value `(String) param`

core/src/main/java/org/hibernate/ogm/dialect/storedprocedure/spi/StoredProcedureAwareGridDialect.java
schernolyas: If we look to relational databases ... then StoredProcedureAwareGridDialect extends QueryableGridDialect. I mean than not all implementations of QueryableGridDialect supports stored procedures.  In this way, QueryableGridDialect and StoredProcedureAwareGridDialect are different and moving  callStoredProcedure to QueryableGridDialectis is not logical.

core/src/main/java/org/hibernate/ogm/dialect/storedprocedure/spi/StoredProcedureAwareGridDialect.java
schernolyas: Cool idea!  I will think !

core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureOutputImpl.java
schernolyas: ok. thanks!

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
schernolyas: ok

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
schernolyas: You right. I just rushed :(

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
DavideD: OK, I get it now. This method seems too simple , I think you should check:

1) What happens if the String contains `'`. You need to escape other characters for sure.
2) What happens if the object is not a String but the result of the toString()  is actually a String (and not a Number for example)?

schernolyas: Because strings must be escaped ... Okey ... good idea ..  I will correct my code.

DavideD:
@schernolyas I've finished with the review.
Please, let me know when you applied all the remarks

core/src/test/java/org/hibernate/ogm/backendtck/storedprocedures/indexed/IndexedStoredProcedureCallTest.java
schernolyas: Okey.  I will separate tests. Thanks

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
schernolyas: I don't know. 
Do you offer to remove the code?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
schernolyas: You right :+1:

core/src/main/java/org/hibernate/ogm/dialect/storedprocedure/spi/StoredProcedureAwareGridDialect.java
DavideD: I'm not saying that it's not logical, but I don't think it's worth the complexity in our use case.

If the dialect does not support the feature you can throw an `UnsupportedOperationException` when an error occurs.  Right now I'm not even sure what happens if you try to call a stored procedure and the dialect does not support it. Probably `UnknownServiceException`.
Mainly. I don't think this needs to be a service, maybe you can keep the interface and check in `NoSQLProcedureCallImpl` if the dialect is an instance of this interface before  assign it to the field `gridDialect`.

DavideD: Remember to add a test to check that the right exception is thrown when the dialect does not support stored procedures.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
gsmet: This one bothers me a bit. We don't escape the string before quoting it? It looks like a source for "NoSQL injection".

It would be nice if we could find an API in the MongoDB driver allowing to escape the strings properly.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1636,6 +1639,84 @@ private static ReadPreference getReadPreference(AssociationContext associationCo
...
+				commandLine.append( param.toString() );
+			}
+			commandLine.append( "," );
+		}
+		commandLine.setLength( commandLine.length() - 1 );
+		commandLine.append( ")" );
+		Document result = provider.getDatabase().runCommand( new Document( "$eval", commandLine.toString() ) );
+		Double ok = result.getDouble( "ok" );
+		if ( ok.intValue() != 1 ) {
+			// return not ok! throw exception?
```
gsmet: I suppose you should do something here?

schernolyas: Hi @gsmet !
MongoDB returns result of execution in parameter "ok". Type of value is double. If execution is success ... "ok"=1.0.

In this way ... I think Should I throw exception if "ok"<>1.0 ?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
DavideD: Could you create a test and check please?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
DavideD: I guess it would be nice to have a test

DavideD: Yes, let's throw an exception, but please, add a test .

schernolyas: okey @DavideD

schernolyas:
@DavideD , please  not close the PR. It needs for save comments.
OK?

core/src/main/java/org/hibernate/ogm/dialect/storedprocedure/spi/StoredProcedureAwareGridDialect.java
schernolyas: @DavideD  ... about "when the dialect does not support stored procedures." Ok.

DavideD:
> @DavideD , please not close the PR. It needs for save comments.
> OK?

For future reference: even if the PR is closed you can still access to it by disabling the filter in the pull request tab or using the link in the email you received.
Sometimes i have to close the PRs because they create too much noise in the pull requests view on GitHub

DavideD:
@schernolyas  Even if it's closed, it doesn't mean it has been deleted

core/src/main/java/org/hibernate/ogm/dialect/storedprocedure/spi/StoredProcedureAwareGridDialect.java
schernolyas: fixed.

core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureOutputImpl.java
schernolyas: fixed

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
schernolyas: fixed

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
schernolyas: I added escaping for String parameters

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
schernolyas: fixed.

core/src/test/java/org/hibernate/ogm/backendtck/storedprocedures/indexed/IndexedStoredProcedureCallTest.java
schernolyas: fixed.

core/src/test/java/org/hibernate/ogm/backendtck/storedprocedures/indexed/IndexedStoredProcedureCallTest.java
schernolyas: fixed

core/src/test/java/org/hibernate/ogm/backendtck/storedprocedures/indexed/IndexedStoredProcedureCallTest.java
schernolyas: fixed

schernolyas:
Hi @DavideD !

Your comments have fixed. PR is ready for your review.

I will prepare 'ready for merge' PR if you haven't any questions.

DavideD:
@schernolyas Could you rebase this to the latest changes, please?

schernolyas:
Okey @DavideD.  "ready for merge" PR will rebased.

schernolyas:
Sent "ready for merge" PR https://github.com/hibernate/hibernate-ogm/pull/901

PR #900, opened 2017-10-30
OGM-1338 avoid cost initialization of logger

DavideD:
https://hibernate.atlassian.net/browse/OGM-1338

It changes the LoggerFactory, it won't throw an exception anymore to obtain the class using the logger.
The goal is to improve performance at startup


Sanne:
re-test this please

core/src/main/java/org/hibernate/ogm/massindex/impl/OptionallyWrapInJTATransaction.java
```diff
@@ -36,7 +37,7 @@
  */
 public class OptionallyWrapInJTATransaction implements ModelConsumer {
 
-	private static final Log log = LoggerFactory.make();
+	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
```
gsmet: Not related to this patch but the fact that the initialization of the logger was different made me take a look to what we are doing.

Thinking out loud, I'm wondering if we could get rid of these copies of Search classes. I think I'll take a look as part of the ORM 5.2/Search 5.8 upgrade.

core/src/main/java/org/hibernate/ogm/util/impl/LoggerFactory.java
```diff
@@ -6,25 +6,27 @@
...
+ * @author Davide D'Alto
  */
 public class LoggerFactory {
 
-	private static final CallerProvider callerProvider = new CallerProvider();
-
-	public static Log make() {
-		return Logger.getMessageLogger( Log.class, callerProvider.getCallerClass().getCanonicalName() );
+	private LoggerFactory() {
+		// now allowed
```
gsmet: Should be `not allowed`. I'll fix it when merging it.

gsmet:
Merged, thanks!

core/src/main/java/org/hibernate/ogm/massindex/impl/OptionallyWrapInJTATransaction.java
DavideD: We can get rid of these but we need to create similar error log in our logger, not sure why we used the one from search, probably a mistake in the import. It's OK if you want to have a lok

gsmet: I'm not talking about getting rid of the logger class. I'm talking about getting rid of the whole `OptionallyWrapInJTATransaction` class (and probably some others in the package). AFAICS, the Search API are far more customizable that they used to be.

Not sure it's possible but it's worth having a fresh look at it, I think.

PR #902, opened 2017-11-06
OGM-1339 Support geospatial queries for MongoDB

gsmet:
* https://hibernate.atlassian.net/browse/OGM-1339

@DavideD so here is my attempt at supporting geospatial queries for MongoDB. I added a `GeoPoint` type that end users can use for that.

I'm not sure it's something we should do but I think providing an easy way to have spatial queries in their applications could be appealing to users.

We could also try to support more types: https://docs.mongodb.com/manual/reference/geojson/ . It shouldn't be too hard.

Another possibility could be to allow an arbitrary object to be serialized as BSON to be stored in MongoDB and unserialized to a POJO. This project might be useful for that: https://github.com/michel-kraemer/bson4jackson . But it will probably be a bit fragile and defeat the purpose of using a JPA mapping.

The patch shouldn't be committed as is. Documentation is missing but I wanted to have a preliminary feedback before writing it.

Btw, I changed a bit how the type of the index is defined to be more generic. It still supports the `text: true` syntax but it's now recommended to use `_type: 'text'` or `_type: '2dsphere'`. I used an underscore to avoid conflicts with the MongoDB options.

### DIFF
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/index/impl/MongoDBIndexType.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/index/impl/MongoDBIndexType.java
@@ -1,0 +1,61 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.datastore.mongodb.index.impl;
+
+import org.hibernate.ogm.options.shared.IndexOption;
+import org.hibernate.ogm.util.impl.StringHelper;
+
+/**
+ * Supported MongoDB index types. Defined as a "_type" entry in {@link IndexOption}.
+ *
+ * @author Guillaume Smet
+ */
+public enum MongoDBIndexType {
+
+	NORMAL(null),
+
+	/**
+	 * Full text index.
+	 */
+	TEXT("text"),
+
+	/**
+	 * Geospatial index on a 2D sphere.
+	 */
+	TWODSPHERE("2dsphere"),
+
+	/**
+	 * Geospatial index on a 2D plan.
+	 */
+	TWOD("2d");
+
+	private String type;
+
+	private MongoDBIndexType(String type) {
+		this.type = type;
+	}
+
+	public String getType() {
+		return type;
+	}
+
+	public static MongoDBIndexType from(String type) {
+		if ( StringHelper.isEmpty( type ) ) {
+			return NORMAL;
+		}
+		if ( TEXT.type.equals( type ) ) {
+			return TEXT;
+		}
+		else if ( TWODSPHERE.type.equals( type ) ) {
+			return TWODSPHERE;
+		}
+		else if ( TWOD.type.equals( type ) ) {
+			return TWOD;
+		}
+		throw new IllegalArgumentException( "Unsupported MongoDB index type " + type );
+	}
+}
--- a/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/Restaurant.java
+++ b/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/Restaurant.java
@@ -1,0 +1,73 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.datastore.mongodb.test.index;
+
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Index;
+import javax.persistence.Table;
+
+import org.hibernate.ogm.datastore.mongodb.type.GeoPoint;
+import org.hibernate.ogm.options.shared.IndexOption;
+import org.hibernate.ogm.options.shared.IndexOptions;
+
+@Entity
+@Table(name = Restaurant.COLLECTION_NAME, indexes = {
+		@Index(columnList = "location", name = "location_spatial_idx")
+})
+@IndexOptions(
+		@IndexOption(forIndex = "location_spatial_idx", options = "{ _type: '2dsphere' }")
+)
+public class Restaurant {
+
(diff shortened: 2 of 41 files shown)


mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/index/impl/MongoDBIndexType.java
```diff
@@ -0,0 +1,61 @@
...
+	 * Geospatial index on a 2D sphere.
+	 */
+	TWODSPHERE("2dsphere"),
+
+	/**
+	 * Geospatial index on a 2D plan.
+	 */
+	TWOD("2d");
+
+	private String name;
```
DavideD: I think this is the worst name you could use :)

It's easy to confuse with the `name()`

DavideD: Maybe `value` or `type`?

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/Restaurant.java
```diff
@@ -0,0 +1,73 @@
...
+import org.hibernate.ogm.options.shared.IndexOption;
+import org.hibernate.ogm.options.shared.IndexOptions;
+
+@Entity
+@Table(name = Restaurant.COLLECTION_NAME, indexes = {
+		@Index(columnList = "location", name = "location_spatial_idx")
+})
+@IndexOptions(
+		@IndexOption(forIndex = "location_spatial_idx", options = "{ _type: '2dsphere' }")
+)
```
DavideD: As an evolution, would it be possible to deduce this from the fact that we are using a GeoPoint?

DavideD:
It can work as a solution, to be fair I had something different in mind.
Something similar to what Hibernate Search does:

```
@Embeddable
@GeoPoint
public class GeoPoint {

   @Latitude
   double latitude;

   @Longitude
   double longitude;
} 
```
and then we could have restaurant:

```
class Restaurant {

    @Type( "2dsphere")
    GeoPoint location;
}
```

Anyway, your solution works fine for me

gsmet:
About the index type, I really think it makes more sense to keep it in the `@IndexOptions` right along with the `text` option as it's exactly the same thing for MongoDB.

About the annotation proposal, it sure looks more as what we use to do in Search and is probably a bit more sexy. What I dislike with this approach is that it's not that easy to introduce the notion of polygons. We have the same issue with Search and that's something I would like to address in 6.

By the way, not sure the annotation approach is that easy. Do you have an idea on how to implement it?

@Sanne @yrodiere ping as maybe we should try to design something similar in OGM and Search!

DavideD:
> About the index type, I really think it makes more sense to keep it in the @IndexOptions right along with the text option as it's exactly the same thing for MongoDB.

Works for me. I don't like that solution so much because using Index annotation is verbose and using the type annotation keeps everything a bit more abstract.

> By the way, not sure the annotation approach is that easy. Do you have an idea on how to implement it?

I haven't thought about the implementation details. It should be doable  but I just  wanted to put on the table some alternatives.

 > What I dislike with this approach is that it's not that easy to introduce the notion of polygon

Isn't that a list of coordinates?

```
class Restaurant {

    @ElementCOllection
    @Type( "MultiPoint")
    List<GeoPoint> locations;
}
```

How would you define a polygon with the current approach?

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/Restaurant.java
gsmet: Yes, it would be possible to have some sort of default type if the property is a GeoPoint.

DavideD:
> @Sanne @yrodiere ping as maybe we should try to design something similar in OGM and Search!

That would be great

gsmet:
> Works for me. I don't like that solution so much because using Index annotation is verbose and using the type annotation keeps everything a bit more abstract.

As mentioned above, it would be possible to choose 2dsphere as the default if the object is a `GeoPoint`.

> I haven't thought about the implementation details. It should be doable but I just wanted to put on the table some alternatives.

Yes, I agree it's not very clean to impose some model choices to the user. It's usually not what we do.

I'm just not familiar enough with OGM to be sure the annotation approach would work.

To be honest, even if not very clean, I find the type approach much easier for the user as he would just have to use carefully prepared types to implement his spatial features.

In Search, it's a bit different, as the search index is an abstraction of the model so we don't have to stick to it. Here, as we have to stick to the model, I don't think it's such a bad approach as the user would have to implement `GeoPoint` one way or another.

> [About polygons] Isn't that a list of coordinates?

It's a bit more complicated than that as you can have a polygon defined with multiple rings. See https://docs.mongodb.com/manual/reference/geojson/#polygon . But, yes you get the idea.

With my approach, I would have introduced `GeoPolygon`, `GeoLineString` and so on.

DavideD:
I'm OK moving on with this approach. We can change it later on if we see it make sense

gsmet:
@DavideD I completed the PR with all GeoJSON basic types. I let CollectionOfGeometry aside for now. Might be a good task for a new contributor.

Once this is in, I'll work on the documentation.

DavideD:
Thanks

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/spatial/MongoDBGeoObjectTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+	@Test
+	public void testBoundingBox() throws Exception {
+		try ( OgmSession session = openSession() ) {
+			checkBoundingBoxResultForField( session, "point" );
+			checkBoundingBoxResultForField( session, "multiPoint" );
+			checkBoundingBoxResultForField( session, "lineString" );
+			checkBoundingBoxResultForField( session, "multiLineString" );
+			checkBoundingBoxResultForField( session, "polygon" );
+			checkBoundingBoxResultForField( session, "multiPolygon" );
```
DavideD: Just curious, any specific reason to not create a single test for each type?

gsmet: Other than laziness? No :).

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/902

DavideD:
Thanks @gsmet, I will wait for the documentation before closing the Jira

DavideD:
Could you also create the Jira for the missing type? Please

gsmet:
Done: https://hibernate.atlassian.net/browse/OGM-1346

PR #908, opened 2017-11-16
[PREVIEW] OGM-1301 Improve MongoDB native query parser (#15)

DMasherov:
See task https://hibernate.atlassian.net/browse/OGM-1301


gsmet:
add to whitelist

gsmet:
@DMasherov hi! Thanks for the PR. Looks interesting. Is there a reason why you made it an entirely different file?

Isn't the new parser supposed to replace entirely the old one? It makes it a bit harder to track the difference between the two parsers.

schernolyas:
Hi @gsmet !

@DMasherov  is new in the project. I hope defferent class is not block your review. 
For track the differences we will replace old parser by new during some time.

DMasherov:
Sorry about that, I just wanted to show this as a prototype. It was just easier for me to create it this way, since I changed the type of parsing result and had to modify all tests. I'll replace the parser if all is OK.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/parsing/nativequery/NativeQueryParserTest.java
```diff
@@ -61,37 +56,25 @@ public void shouldParseComplexAggregateQuery() {
...
 		assertThat( queryDescriptor.getOperation() ).isEqualTo( Operation.AGGREGATE_PIPELINE );
-		assertThat( queryDescriptor.getPipeline() )
-			.containsExactly(
-					Document.parse( match )
-					, Document.parse( unwind )
-					, Document.parse( group )
-					, Document.parse( sort )
-					);
+		assertThat( queryDescriptor.getPipeline() ).containsExactly( Document.parse( match ), Document.parse( unwind ), Document.parse( group ),
+				Document.parse( sort ) );
```
gsmet: Please avoid to reformat automatically our files. It makes it very difficult to find the real changes in there.

To fix it, use your IDE to compare both versions then bring back all the changes that are purely reformatting.

Thanks!

schernolyas: Of course @gsmet ! I try to correct it

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -897,13 +896,13 @@ public int executeBackendUpdateQuery(final BackendQuery<MongoDBQueryDescriptor>
 	@Override
 	public MongoDBQueryDescriptor parseNativeQuery(String nativeQuery) {
 		NativeQueryParser parser = Parboiled.createParser( NativeQueryParser.class );
```
gsmet: This is preexisting to your patch but seeing this, I think we have made a design error here. The parser should be created only once and should not contain any state. The state should be kept in the value stack.

I think it's important to fix this now as you are adding a new state in your patch (`cliQueryInfo`) so it might be a bit harder to migrate to a value stack after these changes.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
```diff
@@ -57,301 +67,300 @@
...
+		builder.setCollection( cliQueryInfo.collectionName );
+		switch ( cliQueryInfo.operationName ) {
+			case "find": {
+				builder.setOperation( Operation.FIND );
+				builder.setCriteria( cliQueryInfo.arguments.get( 0 ) );
+				if ( cliQueryInfo.arguments.size() > 1 ) {
+					builder.setProjection( cliQueryInfo.arguments.get( 1 ) );
+				}
+				break;
+			}
```
gsmet: What annoys me a bit with this new approach is that we are far less selective into what would be a valid option for `find()` (same for any other option). We treat all arguments equal and I'm wondering if it's such a good idea.

I think I would have chosen to keep the existing approach and have a default catch all option that would:
 * say that the syntax is incorrect if the action is in the list of supported command
 * either say the command is not supported for now (for existing commands that we don't support yet),
 * or is an invalid command.

i'm not sure though how it would be compatible with saying that the syntax of `find()` is not valid with an error in the parser. That's probably why I thought we would do that in the query builder instead.

Not sure what's the best outcome here. @DMasherov WDYT about it?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
```diff
@@ -57,301 +67,300 @@
...
+					builder.setCriteria( cliQueryInfo.arguments.get( 1 ) );
+				}
+				if ( cliQueryInfo.arguments.size() > 2 ) {
+					builder.setOptions( cliQueryInfo.arguments.get( 2 ) );
+				}
+				break;
+			}
+			case "mapReduce": {
+				builder.setOperation( Operation.MAP_REDUCE );
+				builder.setMapFunction( JSON.parse( cliQueryInfo.arguments.get( 0 ) ).toString() );
```
gsmet: Is there a reason why you use `JSON.parse()` for some of the options? I would have thought it would be either something to do for everything or something to avoid.

I'm also a bit curious on what sort of error we have when we have an invalid JSON here.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
```diff
@@ -57,301 +67,300 @@
...
+			context.getParseErrors().add(
+					createParseError(
+							context, 2, "First identifier should be 'db'"
+							)
+					);
+			return false;
+		}
+		cliQueryInfo.operationName = identifiers.get( identifiers.size() - 1 );
+		cliQueryInfo.collectionName = String.join( ".", identifiers.subList( 1, identifiers.size() - 1 ) );
+		if ( RESERVED.contains( cliQueryInfo.collectionName ) ) {
```
gsmet: Are we sure that these keywords are reserved and can't be used for a collection name? (just checking)

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
```diff
@@ -57,301 +67,300 @@
...
-	final MongoDBQueryDescriptorBuilder builder;
+public class NativeQueryParser extends BaseParser<MongoDBQueryDescriptor> {
+
+	// TODO There are many more query types than what we support.
+	protected static List<String> RESERVED = Arrays.asList(
+			"findOne", "findAndModify", "find", "insertOne", "insertMany", "insert",
+			"remove", "deleteOne", "deleteMany", "updateOne", "updateMany", "update",
+			"replaceOne", "count", "aggregate", "distinct", "mapReduce" );
+	protected MongoDBQueryDescriptorBuilder builder = new MongoDBQueryDescriptorBuilder();
+	protected CliQueryInfo cliQueryInfo = new CliQueryInfo();
```
gsmet: So following what I said above about the cliQueryInfo, we should either move it to the query builder. Or have a stack with several values in it.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
```diff
@@ -57,301 +67,300 @@
...
 
-	public Rule MapReduce() {
-		return Sequence(
-				Separator(),
-				"mapReduce ",
-				"( ",
-				Sequence( JsonString(), builder.setMapFunction( JSON.parse( match() ).toString() ) ),
-				Sequence( ", ", JsonString(), builder.setReduceFunction( JSON.parse( match() ).toString() ) ),
-				Optional( Sequence( ", ", JsonObject(), builder.setOptions( match() ) ) ),
-				") "
```
gsmet: So what I'm wondering if that maybe we could have something like:
```
FirstOf(
    the valid sequence of parameters (i.e. what I pasted below),
    throw a parse error
)
```
instead of
```
			       "( ",
				Sequence( JsonString(), builder.setMapFunction( JSON.parse( match() ).toString() ) ),
				Sequence( ", ", JsonString(), builder.setReduceFunction( JSON.parse( match() ).toString() ) ),
				Optional( Sequence( ", ", JsonObject(), builder.setOptions( match() ) ) ),
				") "
```

gsmet:
OK, so I added some comments. I think you have understood pretty well how Parboiled works so that's cool :).

The part identifying db.collection.operation is a good idea I think. It's probably the best way to do it.

I'm more hesitant about your proposal on how you would deal with the operations. Could you try to see if my suggestions make sense and you can get it working? Feel free to ask further questions if I'm not clear or even push a partial patch on one of the operation to see if we are on the same page.

Note that my suggestions might not work so feel free to challenge them!

Thanks!

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/parsing/nativequery/NativeQueryParserTest.java
schernolyas: fixed

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
DMasherov: I think it is right to use the value stack instead of keeping the state in the parser. I'll try to reimplement it this way.  I'm not sure though about using the state in the value stack because the parser run result is the state (builder) rather then the object we need (descriptor).  For me it feels awkward to use `run.resultValue.build()` when it could be simply `run.resultValue` and only the parser holds the errors and not both the builder and parser.  

If I understand Parboiled well enough, we could parameterize the parser over `Object` thus making the stack to have values of `Object`.  This way we don't really have to keep any state: the "lower level" rules can put information like the collection name to the stack and the "higher level" rules can pop it and make a decision to fail and put some errors in the error list (although then we have to keep casting things).

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
DMasherov: My bad, I'm not sure why I put this check.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
DMasherov: I think that the more generic `Query` rule should throw errors, this way the parser can first identify the general structure of a query and fail only when it has all the necessary information.

DMasherov:
I probably won't work on this, I've written new version, which might be better https://github.com/hibernate/hibernate-ogm/pull/909

PR #909, opened 2017-11-19
OGM-1301 Improve error reporting of the MongoDB (#20)

DMasherov:
This is a version based on @gsmet's ideas from https://hibernate.atlassian.net/browse/OGM-1301 (I've written another [version](https://github.com/hibernate/hibernate-ogm/pull/908) but I think this is one is a little better).

Basically now when the parser hits some rules it updates the state of the builder so if the input is not valid the builder is able to produce adequate errors/suggestions. If the parser fails the rule it might also capture the input and put it in the state.

Some details and tricks for this version:
- Invoking `build` method might result in `NativeQueryParseErrror` which contains error messages or wraps JsonParseException.
- `CriteriaOnlyFindQuery` is more restrictive: now it does not match anything and can fail. However the `Query` rule behaves as before: it matches everything and stores the builder in the result stack (this is done by optionally trying the query rules and pushing the builder after).
- When the parser hits the `JsonObject` rule it stores the string if matching has failed. (this is done with `rule / ( action FAIL)`, this almost the same as something used in [peg](http://piumarta.com/software/peg/peg.1.html)  tool `exp ~ { action }`)

(about the stateless parser suggestion in the previous pull request: I think it would be easy to implement this by simply pushing the new builder to the value stack when the `Query` rule is invoked and every time we need the builder call the `peek()` method of the value stack but I'm not sure it is the best way)

### DIFF
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
@@ -36,8 +36,8 @@ public class MongoDBQueryDescriptorBuilder {
 	 * Helper fields for recording the states of the parser
 	 */
 	private boolean isCliQuery;
-	private String invalidJsonInPrefix;
-	private boolean everyParameterIsValid;
+	private String invalidJsonParameter;
+	private boolean areParametersValid;
 	private String operationName;
 	private boolean isQueryValid;
 
@@ -122,13 +122,13 @@ public class MongoDBQueryDescriptorBuilder {
 		return true;
 	}
 
-	public boolean setInvalidJsonInPrefix(String invalidJsonInPrefix) {
-		this.invalidJsonInPrefix = invalidJsonInPrefix;
+	public boolean setInvalidJsonParameter(String invalidJsonParameter) {
+		this.invalidJsonParameter = invalidJsonParameter;
 		return true;
 	}
 
-	public boolean setEveryParameterIsValid(boolean everyParameterIsValid) {
-		this.everyParameterIsValid = everyParameterIsValid;
+	public boolean setParametersValid(boolean areParametersValid) {
+		this.areParametersValid = areParametersValid;
 		return true;
 	}
 
@@ -188,34 +188,36 @@ public class MongoDBQueryDescriptorBuilder {
 			if ( isCliQuery ) {
 				// db ???
 				if ( collection == null ) {
-					throw new NativeQueryParseException( "Cli query should match the format db.collection.oper()" );
+					throw new NativeQueryParseException( "Cli query should match the format db.collection.operation(<arguments>)" );
 				}
 				// db . collection ???
 				if ( operation == null && operationName == null ) {
-					throw new NativeQueryParseException( "Cli query should match the format db.collection.oper()" );
+					throw new NativeQueryParseException( "Cli query should match the format db.collection.operation(<arguments>)" );
 				}
 				// db . collection . UNSUPPORTED
 				if ( operation == null ) {
 					throw new NativeQueryParseException( "Operation '" + operationName + "' is unsupported" );
 				}
 				// db . collection . operation ( ???
-				if ( invalidJsonInPrefix != null ) {
+				if ( invalidJsonParameter != null ) {
 					try {
-						Document.parse( invalidJsonInPrefix );
+						Document.parse( invalidJsonParameter );
 					}
 					catch (JsonParseException e ) {
 						throw new NativeQueryParseException( e );
 					}
 				}
 				// db . collection . operation ( ???
-				if ( !everyParameterIsValid ) {
+				if ( !areParametersValid ) {
 					throw new NativeQueryParseException( "Parameters are invalid for " + operationName );
 				}
-				throw new NativeQueryParseException( "Missing ')'" );
+				else {
+					throw new NativeQueryParseException( "Missing ')'" );
+				}
 			}
 			// the input is not a json object: might be error in { ... } or does not start with '{'
 			try {
-				Document.parse( invalidJsonInPrefix );
+				Document.parse( invalidJsonParameter );
 			}
 			catch (JsonParseException e ) {
 				throw new NativeQueryParseException( e );
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
@@ -18,6 +18,7 @@ import org.hibernate.ogm.datastore.mongodb.query.impl.MongoDBQueryDescriptor.Ope
 import org.hibernate.ogm.util.impl.StringHelper;
 
 import org.bson.Document;
+import org.bson.json.JsonParseException;
 
 /**
  * Builder for {@link MongoDBQueryDescriptor}s.
@@ -30,6 +31,16 @@ public class MongoDBQueryDescriptorBuilder {
 
 	private String collection;
 	private Operation operation;
+
+	/**
+	 * Helper fields for recording the states of the parser
+	 */
+	private boolean isCliQuery;
+	private String invalidJsonInPrefix;
+	private boolean everyParameterIsValid;
+	private String operationName;
+	private boolean isQueryValid;
+
 	/**
 	 * Overloaded to be the 'document' for a FINDANDMODIFY query (which is a kind of criteria),
 	 */
@@ -102,6 +113,35 @@ public class MongoDBQueryDescriptorBuilder {
 		return true;
 	}
 
+	public boolean isCliQuery() {
+		return isCliQuery;
+	}
+
+	public boolean setCliQuery(boolean isCliQuery) {
+		this.isCliQuery = isCliQuery;
+		return true;
+	}
+
+	public boolean setInvalidJsonInPrefix(String invalidJsonInPrefix) {
+		this.invalidJsonInPrefix = invalidJsonInPrefix;
+		return true;
+	}
+
+	public boolean setEveryParameterIsValid(boolean everyParameterIsValid) {
+		this.everyParameterIsValid = everyParameterIsValid;
+		return true;
+	}
+
+	public boolean setOperationName(String operationName) {
+		this.operationName = operationName;
+		return true;
+	}
+
+	public boolean setQueryValid(boolean isQueryValid) {
+		this.isQueryValid = isQueryValid;
+		return true;
+	}
+
 	public boolean setCriteria(String criteria) {
 		this.criteria = criteria;
 		return true;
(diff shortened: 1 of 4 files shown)


mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
```diff
@@ -142,7 +182,46 @@ public boolean setReduceFunction(String reduceFunction) {
...
 	}
 
-	public MongoDBQueryDescriptor build() {
+	public MongoDBQueryDescriptor build() throws NativeQueryParseException {
+		if ( !isQueryValid ) {
+			// the input is CLI query: it starts with 'db'
+			if ( isCliQuery ) {
+				// db ???
+				if ( collection == null ) {
+					throw new NativeQueryParseException( "Cli query should match the format db.collection.oper()" );
```
gsmet: Maybe something like `db.collection.operation(<arguments>)` would be better.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
```diff
@@ -142,7 +182,46 @@ public boolean setReduceFunction(String reduceFunction) {
...
+				}
+				// db . collection ???
+				if ( operation == null && operationName == null ) {
+					throw new NativeQueryParseException( "Cli query should match the format db.collection.oper()" );
```
gsmet: Same here.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
```diff
@@ -102,6 +113,35 @@ public boolean setOperation(Operation operation) {
...
+	public boolean isCliQuery() {
+		return isCliQuery;
+	}
+
+	public boolean setCliQuery(boolean isCliQuery) {
+		this.isCliQuery = isCliQuery;
+		return true;
+	}
+
+	public boolean setInvalidJsonInPrefix(String invalidJsonInPrefix) {
```
gsmet: I don't understand why you called it `inPrefix`?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
```diff
@@ -142,7 +182,46 @@ public boolean setReduceFunction(String reduceFunction) {
...
+						throw new NativeQueryParseException( e );
+					}
+				}
+				// db . collection . operation ( ???
+				if ( !everyParameterIsValid ) {
+					throw new NativeQueryParseException( "Parameters are invalid for " + operationName );
+				}
+				throw new NativeQueryParseException( "Missing ')'" );
+			}
+			// the input is not a json object: might be error in { ... } or does not start with '{'
```
gsmet: I think you're missing an else here.

It shouldn't be necessary I agree as you manage all the conditions, but it's more logical to have it.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
```diff
@@ -66,7 +68,11 @@ public NativeQueryParser() {
 	}
 
 	public Rule Query() {
-		return Sequence( FirstOf( ParsedQuery(), CriteriaOnlyFindQuery() ), EOI, push( builder ) );
+		return Sequence( Optional( QueryCliOrSimple() ), push( builder ) );
```
gsmet: I'm a bit puzzled by this change, especially the `Optional`? It is to support returning an error for empty queries?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
```diff
@@ -66,7 +68,11 @@ public NativeQueryParser() {
...
+	}
+
+	public Rule QueryCliOrSimple() {
```
gsmet: Let's rename it CliQueryOrJsonFindQuery?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
```diff
@@ -79,21 +85,35 @@ public Rule ParsedQuery() {
 	 * @return the {@link Rule} to identify a find query only
 	 */
 	public Rule CriteriaOnlyFindQuery() {
-		return Sequence( ZeroOrMore( ANY ), builder.setOperation( Operation.FIND ), builder.setCriteria( match() ) );
+		return Sequence( !builder.isCliQuery(), JsonObject() , builder.setOperation( Operation.FIND ), builder.setCriteria( match() )
+		);
```
gsmet: please put the `);` on the same line as the first one!

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/parsing/nativequery/NativeQueryParserTest.java
```diff
@@ -823,5 +827,142 @@ public void shouldSupportDotInCollectionName() {
...
+			run.resultValue.build();
+			Assert.fail( "An exception was expected" );
+		}
+		catch (Exception e) {
+			assertThat( e ).isInstanceOf( NativeQueryParseException.class ).hasMessage( "Parameters are invalid for aggregate" );
+		}
+	}
+
+	@Test
+	public void shouldRecognizeMissingDBPart() { // ????
```
gsmet: Why the "// ????" ?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
```diff
@@ -30,6 +31,16 @@
 
 	private String collection;
 	private Operation operation;
+
+	/**
+	 * Helper fields for recording the states of the parser
+	 */
+	private boolean isCliQuery;
+	private String invalidJsonInPrefix;
+	private boolean everyParameterIsValid;
```
gsmet: Let's rename it areParametersValid.

gsmet:
> (about the stateless parser suggestion in the previous pull request: I think it would be easy to implement this by simply pushing the new builder to the value stack when the Query rule is invoked and every time we need the builder call the peek() method of the value stack but I'm not sure it is the best way)

I don't see any other way to do it so let's do it this way.

I created https://hibernate.atlassian.net/browse/OGM-1345 to track this one. Let's make it a separate commit on top of your changes!

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
DMasherov: Suppose the `Optional` is omitted and then we run the parser and `QueryCliOrSimple` fails. The `push( builder )` does not get executed and the value stack (and resultValue) is empty and there is no builder. 

If we swaped `push( builder )` and `QueryCliOrSimple` the value stack will be empty also because according to the [documentation](https://github.com/sirthias/parboiled/wiki/The-Value-Stack#rule-mismatches):

> If a rule did not match for whatever reason the value stack is reset to the exact state it had before the execution of the failing rule was started.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/MongoDBQueryDescriptorBuilder.java
DMasherov: I changed the definition of `JsonObject` rule to catch the rest of the input in this field if the essential part fails when the `JsonObject` rule is called. The rest of the input is not a valid JSON but I don't know what exactly is wrong - this field simply stores everything after position of a supposed JSON object. If we could "fix" this JSON part of the input it would be in the prefix part of the captured string. This was my reasoning for naming but I see it is confusing. Do you think I should rename it?

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/parsing/nativequery/NativeQueryParserTest.java
DMasherov: I just marked the line and forgot to remove it :( Actually I thought the thrown error would be confusing for a user because this query looks very much like the ordinary query but it says it's a JSON parse error. I wonder if there is some clever way to identify this but maybe I'm overthinking this.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
gsmet: Okay, I think we probably won't have an issue once the `push()` is at the beginning of everything (we will need that to make the parser stateless).

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/query/parsing/nativequery/impl/NativeQueryParser.java
```diff
@@ -339,19 +382,22 @@ public Rule JsonComposite() {
...
+			Sequence(
 				ZeroOrMore( WhiteSpace() ).skipNode(),
 				"{ ",
 				FirstOf(
 						Sequence( Pair(), ZeroOrMore( Sequence( ", ", Pair() ) ) ),
 						Optional( Pair() )
 				).suppressNode(),
 				"} "
+			),
+			Sequence( ZeroOrMore( ANY ), builder.setInvalidJsonInPrefix( match() ), ACTION( false )  )
```
gsmet: OK, somehow I missed this change when doing my first review. It's a good idea but it needs some more love.

So I think the plan would be:
 * keep JsonObject as is. It's better this way as it has a clear meaning of what it is.
 * add a rule named `JsonObjectParameter` that contains your new stuff and wraps the `JsonObject` rule and use it wherever it should be used
 * rename `invalidJsonInPrefix` to `invalidJsonParameter` (the property and the method)

DMasherov: `JsonObject` is used in `JsonComposite` and some operations rules use `JsonComposite`. Maybe it is better to wrap both this rules something like this?
```java
public Rule Insert() {
    return Sequence(
            "insert ",
            builder.setOperation( Operation.INSERT ),
            "( ",
            JsonParameter( JsonComposite() ), builder.setUpdateOrInsert( match() ),
            Optional( Sequence( ", ", JsonParameter( JsonObject() ), builder.setOptions( match() ) ) ),
            builder.setParametersValid( true ),
            ") "
    );
}

...
public Rule JsonParameter(Rule Parameter) {
    return FirstOf(
            Parameter,
            Sequence( ZeroOrMore( ANY ), builder.setInvalidJsonParameter ( match() ), ACTION( false )  )
    );
}
```

gsmet: Yes, sounds like a good idea!

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/parsing/nativequery/NativeQueryParserTest.java
gsmet: I would say let's make the error message meaningful and that should be enough.

DMasherov:
Should I create another pull request for https://hibernate.atlassian.net/browse/OGM-1345 or use this one?

gsmet:
Squashed the commits and merged. Thanks!

I added one more commit to effectively create the `NativeQueryParser` only once.

PR #911, opened 2017-11-22
OGM-1073 moving constant pool size to configuration

schernolyas:
Moving constant pool size to parameter of configuration

### DIFF
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/Neo4jProperties.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/Neo4jProperties.java
@@ -55,12 +55,12 @@ public final class Neo4jProperties implements OgmProperties {
 	public static final String ESTABLISH_CONNECTION_TIMEOUT = "hibernate.ogm.neo4j.client.establish_connection_timeout";
 
 	/**
-	 * The size of the client connection pool used by the RestEasy HTTP client
+	 * The size of the client connection pool
 	 * <p>
 	 *     Default value is 10
 	 * </p>
 	 */
-	public static final String REST_CONNECTION_POOL_SIZE = "hibernate.ogm.neo4j.client.rest.connection_pool_size";
+	public static final String CONNECTION_POOL_SIZE = "hibernate.ogm.neo4j.client.connection_pool_size";
 
 	private Neo4jProperties() {
 	}
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/Neo4jProperties.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/Neo4jProperties.java
@@ -54,6 +54,14 @@ public final class Neo4jProperties implements OgmProperties {
 	 */
 	public static final String ESTABLISH_CONNECTION_TIMEOUT = "hibernate.ogm.neo4j.client.establish_connection_timeout";
 
+	/**
+	 * The size of the client connection pool used by the RestEasy HTTP client
+	 * <p>
+	 *     Default value is 10
+	 * </p>
+	 */
+	public static final String REST_CONNECTION_POOL_SIZE = "hibernate.ogm.neo4j.client.rest.connection_pool_size";
+
 	private Neo4jProperties() {
 	}
 }
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/bolt/impl/BoltNeo4jClient.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/bolt/impl/BoltNeo4jClient.java
@@ -13,6 +13,7 @@ import org.hibernate.ogm.datastore.neo4j.remote.common.impl.RemoteNeo4jConfigura
 import org.hibernate.ogm.datastore.neo4j.remote.common.impl.RemoteNeo4jDatabaseIdentifier;
 import org.neo4j.driver.v1.AuthToken;
 import org.neo4j.driver.v1.AuthTokens;
+import org.neo4j.driver.v1.Config;
 import org.neo4j.driver.v1.Driver;
 import org.neo4j.driver.v1.GraphDatabase;
 import org.neo4j.driver.v1.exceptions.Neo4jException;
@@ -35,13 +36,15 @@ public class BoltNeo4jClient {
 
 	private Driver createNeo4jDriver(RemoteNeo4jDatabaseIdentifier identifier, RemoteNeo4jConfiguration configuration) {
 		String uri = identifier.getDatabaseUri();
+		Config.ConfigBuilder configBuilder = Config.build();
+		configBuilder.withMaxIdleSessions( configuration.getClientRestPoolSize()  );
 		try {
 			if ( configuration.isAuthenticationRequired() ) {
 				AuthToken authToken = AuthTokens.basic( configuration.getUsername(), configuration.getPassword() );
-				return GraphDatabase.driver( uri, authToken );
+				return GraphDatabase.driver( uri, authToken, configBuilder.toConfig() );
 			}
 			else {
-				return GraphDatabase.driver( uri );
+				return GraphDatabase.driver( uri, configBuilder.toConfig() );
 			}
 		}
 		catch (Neo4jException e) {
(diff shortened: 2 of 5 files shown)


neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/Neo4jProperties.java
```diff
@@ -54,6 +54,14 @@
 	 */
 	public static final String ESTABLISH_CONNECTION_TIMEOUT = "hibernate.ogm.neo4j.client.establish_connection_timeout";
 
+	/**
+	 * The size of the client connection pool used by the RestEasy HTTP client
+	 * <p>
+	 *     Default value is 10
+	 * </p>
+	 */
+	public static final String REST_CONNECTION_POOL_SIZE = "hibernate.ogm.neo4j.client.rest.connection_pool_size";
```
DavideD: I would prefer not to mention the term REST, let's just remove it. Same for all the other places

DavideD: or maybe just use http instead of rest

schernolyas:
Hi @DavideD !

I see https://stackoverflow.com/questions/39339708/neo4j-bolt-driver-connection-pooling . I will try  configure it

schernolyas:
Davide, class org.hibernate.ogm.datastore.neo4j.remote.bolt.impl.BoltNeo4jClient has incorrect order of import. I fix it too.

DavideD:
> Davide, class org.hibernate.ogm.datastore.neo4j.remote.bolt.impl.BoltNeo4jClient has incorrect order of import. I fix it too.

@schernolyas in the future could you make an effort and keep this kind of changes in separate commits, please?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/common/impl/RemoteNeo4jConfiguration.java
```diff
@@ -150,4 +158,12 @@ public Long getEstablishConnectionTimeout() {
 	public boolean isAuthenticationRequired() {
 		return authenticationRequired;
 	}
+
+	/**
+	 * @see Neo4jProperties#CONNECTION_POOL_SIZE
+	 * @return the size of REST connection pool
+	 */
+	public Integer getClientRestPoolSize() {
```
DavideD: Remove the Rest part, you can call it ClientPoolSize().

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/bolt/impl/BoltNeo4jClient.java
```diff
@@ -35,13 +38,15 @@ public BoltNeo4jClient(RemoteNeo4jDatabaseIdentifier identifier, RemoteNeo4jConf
 
 	private Driver createNeo4jDriver(RemoteNeo4jDatabaseIdentifier identifier, RemoteNeo4jConfiguration configuration) {
 		String uri = identifier.getDatabaseUri();
+		Config.ConfigBuilder configBuilder = Config.build();
+		configBuilder.withMaxIdleSessions( configuration.getClientRestPoolSize()  );
```
DavideD: Do these two attibutes have the same meaning?

schernolyas:
Okey @DavideD  I will try. For now, I will send commit for restore previous formatting and new commit for correct formatting

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/common/impl/RemoteNeo4jConfiguration.java
schernolyas: OK. thanks!

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/bolt/impl/BoltNeo4jClient.java
schernolyas: Hm ... What do you mean ? Code in row 42 is correct default configuration of Bolt driver.

DavideD:
> Okey @DavideD I will try. For now, I will send commit for restore previous formatting and new commit for correct formatting

Just to clarify, the reason I'm asking to be careful when you make this changes is that they make the commit diff larger and therefore harder to review. If, in the same commit, you add new imports and move around existing ones, it becomes harder for the reviewer to understand what's new and what was already there.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/bolt/impl/BoltNeo4jClient.java
DavideD: This line is setting the MaxIdleSessions (a time) using the value in ClientRestPoolSize ( a number of connections). I don't understand how the two values are connected.

schernolyas: Hm ... I am look to Javadoc
/**
         * The max number of idle sessions to keep open at once. Configure this
         * higher if you want more concurrent sessions, or lower if you want
         * to lower the pressure on the database instance.
         *
         * @param size the max number of idle sessions to keep open
         * @return this builder
         */
        public ConfigBuilder withMaxIdleSessions( int size )

It is not a time ... it is count of connections.

DavideD: Ah, OK. I probably checked the wrong property. But they still seems quite different to me. One is a pool size and the other one is a limit. I think it's better to keep two separates properties for this. Let's ignore bolt for this PR.

schernolyas:
Hi @DavideD !
The PR is ready for review.

DavideD:
> The PR is ready for review.

Could you remove the Bolt part? `MaxIdleSessions` is different from a connection pool.
Let's use the property only in the `HttpNeo4jClient` only.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/bolt/impl/BoltNeo4jClient.java
DavideD: This method is still called `getClientRestPoolSize()`, could you rename it to `getClientPoolSize()`

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/common/impl/RemoteNeo4jConfiguration.java
```diff
@@ -24,6 +24,11 @@
 
 	public static final int DEFAULT_BOLT_PORT = 7687;
 
+	/**
+	 * Default Size of the client connection pool used by the RestEasy HTTP client
+	 */
+	public static final int DEFAULT_REST_CONNECTION_POOL_SIZE = 10;
```
DavideD: Please, rename this to DEFAULT_CONNECTION_POOL_SIZE

schernolyas: Hi @DavideD !
Ok. 5 minutes

DavideD:
Jenkins is failing can you have a look if it's related to your PR?: http://ci.hibernate.org/job/hibernate-ogm-PR/1459/console

DavideD:
> Jenkins is failing can you have a look if it's related to your PR?: http://ci.hibernate.org/job/hibernate-ogm-PR/1459/console

It might be unrelated

schernolyas:
Hm .... May be ... I have fixed name of method.

DavideD:
It seems it was a problem in CI

schernolyas:
Hi @DavideD !
The PR is ready for review.

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/911

I squashed the commits together and removed the changes to the bolt dialect.

@schernolyas Can you add the property to the documentation and send a new PR?

Thanks

PR #912, opened 2017-11-24
OGM-1329 add cache template annotation (#24)

alaushkin:


gsmet:
add to whitelist

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/ProtobufSchemaInitializer.java
```diff
@@ -48,9 +53,21 @@ public void initializeSchema(SchemaDefinitionContext context) {
 		datastoreProvider.registerSchemaDefinitions( sd );
 	}
 
+	private String getCacheTemplate(Map tableEntityTypeMapping, String tableName) {
+		Class entity = (Class) tableEntityTypeMapping.get( tableName );
+		if ( entity == null ) {
+			return null;
+		}
+		CacheTemplate cacheTemplate = (CacheTemplate) entity.getAnnotation( CacheTemplate.class );
```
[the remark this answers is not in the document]
DavideD: Once you defined the right classes you can access the option using the following:
```
    CacheTemplate cacheTemplate = serviceRegistry.getService( OptionsService.class ).context()
        .getEntityOptions( Entity.class )
        .getUnique( CacheTemplateOption.class );
```

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/options/cache/CacheTemplate.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+package org.hibernate.ogm.datastore.infinispanremote.options.cache;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE) //on class level
```
DavideD: Is this comment necessary?

infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/utils/InfinispanRemoteTestHelper.java
```diff
@@ -65,7 +65,7 @@ public long getNumberOfAssociations(SessionFactory sessionFactory) {
 	@Override
 	public void dropSchemaAndDatabase(SessionFactory sessionFactory) {
 		final InfinispanRemoteDatastoreProvider datastoreProvider = getProvider( sessionFactory );
-		final Set<String> mappedCacheNames = datastoreProvider.getMappedCacheNames();
+		final Set<String> mappedCacheNames = datastoreProvider.getCacheTemplatesByName().keySet();
```
DavideD: Why don't you keep the other method? It seems easier and clearer.

alaushkin: fixed

alaushkin:
Sorry, I'm not understand, what exactly I should do with infinispan-server-testconfig/wildfly-trimmed-config.xml configuration.

DavideD:
> Sorry, I'm not understand, what exactly I should do with infinispan-server-testconfig/wildfly-trimmed-config.xml configuration.

Your patch does not come with a test. Meaning I don't know if it works when I use a different template name.

In my previous comment I'm trying to suggest a way to add a test. The test should work this way:

1. Start infinispan server without default cache configuration.
2. Create a testcase that use the annotation to select a template from the infinispan server configuration.
3. Create a testcase that doen't select any template or select a template that does not exist (an exception is thrown).

The cache used as default is defined here: https://github.com/hibernate/hibernate-ogm/blob/master/integrationtest/infinispan-server-testconfig/wildfly-trimmed-config.xml#L96

You will basically have to copy this file and change one or both of this parts:
```
default-cache="default"
```
In your new test, somewhere, there should be this user of the new annotation:
```
@CacheTemplate("templateForTest")
```
I don't remember right now the details for the infinispan configuration, you will need to check the documentation: http://infinispan.org/docs/8.2.x/user_guide/user_guide.html#configuring_cache

I hope this help to clarify what I meant.

DavideD:
I'm going to close this PR for now, fell free to re-open it when you have done the changes or send a new one when you are ready.

Thanks

PR #913, opened 2017-11-28
OGM-1326 Add example of associations between entities of the same class

batousik:
Added an example to the documentation that shows a OneToOne mapping to the same entity. ie. A person has a spouse that is another person.


DavideD:
How come have you closed this PR?

batousik:
I have created it by mistake - first it has to be pulled from -draft branch to -clean branch internally in EPAM.

PR #914, opened 2017-11-28
OGM-1326 Add example of associations between entities of the same class

batousik:
https://hibernate.atlassian.net/browse/OGM-1326

Added an example to the documentation that shows a OneToOne mapping to the same entity. ie. A person has a spouse that is another person.


documentation/manual/src/main/asciidoc/modules/mongodb.asciidoc
```diff
@@ -1563,6 +1563,47 @@ Wife collection as JSON in MongoDB
...
+.Bidirectional one-to-one (Mapping to the same Entity)
+====
+[source, JAVA]
+----
+@Entity
+public class Person {
+
+    @Id
+    private String id;
+    private String name;
```
DavideD: Let's keep the example simpler using the field `name` as `@Id` and removing the `field` `id`.

DavideD: On second thought adding a field role might make the example easier to understand.

documentation/manual/src/main/asciidoc/modules/mongodb.asciidoc
```diff
@@ -1563,6 +1563,47 @@ Wife collection as JSON in MongoDB
...
+Person (wife) collection as JSON in MongoDB
+
+[source, JSON]
+----
+{
+  "_id" : "bea",
+  "name" : "Bea",
+  "spouse" : "alex"
+}
+----
```
DavideD: Considering the previous comment I would rewrite this as:
```
Person  collection as JSON in MongoDB
[source, JSON]
----
{  "_id" : "Alex", role: "husband", "spouse" : "Bea" }
{  "_id" : "Bea",  role: "wife", "spouse" : "Alex" }
----
```

This make it clearer that they are part of the same collection.

DavideD: We need to also add a one-to-many association and show an example for that scenario as well.
Maybe, you can add an attribute children or similar to the same class and see if it doesn't get too confusing  to keep everything in one place. Alternatively, we can add a second example. Your choice

batousik:
I don't know how embedded infinispan stores the data, and for remote infinispan I get the following exeception: [log.gist](https://gist.github.com/batousik/db7a996e234eb3d14164edfdc0f43d84)

documentation/manual/src/main/asciidoc/modules/mongodb.asciidoc
batousik: Added children and made example simpler by removing the roles completely - spouse

batousik:
@DavideD can you please review again?

DavideD:
> don't know how embedded infinispan stores the data is there a way to create a dump?

It uses the same mapping used by remote infinispan

> And for remote infinispan I get the following exeception: log.gist

I need to know what you did to cause this exception. Did you add a new test?

> Neo4j has got more than enough examples

Is there already an example for these use cases? If No, than it seems there are not enough examples.

batousik:
> don't know how embedded infinispan stores the data is there a way to create a dump?
>> It uses the same mapping used by remote infinispan

Ill find the answer when the remote will work then :)

> And for remote infinispan I get the following exeception: log.gist
>> I need to know what you did to cause this exception. Did you add a new test?

I have tried to create the same entity as for the mongoDB and neo4j, it fails trying to process `@`Entity for the person, to reproduce you could clone this [repo](https://github.com/batousik/OGM1326) and run Runner.main

> Neo4j has got more than enough examples
>> Is there already an example for these use cases? If No, than it seems there are not enough examples.

- Sure, I will add this one in, do you want at specific example number?
- Should I include such diagram?

neo4j_family

DavideD:
Make sure that the diagram you include is similar to the ones we already have. You can see them in the documentation: https://docs.jboss.org/hibernate/ogm/5.2/reference/en-US/html_single/#_associations_3

To generate them I usually run the examples on a remote Neo4j instance, check the content of the db using the browser at http://localhost:7474 and  add a small description of the labels for each node.
If they have different labels I try to use different colors (I don't believe it applies to this example).

Try also to avoid too small arrows and make sure that they don't cross each other, it makes it harder to read it.

DavideD:
> Sure, I will add this one in, do you want at specific example number?

I don't know about that, whatever you think it's approapriate

batousik:
@DavideD should I create a new issue for the infinispan exception?

batousik:
Updated documentation for neo4j

screenshot

batousik:
@DavideD could you please reply about the infinispan.

DavideD:
> @DavideD should I create a new issue for the infinispan exception?

I'm not sure it's an error, the message seems pretty clear, it feels more like an unsupported case.
It's possible that that usecase cannot work on Infinspan. If that's the case we can skip the test for that dialect.

batousik:
@DavideD, I apologize for this long discussion, the final commit then will contain documentation examples for the:

- neo4j
- mongodb

Questions:

1. What would you like me to do about the unsupported issue with the infinispan?
2. Do you want the test cases that I have created for the new documentation to be included in the commit, if so, what would be the appropriate place for them?

batousik:
@DavideD Requested changes are done, build didn't complete due to an authorization error

DavideD:
Thanks, I will check it in a bit

schernolyas:
HI @DavideD !

What is status of the PR ?

fax4ever:
Merged thanks! Nice example for the doc

PR #916, opened 2017-11-29
OGM-359 Stored procedure support

DavideD:
@schernolyas sent a patch but there were some things I wasn't convinced about and  I ended up doing some refactoring.

This PR will add support for stored procedure, In particular  for MongoDB only, but I think it's good enough for now. As a follow up we will need to add support it for Neo4j (@schernolyas can you do it?).

I changed several things, @schernolyas and @gsmet  can you have a look at it? I think my changes follow more the ORM approach, but I'm open to suggestion.

Thanks


core/src/main/java/org/hibernate/ogm/storedprocedure/ProcedureQueryParameters.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Named or positional parameters to pass to a stored procedure.
+ *
+ * @author Davide D'Alto
+ */
+public class ProcedureQueryParameters {
```
schernolyas: Why did You not use org.hibernate.ogm.dialect.query.spi.QueryParameters ? What is reason for create new class that looks like org.hibernate.ogm.dialect.query.spi.QueryParameters?

schernolyas:
Hi @DavideD !

I reviewed your version. All right except one question.
Question is about support the feature in Neo4j. Yes, I can. No problems with this.  But  I have a question about time. I mean that I can do it by self (it will fast) or I can pass it to EPAM developers (it is cool task for them but ... it will slow). What variant do you prefer ?

core/src/main/java/org/hibernate/ogm/storedprocedure/ProcedureQueryParameters.java
DavideD: I didn't want to deal with the TypeGridValue, I'm not sure if it make sense when it comes with stored procedures.

schernolyas: Hm ... What is problem with TypeGridValue ?

DavideD: Add complexity that I'm not sure we need

schernolyas: Okey. May be ... you are right. Anyway, we will have possibility return to using TypeGridValue .

DavideD: Yes, as I said. I'm not completely sure about this solution. But I prefer to start with something simple and add the missing bit later

schernolyas: I see that method "executeBackendQuery" uses org.hibernate.ogm.dialect.query.spi.QueryParameters . I think it is not logical. Queries use one class for parameters but stored procedures use other class for parameters. What do you think about it?

DavideD: Why not? calling a query is different from calling a stored procedure, in fact we use two different methods to execute them. What's so surprising if they also use 2 different classes for the parameters? By the way, parameters for stored procedures are a a bit different than parameters for queries, because you also have to consider if they are IN, OUT, etc.

DavideD:
> What variant do you prefer ?

Feel free to pick the one you prefer

schernolyas:
Hi @DavideD !

Now I am developing OGM-167.  It is complex task. I will prepare task for EPAM developers. They will grad :) . I will create subtask in OGM-359 for support stored procedures in Neo4j.

core/src/main/java/org/hibernate/ogm/storedprocedure/ProcedureQueryParameters.java
schernolyas: okey.  Time will tell .

schernolyas:
Hi @gsmet !
What do you think about the PR?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -25,6 +25,8 @@
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 
+import org.apache.commons.lang3.StringEscapeUtils;
```
gsmet: OK, now that both of you have converged, I would vote for suppressing the dependency to commons-lang3. It's a big dependency for a single method.

We should be able to reimplement it.

core/src/main/java/org/hibernate/ogm/jpa/impl/OgmStoredProcedureQuery.java
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	private static HibernateEntityManagerImplementor convert(EntityManager em) {
+		return em.unwrap( HibernateEntityManagerImplementor.class );
+	}
+
+	@Override
+	public List getResultList() {
+		try {
+			final NoSQLProcedureResultSetOutputImpl rtn = (NoSQLProcedureResultSetOutputImpl) outputs().getCurrent();
```
gsmet: Do we really need this class? Apart from this cast, which is quite artificial and doesn't look like required, it seems to be very generic?

Is there something specific forcing us to not use an ORM class?

core/src/main/java/org/hibernate/ogm/storedprocedure/ProcedureQueryParameters.java
gsmet: So I have a question similar to what Sergey asked but a bit different: do we need a class specific to OGM? Can't we reuse ORM's mechanisms? (I didn't check so it's a real question)

A big part of my work on the ORM 5.2 port was to try to reduce to the minimum the number of classes reimplemented in OGM so I'm wondering if maybe we could be a bit more "ORM integrated"?

I have the same interrogation for a few other classes.

I haven't digged if it was possible or not so this question is more to share a concern I have. Might be totally misplaced.

DavideD: This PR was sent when we were still on ORM 5.1

I'm pretty sure at the time I couldn't reuse existing classes. I'm rebasing this PR now.
I will check what I can remove.

schernolyas:
Hi @gscheibel !
I got many classes from ORM and I removed code that linked with JDBC. I tried reuse classes from ORM ... but it is impossible :(.  If you see the reusing .... let me know

core/src/main/java/org/hibernate/ogm/storedprocedure/ProcedureQueryParameters.java
gsmet: Yeah, as discussed with Sergey on HipChat, not sure we can do better, just wondering :).

DavideD:
@schernolyas Do you have time to rebase this PR to the latest changes

schernolyas:
Hi @DavideD !
Ok. I try to upgrade the PR.

schernolyas:
HI @DavideD .
Please, see PR https://github.com/DavideD/hibernate-ogm/pull/29

DavideD:
See https://github.com/hibernate/hibernate-ogm/pull/916

DavideD:
Closing this one, @schernolyas sent an update

PR #917, opened 2017-12-10
[PREVIEW] OGM-167 reading/writing tuple directly

schernolyas:
It is preview PR. 
I corrected for CRUDTest working.


DavideD:
@schernolyas What about this PR? It seems the approach is correct. Are you going to continue the work with it?

schernolyas:
HI @DavideD !
Yes, I am.
But main complexity in class https://github.com/schernolyas/hibernate-ogm/blob/OGM-167_reading_bson_directly_dev2/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/codec/TupleCodec.java . Problem in method "decode" because ... I don't know what I read.

It is complex PR.

DavideD:
@schernolyas  I don't have a solution right  now for this, I will close this PR because I don't have the time to dig into it. Feel free to explore a bit more if you want to. Otherwise, make sure to describe the obstacle you found on the JIRA and we leave it open so that we can try again in the future. Thanks a lot

PR #919, opened 2017-12-11
OGM-1073 document new parameter

schernolyas:


DavideD:
Merged with some changes https://github.com/hibernate/hibernate-ogm/pull/919

PR #929, opened 2017-12-22
OGM-1361 Build modules for WildFly 11

gsmet:
* https://hibernate.atlassian.net/browse/OGM-1361
 * https://hibernate.atlassian.net/browse/OGM-1369

This is a partial backport of my work supporting ORM 5.2 (for which upgrading to WF 11 was necessary).

OGM 5.2 modules will be released as WF 11 modules. Note that, contrary to the other versions, it uses the main slot for ORM as OGM 5.2 uses ORM 5.1 and it is the version bundled with WF 11.

I also upgraded Search to 5.6.4.Final.

### DIFF
--- a/integrationtest/pom.xml
+++ b/integrationtest/pom.xml
@@ -11,7 +11,6 @@
     <description>Hibernate OGM integration and performance tests</description>
 
     <properties>
-        <version.org.wildfly.arquillian>1.0.2.Final</version.org.wildfly.arquillian>
         <jboss.home>${project.build.directory}/wildfly-${version.wildfly}</jboss.home>
         <infinispan-server.home>${project.build.directory}/infinispan-server-${infinispanVersion}</infinispan-server.home>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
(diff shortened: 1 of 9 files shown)


Sanne:
Looks good, but why override the Search version (and force different modules to be used) when there is a compatible build included in WildFly?

gsmet:
Well, I wasn't sure it would work and 5.5 would be totally compatible with 5.6.

Moreover, we provided 5.6 before so it feels a bit weird to go back to 5.5 (and remove the Elasticsearch support).

integrationtest/pom.xml
```diff
@@ -160,7 +150,12 @@
                                         <groupId>org.hibernate</groupId>
                                         <artifactId>hibernate-search-modules</artifactId>
                                         <version>${hibernateSearchVersion}</version>
-                                        <classifier>${hibernateWildflyClassifier}</classifier>
+                                        <!--
+                                        We don't have a wildfly-11-dist version of the Hibernate Search 5.6 modules.
+                                        Using the ones built for WildFly 10 will do so we hardcode the classifier here
+                                        for now.
+                                        -->
+                                        <classifier>wildfly-10-dist</classifier>
```
DavideD: I'm a bit concerned that we might forget to change this in the future, could you use a separate property and add it close toHibernateSearchVersion or version.wildfly or hibernateWildflyClassifier, so that when we upgrade the libraries we remember to have a look at this? Does it make sense?

Sanne:
> we provided 5.6 before so it feels a bit weird to go back to 5.5

+1

integrationtest/pom.xml
gsmet: We won't as I will rebase my ORM 5.2 PR right away and we won't have this weird thing there.

That's why I didn't bother about adding a property.

DavideD: ok

DavideD:
@gsmet Looks good, feel free to merge it

PR #930, opened 2017-12-27
 OGM-1353 Infinispan sequence generation is not thread safe

fax4ever:


gsmet:
add to whitelist

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/persistencestrategy/counter/ClusteredCounterCommand.java
```diff
@@ -0,0 +1,65 @@
...
+
+			// try to define new counter at runtime
+			boolean definedByCurrentThread = counterManager.defineCounter( counterName,
+				CounterConfiguration.builder( CounterType.UNBOUNDED_STRONG )
+					.initialValue( request.getInitialValue() )
+					.storage( Storage.VOLATILE )
+					.build() );
+
+			if ( definedByCurrentThread ) {
+				return new Long( request.getInitialValue() );
```
DavideD: Wouldn't be better `Long.valueOf(...)`?

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/persistencestrategy/counter/ClusteredCounterCommand.java
```diff
@@ -0,0 +1,65 @@
...
+
+		}
+
+		StrongCounter strongCounter = counterManager.getStrongCounter( counterName );
+		try {
+			// consistent "single unit" increment && get
+			return strongCounter.addAndGet( request.getIncrement() ).get();
+		}
+		catch (ExecutionException | InterruptedException e) {
+			throw new HibernateException( "Interrupting Operation " + e.getMessage(), e );
```
DavideD: I think this exception is a bit too generic, maybe we can add some information about the fact that it happened while incrementing the sequence and add the CounterName as well (at least).
It should probably includeded in the Log interface as well

infinispan-embedded/src/test/java/org/hibernate/ogm/datastore/infinispan/test/dialect/impl/clustered/ClusteredSequenceNextValueGenerationTest.java
```diff
@@ -0,0 +1,45 @@
...
+public class ClusteredSequenceNextValueGenerationTest extends SequenceNextValueGenerationTest {
+
+	@Override
+	protected void configure(GetterPersistenceUnitInfo info) {
+		info.getProperties().setProperty( InfinispanProperties.CONFIGURATION_RESOURCE_NAME, "infinispan-dist.xml" );
+	}
+
+	/**
+	 * This test is skipped for Infinispan dialect on base Class.
+	 * Using clustered counters on top of Infinispan clustered configuration can be enabled.
```
DavideD: I think I would keep only the first line of this javadoc.

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/InfinispanDialect.java
```diff
@@ -296,4 +307,16 @@ public Tuple next() {
 		public void close() {
 		}
 	}
+
+	/**
+	 * With the introduction of Infinispan clustered counter
+	 * SequenceGenerator strategy is now supported
+	 *
+	 * @return true
+	 */
```
DavideD: To be fair I wouldn't add this javadoc, not big deal though

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/InfinispanDialect.java
```diff
@@ -192,6 +195,14 @@ public boolean isStoredInEntityStructure(AssociationKeyMetadata associationKeyMe
 	@Override
 	//TODO should we use GridTypes here?
 	public Number nextValue(NextValueRequest request) {
+
+		// if transport is defined on Infinispan cache manager
+		// we can use clustered counter to generate next value
+		EmbeddedCacheManager cacheManager = getCacheManager().getCacheManager();
```
DavideD: What happens if the transport is not defined? Is there an exception?
If not, would it make sense to add one? This way we can tell the user that something is missing if he wants to generate values

DavideD:
@fax4ever Could you also add a small note in the documentation about the configuration needed to make sequences work, please? Thanks

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/InfinispanDialect.java
fax4ever: At the moment I left the fallback to default behavior. But as you wish, in fact it could make sense to make the transport mandatory in case of sequence use. Do we throw an exception? At that point we could also remove the cache of the sequences that we will not use anymore. But before doing it I wanted your feedback.

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/persistencestrategy/counter/ClusteredCounterCommand.java
fax4ever: absolutely

documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
```diff
@@ -512,6 +515,14 @@ Start with the default configuration and see if that fits you.
 You can fine tune these setting when you are closer to going in production.
 ====
 
+====
+
+In order to enable `Clustered Counter` support we have defined a `counters` tag.
+Cluster counter support it is used to implement a reliable and consist stategy to generate unique source ID value.
```
gsmet: Just a few typos noted:
s/it//
s/consist/consistent/
s/value/values/ ?

documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
```diff
@@ -512,6 +515,14 @@ Start with the default configuration and see if that fits you.
...
+Cluster counter configuration require jGroups tranport defined configuration.
```
gsmet: s/require/requires/
s/jGroups/JGroups/

Not sure of what you want to say here. Maybe "requires the definition of a JGroups tranport configuration." would be better?

HTH

Sanne:
Hi @fax4ever ! I'll do a quick review too.

Some suggestions for the git process we use:
 - keep your work in a branch with a good name. For example `OGM-1353` (not `master`)
 - when OGM/master gets other updates merged before yours, you'll have to update your own: don't use _merge_ in this case but: 1) update your own `master` to match `hibernate/master` 2)rebase your work on your master 

See also:
 - http://hibernate.org/orm/contribute/
 - https://github.com/hibernate/hibernate-orm/blob/master/CONTRIBUTING.md

sorry, we'll be particularly picky at the beginning ;)

Sanne:
Good pattern:
    
    OGM-1353 Infinispan sequence generation is not thread safe

Bad pattern:

    OGM-1353: Fixing Wildfly Distribution Module for Clustered Count

(we don't use the "**:**" )

Also, the author name is different. Some commits use "Fabio Massimo Ercoli", others use "fax4ever". It's not a problem, but maybe you prefer to choose one and stick to it?

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/InfinispanDialect.java
Sanne: The problem is that the previous code is not correct (I thought it was but I was proved wrong).

I was hoping that we'd now be able to use the `CounterManager` to entirely replace (and fix) the current approach.

documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
Sanne: s/tranport/transport/

Sanne: > Cluster counter support it is used to implement

I think some quotes are missing? Currently I can't understand this.

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/InfinispanDialect.java
Sanne: Also for performance reasons could you avoid extracting the `EmbeddedCacheManager` each time a `nextValue` is needed?

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/persistencestrategy/counter/ClusteredCounterCommand.java
```diff
@@ -0,0 +1,65 @@
...
+import org.infinispan.manager.EmbeddedCacheManager;
+
+/**
+ * @author Fabio Massimo Ercoli (C) 2017 Red Hat Inc.
+ */
+public class ClusteredCounterCommand {
+
+	public Number nextValue(EmbeddedCacheManager cacheManager, NextValueRequest request) {
+
+		CounterManager counterManager = EmbeddedCounterManagerFactory.asCounterManager( cacheManager );
```
Sanne: Could you extract all the "preparation" code so that it's run only once?

Then we can reuse the `ClusteredCounterCommand` for a specific counter. Make sure to have one of these created upfront for each sequence source we need.

infinispan-embedded/src/test/resources/infinispan-dist.xml
```diff
@@ -79,6 +84,9 @@
                 await-initial-transfer="true" />
         </distributed-cache>
 
+        <!-- Clustered Counters are defined at runtime by InfinispanDialect -->
+        <counters xmlns="urn:infinispan:config:counters:9.1" num-owners="1" reliability="CONSISTENT" />
```
Sanne: Let's not suggest to users a "num-owners=1", that's not very responsible ;)
Make it two in the documentation at least?

+1 to keep it at one in the integration tests.

Sanne: Let me clarify. I mean we need to add this line in the examples and make sure the documentation mentions it's needed.

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/InfinispanDialect.java
fax4ever: absolutely

infinispan-embedded/src/test/resources/infinispan-dist.xml
fax4ever: I completely agree

DavideD:
@fax4ever doesn't have the time right now to work on this. I'll apply the remarks on top of his work and send a new PR

PR #932, opened 2018-01-10
OGM-1371 MongoDB native parser does not recognize 0 as integer

DavideD:
https://hibernate.atlassian.net/browse/OGM-1371

A quick one found by a user: https://stackoverflow.com/questions/48056260/hibernate-ogm-with-mongodb-count-query-with-gt-gte-lt-lte-not-working-with-l


Sanne:
merged, thanks!

PR #933, opened 2018-01-11
OGM-1372 Put the distribution files to upload into distribution/target/dir so that release scripts find them

yrodiere:
https://hibernate.atlassian.net//browse/OGM-1372

See also hibernate/hibernate-noorm-release-scripts#4 for more information about how this `dist` directory will be uploaded.

Note this also fixes an issue: the WildFly module would not be uploaded to Sourceforge without this change, because it doesn't match the pattern in the release script (`*-wildfly-10-*` vs. `wildfly-11`).


Sanne:
merged! Thanks @yrodiere

PR #934, opened 2018-01-11
OGM-1357 Build failing consistently on Travis

DavideD:
https://hibernate.atlassian.net/browse/OGM-1357

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,9 +37,9 @@ script:
   # we run checkstyle first to fail fast if there is a styling error
   - ./mvnw $BUILD_OPTIONS -DskipDistro=true checkstyle:check
     && BUILD_OPTIONS+=' -Dcheckstyle.skip'
-    && ./mvnw $BUILD_OPTIONS clean install
-    && ./mvnw $BUILD_OPTIONS -pl core,neo4j -Pneo4j-http clean install
-    && NEO4J_PORT=7687 ./mvnw $BUILD_OPTIONS -pl core,neo4j -Pneo4j-bolt clean install
+    && ./mvnw $BUILD_OPTIONS help:active-profiles clean install
+    && ./mvnw $BUILD_OPTIONS -pl core,neo4j -Pneo4j-http help:active-profiles clean install
+    && NEO4J_PORT=7687 ./mvnw $BUILD_OPTIONS -pl core,neo4j -Pneo4j-bolt help:active-profiles clean install
 before_cache:
   # Do not put Hibernate OGM artifacts into the cache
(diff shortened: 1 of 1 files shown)


DavideD:
It seems I talked too soon

DavideD:
It works now

.travis.yml
```diff
@@ -36,9 +37,10 @@ script:
   # we run checkstyle first to fail fast if there is a styling error
   - ./mvnw $BUILD_OPTIONS -DskipDistro=true checkstyle:check
     && BUILD_OPTIONS+=' -Dcheckstyle.skip'
-    && ./mvnw $BUILD_OPTIONS clean install
-    && ./mvnw $BUILD_OPTIONS -pl neo4j -Pneo4j-http clean install
-    && NEO4J_PORT=7687 ./mvnw $BUILD_OPTIONS -pl neo4j -Pneo4j-bolt clean install
+    && ./mvnw $BUILD_OPTIONS help:active-profiles clean install
+    && ./mvnw $BUILD_OPTIONS -pl core,neo4j -Pneo4j-http help:active-profiles clean install
+    && NEO4J_PORT=7687
```
yrodiere: As far as I can tell this statement doesn't do anything. The variable will not be exported automatically as an environment variable to the process launched in the following line.
Just thought I'd warn you...

gsmet: Yeah, I had the same comment when I read that last night. There's something wrong with this change.

And if we export it, I don't see how it would change anything to the fact that the Bolt tests wouldn't be run if the HTTP tests fail. I think it's already guaranteed. Or am I missing something obvious?

yrodiere: > I think it's already guaranteed

It is. But maybe there's a confusion, because I introduced this guarantee "while I was at it", when working on another ticket. Maybe Davide didn't notice this change?

In any case, I don't see how not running the Bolt tests when the HTTP tests fail would solve OGM-1357 ("Build failing consistently on Travis"). There is still some work required to solve the test failures.

DavideD: I'm pretty sure I've introduced this change because the test failures were hidden by the log of the build using the Bolt protocol. But checking the build history it seems I dreamed it.

Anyway, I removed the commit

DavideD:
Clsoing this PR because now that I removed the commit the build doesn't work anymore...

yrodiere:
@DavideD Are you sure the failure isn't transient? It might very well be unrelated to this commit...

DavideD:
It's always a failure of the same two tests. I don't think the commit is the problem though

DavideD:
I need to check anyway

PR #935, opened 2018-01-11
[PREVIEW] OGM-1047 using single instance of embedded Neo4j

schernolyas:


neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
```diff
@@ -24,13 +28,15 @@
  * @author Davide D'Alto &lt;davide@hibernate.org&gt;
  */
 public class EmbeddedNeo4jGraphDatabaseFactory implements GraphDatabaseServiceFactory {
+	private static GraphDatabaseService graphDatabaseService;
```
DavideD: I don't think this can be a static variable. What if I want to create 2 factories

schernolyas: It is a problem. One instance of embedded Neo4J uses one path of database.

Sanne: @schernolyas Davide means that people might want to create two `different` Hibernate SessionFactory instances which point to different databases. This is a commond requirement so we can't use static variables for such things.

schernolyas: Good morning!

Okey @Sanne . I understood. I will support many storage paths. Current implementation supports many SessionFactories instances which point to one storage path.

DavideD: Are you sure? People can create factories in different ways and can specify `hibernate.ogm.neo4j.database_path` from a properties file or programmatically.

Even f this feature is not supported now,  it doesn't seem a good idea having static variables in this class.

schernolyas: Yes. I am sure. Because tests that use many factories works. You can see exception in previous build of the feature. 
Now we have other question. How we can support many database paths? I will think.

DavideD:
Closing this PR now, the static variable is not a viable approach

PR #937, opened 2018-01-14
[PREVIEW] OGM-1047 clear embedded neo4j db

schernolyas:
The PR contains migration to Java 8 and support many database paths

### DIFF
--- a/neo4j/pom.xml
+++ b/neo4j/pom.xml
@@ -18,6 +18,9 @@
 
     <name>Hibernate OGM for Neo4j</name>
     <description>Persist objects in Neo4j</description>
+    <properties>
+        <jdkTargetForProcessor>1.8</jdkTargetForProcessor>
+    </properties>
 
     <build>
         <testResources>
@@ -41,6 +44,25 @@
             </plugin>
             <plugin>
                  <artifactId>maven-checkstyle-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>animal-sniffer-maven-plugin</artifactId>
+                <dependencies>
+                    <dependency>
+                        <groupId>org.ow2.asm</groupId>
+                        <artifactId>asm-all</artifactId>
+                        <version>5.0.3</version>
+                        <scope>compile</scope>
+                    </dependency>
+                </dependencies>
+                <configuration>
+                    <signature>
+                        <groupId>org.codehaus.mojo.signature</groupId>
+                        <artifactId>java18</artifactId>
+                        <version>1.0</version>
+                    </signature>
+                </configuration>
             </plugin>
         </plugins>
     </build>
@@ -170,6 +192,7 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
 
     <profiles>
         <!-- Executes the tests against an embedded Neo4j -->
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
@@ -81,6 +81,7 @@ public class EmbeddedNeo4jGraphDatabaseFactory implements GraphDatabaseServiceFa
 			service =  builder.newGraphDatabase();
 		}
 		catch (Exception e) {
+			LOG.error( "Cannot create new GraphDatabaseService!",e );
 			throw LOG.cannotCreateNewGraphDatabaseServiceException( e );
 		}
 		finally {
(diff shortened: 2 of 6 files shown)


Sanne:
Hi @schernolyas ! 
 Can you explain how the `update to Java 8` commit relates to the OGM-1047 issue? Should be it a separate issue and separate PR?

About the approach of emptying the database: good idea! It's still causing a test to fail, but I trust you'll fix that before sending the non-preview PR. At an high level, looks good! Please go ahead and make it work properly.

schernolyas:
Hi @Sanne !
About Java 8. Yes, I can.  Nohow.  About separate task ... it is good idea. But it must be for all dialects, except dialects, that cann't work with JDK 8.

About failed test ... Yes ... I will investigate. Please, don't close the PR :)

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
```diff
@@ -24,13 +29,16 @@
  * @author Davide D'Alto &lt;davide@hibernate.org&gt;
  */
 public class EmbeddedNeo4jGraphDatabaseFactory implements GraphDatabaseServiceFactory {
+	private static final Map<String, GraphDatabaseService> GRAPH_DATABASE_SERVICE_MAP = new ConcurrentHashMap<>();
```
DavideD: I  don't understand the purpose of this.

Can you create a separate test and add it to the PR so that I can see which use case you are trying to solve?
This is the approach we work with usually, create a test first and then try to fix the problem.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
```diff
@@ -50,6 +58,18 @@ public void initialize(Map<?, ?> properties) {
 
 	@Override
 	public GraphDatabaseService create() {
+		LOG.infof( "1. Try  create new service instance for dbPath  %s", dbLocation );
+		final String dbLocationAbsolutePath = dbLocation.getAbsolutePath();
+		LOG.infof( "2. Try  create new service instance for dbPath  %s", dbLocationAbsolutePath );
+		return GRAPH_DATABASE_SERVICE_MAP.computeIfAbsent( dbLocationAbsolutePath, ( String path ) -> {
+			LOG.infof( " Create new service instance for dbPath  %s", dbLocationAbsolutePath );
+			return buildGraphDatabaseService();
```
DavideD: Why do you need the same log 3 times? Can't you just print the absolutePath? I would prefer not to add a lot of redundant information in the output of the log,

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
```diff
@@ -60,11 +80,16 @@ public GraphDatabaseService create() {
 			//Neo4J relies on the context classloader to load its own extensions:
 			//Allow it to load even the ones we're not exposing directly to end users.
 			currentThread.setContextClassLoader( neo4JClassLoader );
-			return builder.newGraphDatabase();
+			service =  builder.newGraphDatabase();
+		}
+		catch (Exception e) {
+			LOG.error( "Cannot create new GraphDatabaseService!",e );
+			throw LOG.cannotCreateNewGraphDatabaseServiceException( e );
```
DavideD: Why do you need the Log.error when there is already an exception?

neo4j/pom.xml
```diff
@@ -42,6 +45,25 @@
             <plugin>
                  <artifactId>maven-checkstyle-plugin</artifactId>
             </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>animal-sniffer-maven-plugin</artifactId>
```
DavideD: Animal sniffer is already configured in the root pom, why do you need it here as well?

neo4j/pom.xml
```diff
@@ -18,6 +18,9 @@
 
     <name>Hibernate OGM for Neo4j</name>
     <description>Persist objects in Neo4j</description>
+    <properties>
+        <jdkTargetForProcessor>1.8</jdkTargetForProcessor>
+    </properties>
```
DavideD: How come you need this here?

DavideD: By the way, we already have  a property called minJdkVersion set t java 1.8, so what is the purpose of this commit?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
schernolyas: Hi @DavideD !
Main idea is support many database instances. I reminder you that one instance of embedded neo4j uses one path ONLY. In this way, Hibernate MUST manage this specialty.

About separate test. Okey. It is good and right way.

PR #938, opened 2018-01-15
Infinispan server provisioning

Sanne:
This is to make it possible for us to consume feature packs soon, as e.g. Hibernate Search & Infinispan will start producing them.

Also happens to make our testsuite a bit lighter.

 - https://hibernate.atlassian.net/browse/OGM-1373
 - https://hibernate.atlassian.net/browse/OGM-1374

Later we should have OGM also produce "feature packs", and this will be useful to run integration tests on our own built packs.


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/938

DavideD:
Thanks

PR #939, opened 2018-01-15
[FIX_NPE]OGM-1047 Clear embedded neo4j db 

schernolyas:
The PR contains improvement that allow start tests faster. 
Reason of it is reusing exist instance of database by clear all data from previous test.

PLEASE!!! DON'T CLOSE THE PR ! I CAN INVESTIGATE A PROBLEM WITH NPE BY THE CI ONLY.

### DIFF
--- a/neo4j/pom.xml
+++ b/neo4j/pom.xml
@@ -34,6 +34,12 @@
             <plugin>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
+                    <!--
+                       see https://www.igorkromin.net/index.php/2017/08/17/how-to-enable-the-full-stack-trace-in-mavens-surefire-plugin-for-junit-testing/
+                       and
+                       http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#trimStackTrace
+                        -->
+                    <trimStackTrace>false</trimStackTrace>
                     <dependenciesToScan>
                         <dependency>org.hibernate.ogm:hibernate-ogm-core</dependency>
                     </dependenciesToScan>
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
@@ -7,13 +7,17 @@
 package org.hibernate.ogm.datastore.neo4j.embedded.impl;
 
 import java.io.File;
+import java.lang.invoke.MethodHandles;
 import java.net.URL;
 import java.util.HashMap;
 import java.util.Map;
 
 import org.hibernate.ogm.datastore.neo4j.Neo4jProperties;
+import org.hibernate.ogm.datastore.neo4j.logging.impl.Log;
+import org.hibernate.ogm.datastore.neo4j.logging.impl.LoggerFactory;
 import org.hibernate.ogm.datastore.neo4j.spi.GraphDatabaseServiceFactory;
 import org.hibernate.ogm.util.configurationreader.spi.ConfigurationPropertyReader;
+
 import org.neo4j.graphdb.GraphDatabaseService;
 import org.neo4j.graphdb.factory.GraphDatabaseBuilder;
 import org.neo4j.graphdb.factory.GraphDatabaseFactory;
@@ -24,6 +28,8 @@ import org.neo4j.graphdb.factory.GraphDatabaseFactory;
  * @author Davide D'Alto &lt;davide@hibernate.org&gt;
  */
 public class EmbeddedNeo4jGraphDatabaseFactory implements GraphDatabaseServiceFactory {
+	private static final Map<String, GraphDatabaseService> GRAPH_DATABASE_SERVICE_MAP = new HashMap<>();
+	private static Log LOG = LoggerFactory.make( MethodHandles.lookup() );
 
 	private File dbLocation;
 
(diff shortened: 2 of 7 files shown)


Sanne:
> DON'T CLOSE THE PR ! I CAN INVESTIGATE A PROBLEM WITH NPE BY THE CI ONLY.

Ok but I need to mute notifications. When you're done with experiments, please close it: you'll have to send a new PR as we won't see any more updates about this.

neo4j/pom.xml
```diff
@@ -34,6 +34,12 @@
             <plugin>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
+                    <!--
+                       see https://www.igorkromin.net/index.php/2017/08/17/how-to-enable-the-full-stack-trace-in-mavens-surefire-plugin-for-junit-testing/
+                       and
+                       http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#trimStackTrace
+                        -->
+                    <trimStackTrace>false</trimStackTrace>
```
DavideD: Please, add a small description and not only the link.

The website can change and we will never know what was pointing to.

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/EmbeddedGraphDatabaseFactoryTest.java
```diff
@@ -68,6 +66,51 @@ public void testLoadMalformedPropertiesLocation() throws Exception {
...
+		EmbeddedNeo4jGraphDatabaseFactory factory2 = new EmbeddedNeo4jGraphDatabaseFactory();
+		factory2.initialize( properties );
+		GraphDatabaseService graphDatabaseService2 = factory2.create();
+
+		assertTrue( "Two instances must be same!", graphDatabaseService1 == graphDatabaseService2 );
+
+		graphDatabaseService1.shutdown();
+		graphDatabaseService2.shutdown();
+	}
+
```
DavideD: When you mention that it was not possible to have different paths I didn't understand you meant two factories using the same folder for the same db.

Using the same folder in Neo4j embedded is not possible and the fact the there is an error when one tries to is correct. Also in this case if I want to use two different set of properties but the path is the same only the irst set of properties is accepted.

And what happens if I want to shutdown a factory but not the other?

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/EmbeddedGraphDatabaseFactoryTest.java
```diff
@@ -68,6 +66,51 @@ public void testLoadMalformedPropertiesLocation() throws Exception {
...
+		Properties properties = new Properties();
+		properties.put( Neo4jProperties.DATABASE_PATH, dbLocation );
+		properties.put( Neo4jProperties.CONFIGURATION_RESOURCE_NAME, neo4jPropertiesUrl().toExternalForm() );
+		factory1.initialize( properties );
+		GraphDatabaseService graphDatabaseService1 = factory1.create();
+
+		//create second factory
+		EmbeddedNeo4jGraphDatabaseFactory factory2 = new EmbeddedNeo4jGraphDatabaseFactory();
+		factory2.initialize( properties );
+		GraphDatabaseService graphDatabaseService2 = factory2.create();
```
DavideD: When you mention that it wasn't working with multiple path I thought you meant when using different folders. If you are using the same folder is supposed to give you errors because Neo4j acquire a lock on it and it won't let you change the database. It's not a use case we need to solve.

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/EmbeddedGraphDatabaseFactoryTest.java
```diff
@@ -68,6 +66,51 @@ public void testLoadMalformedPropertiesLocation() throws Exception {
...
+
+		assertTrue( "Two instances must be same!", graphDatabaseService1 == graphDatabaseService2 );
+
+		graphDatabaseService1.shutdown();
+		graphDatabaseService2.shutdown();
+	}
+
+	@Test
+	public void testInstantiationTwoFactoriesForTwoPaths() {
```
DavideD: This test is nice and if you apply it on master you will notice that it passes without additional changes

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
```diff
@@ -86,4 +116,17 @@ private void setConfigurationFromLocation(GraphDatabaseBuilder builder, URL cfgL
...
+	static void shutdownGraphDatabaseService(GraphDatabaseService neo4jDb) {
+		LOG.info( " Shutdown service instance" );
+		String key = null;
+		for ( String currentKey : GRAPH_DATABASE_SERVICE_MAP.keySet() ) {
+			if ( GRAPH_DATABASE_SERVICE_MAP.get( currentKey ) == neo4jDb ) {
+				key = currentKey;
+				break;
+			}
+		}
+		GRAPH_DATABASE_SERVICE_MAP.remove( key );
```
DavideD: Do you realize that you going to close the db for every sessionFactory?

DavideD:
> PLEASE!!! DON'T CLOSE THE PR ! I CAN INVESTIGATE A PROBLEM WITH NPE BY THE CI ONLY.

Asking for something using uppercase is not polite and it's the equivalent of screaming. Please, next time use lowecase.

neo4j/pom.xml
DavideD: I also would love to have an idea of the change without having to surf the web

schernolyas: Hi @DavideD .
I can read about the option at http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#trimStackTrace  and https://www.igorkromin.net/index.php/2017/08/17/how-to-enable-the-full-stack-trace-in-mavens-surefire-plugin-for-junit-testing/

schernolyas:
PLEASE!!! DON'T CLOSE THE PR ! I CAN INVESTIGATE A PROBLEM WITH NPE BY THE CI ONLY.

Asking for something using uppercase is not polite and it's the equivalent of screaming. Please, next time use lowercase.

Please, excuse me.

neo4j/pom.xml
schernolyas: Maven plugin for junit cut stack traces. And it is not comfortable. The option configure the plugin for not cut stacktrces. It is very useful option

schernolyas:
I reproduced the exception locally.
Sorry for spam.

PR #940, opened 2018-01-19
[PREVIEW] OGM-1047 Using database pool

schernolyas:
Main idea of the PR is using embedded database pool. In other words,
only one database is for one database path. Application can have many
database factories that use one database path. In this way, these
factories will use only one database.

### DIFF
--- a/neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/EmbeddedGraphDatabaseFactoryTest.java
+++ b/neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/EmbeddedGraphDatabaseFactoryTest.java
@@ -6,18 +6,21 @@
  */
 package org.hibernate.ogm.datastore.neo4j.test;
 
-import java.io.File;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
 import java.net.URL;
 import java.util.Properties;
 
-import org.fest.util.Files;
 import org.hibernate.HibernateException;
 import org.hibernate.ogm.datastore.neo4j.Neo4jProperties;
 import org.hibernate.ogm.datastore.neo4j.embedded.impl.EmbeddedNeo4jGraphDatabaseFactory;
 import org.hibernate.ogm.datastore.neo4j.utils.EmbeddedNeo4jTestHelperDelegate;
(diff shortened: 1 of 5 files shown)


neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/EmbeddedGraphDatabaseFactoryTest.java
```diff
@@ -68,6 +66,51 @@ public void testLoadMalformedPropertiesLocation() throws Exception {
...
+
+		//create second factory
+		EmbeddedNeo4jGraphDatabaseFactory factory2 = new EmbeddedNeo4jGraphDatabaseFactory();
+		factory2.initialize( properties );
+		GraphDatabaseService graphDatabaseService2 = factory2.create();
+
+		assertTrue( "Two instances must be same!", graphDatabaseService1 == graphDatabaseService2 );
+
+		graphDatabaseService1.shutdown();
+		graphDatabaseService2.shutdown();
```
DavideD: I thought I already explained that this is not a requirements. It is actually different from what happens when you use Neo4j embedded without Hibernate OGM and I don't see what it has to do with the issues you are trying to solve.

DavideD:
I'll have a look at it, but this is not was looking for for this PR.
It's still not clear why you need it. Which tests aren't working if you don't apply this PR?

schernolyas:
Hi @DavideD !

I see I faced to problem with exception"Cannot perform schema updates in a transaction that has performed data updates". And ... Now I am not understood why it occurs.

The PR is ready for review but not ready for merge :(

schernolyas:
HI @DavideD !

I found description of problem. Reason of failed tests described here https://stackoverflow.com/questions/33956238/whats-the-difference-between-placebotransaction-and-topleveltransaction .

It means that we have nested transactions.  Now  main questions are: where and how.

DavideD:
With the ORM update something has changed in the way transactions are managed. I'll have a look

DavideD:
I tried to have a look a this but I couldn't figure out the issue.
The problem occurs during the build but not when we run the single test but it only happens when we merge this PR.
I'm considering closing it for now

schernolyas:
okey. 
it is too long song  as say in Russia

PR #941, opened 2018-01-24
Ogm 1268 add integration test with spring (#32)

batousik:
Changes:
1) Refactored code, moved all IntTests into wildfly package
2) First runnable integration test with embedded mongodb


batousik:
has conflits, will create another pull request

PR #945, opened 2018-01-24
OGM-1379 Pass the right parameters to RemoteAuthenticationFailureTest

DavideD:
https://hibernate.atlassian.net/browse/OGM-1379

It turns out that the tests wasn't reading the right variables and also ci was launching the build with the wrong parameters.

This should fix the error we have on CI


DavideD:
This is a simple one, I'm going to integrate it because it fixes CI

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/945

gsmet:
@DavideD yeah sorry, it was OK for me, forgot to post something. It's a bit like dark magic but as long as it works, better merge it!

PR #949, opened 2018-02-05
OGM-1262 Remove experimental features from doc

DavideD:
https://hibernate.atlassian.net/browse/OGM-1262

Most of it was removed when we split the repositories


Sanne:
merged. Thanks!

PR #951, opened 2018-02-05
OGM-1384 If the global state config is not enabled a validation error…

fax4ever:
… is thrown.
https://hibernate.atlassian.net/browse/OGM-1384.

### DIFF
--- a/infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/logging/impl/Log.java
+++ b/infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/logging/impl/Log.java
@@ -51,10 +51,10 @@ public interface Log extends org.hibernate.ogm.util.impl.Log {
 			+ " Enable this property explicitly in your configuration to resolve this warning.")
 	void clusteringHashGroupsMustBeEnabled(String cacheName);
 
-	@Message(id = 1108, value = "Sequences or id generation is not supported for local caches")
+	@Message(id = 1108, value = "Sequences or di generation is not supported for local caches")
 	HibernateException counterCannotBeCreatedForLocalCaches();
 
-	@Message(id = 1109, value = "Sequences or id generation is not supported if global configuration is not enabled")
+	@Message(id = 1109, value = "Counter is not defined and cannot be created. Global persistent-location is missing in the Infinispan configuration")
 	HibernateException counterCannotBeCreatedWithoutGlobalConfiguration();
 
 }
--- a/infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/logging/impl/Log.java
+++ b/infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/logging/impl/Log.java
@@ -51,6 +51,10 @@ public interface Log extends org.hibernate.ogm.util.impl.Log {
 			+ " Enable this property explicitly in your configuration to resolve this warning.")
 	void clusteringHashGroupsMustBeEnabled(String cacheName);
 
-	@Message(id = 1108, value = "Sequences or di generation is not supported for local caches")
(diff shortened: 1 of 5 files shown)


infinispan-embedded/src/test/resources/infinispan-dist-noglobal.xml
```diff
@@ -0,0 +1,95 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ ~ Hibernate OGM, Domain model persistence for NoSQL datastores
+ ~
+ ~ License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ ~ See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+  -->
+
+<!--
+    This is a testing configuration, running in clustered mode.
```
DavideD: Could you add a line to explain what' s different in this configuration and why? In a couple of weeks we won't remember that this configuration file is used to test the missing property

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/logging/impl/Log.java
```diff
@@ -51,6 +51,10 @@
 			+ " Enable this property explicitly in your configuration to resolve this warning.")
 	void clusteringHashGroupsMustBeEnabled(String cacheName);
 
-	@Message(id = 1108, value = "Sequences or di generation is not supported for local caches")
+	@Message(id = 1108, value = "Sequences or id generation is not supported for local caches")
```
DavideD: It is not critical but this is a change that should be in a separate commit

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/logging/impl/Log.java
```diff
@@ -51,6 +51,10 @@
...
 	HibernateException counterCannotBeCreatedForLocalCaches();
+
+	@Message(id = 1109, value = "Sequences or id generation is not supported if global configuration is not enabled")
```
DavideD: Maybe we can add more details to the message:

Something like,  "Counter <counterName> is not defined and cannot be created. Global persistent-location is missing in the Infinispan configuration"

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/persistencestrategy/counter/ClusteredCounterCommand.java
```diff
@@ -71,6 +75,16 @@ public Number nextValue(NextValueRequest request) {
 
 	}
 
+	private void validateGlobalConfiguration() {
+		boolean globalConfigIsEnabled = cacheManager.getGlobalComponentRegistry()
+			.getGlobalConfiguration().globalState()
```
DavideD: I guess this works but is it possible to have the globalState without setting the persitent-location?

fax4ever: we have to try... but it is possible. I'll prepare a test for it.

fax4ever: if the persistence location is not present under global state the infinispan.xml file is not parsable by xsd, but counters are stored under project folder.

DavideD: ok

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/951#discussion_r166026543

PR #956, opened 2018-02-08
Upgrading to Hibernate Search 5.9.0.Final

Sanne:
And some related build cleanup:
 - https://hibernate.atlassian.net/browse/OGM-1389
 - https://hibernate.atlassian.net/browse/OGM-1395


bom/pom.xml
```diff
@@ -41,10 +41,11 @@
 
         <!-- Hibernate ORM -->
         <hibernateVersion>5.2.13.Final</hibernateVersion>
+        <hibernate-orm.module.slot>5.2</hibernate-orm.module.slot>
```
gsmet: I'm not sure it makes sense to declare them in the bom. I'm not sure it's worth it to make them visible to the users using our BOM.

What's the rationale for this change?

Sanne: 1# consistency: the Search one is already there.
2# It helps to have the module slot property close the the full version.
3# I see no problem in having these in the BOM, it doesn't hurt and it might be useful to some.

gsmet: Ah missed that the Search one was already there. Would you mind fixing the case to be consistent with the rest of the properties? I know the Search one was already wrong but we should better fix it once and for all.

I consider the BOM as being exposed to the users so it would be nice if we could be consistent.

Thanks!

DavideD:
Rebased and Merged https://github.com/hibernate/hibernate-ogm/pull/956

DavideD:
I don't think the position of the properties should postpone merging this PR.
@gsmet  Feel free to open a new issue. We might also check if there are other properties to move around

PR #957, opened 2018-02-09
OGM-1380 Create counters at startup

DavideD:
Follows https://github.com/hibernate/hibernate-ogm/pull/953

https://hibernate.atlassian.net/browse/OGM-1380

I've squashed together the 2 commits of the previous pr from @fax4ever and added a new commit that
removes some duplicated code and remove the needs of keeping track which Thread created the counter.

The difference from before is that I read the updated value and then remove the increment.

@fax4ever What do you think?


DavideD:
@Sanne do you have some other optimization in mind for the counters?

fax4ever:
@DavideD Very nice solution: update always the counter and return {counter - initial value}. In this way all the call are uniform, not distinguish between the first and the following invocation. Because the value to add is a constant value. I will reviewing.

fax4ever:
Merged, thank you @DavideD

PR #958, opened 2018-02-10
OGM-1397 Counter name should contain the table for @TableGenerator

DavideD:
https://hibernate.atlassian.net/browse/OGM-1397

Is it OK to include it for the next release?
If they update the library, people might have problem since the name of the counters has changed


fax4ever:
I reviewed it: in my opinion it is ok.
If we want insert in the next release I can merge it.

DavideD:
@Sanne is it OK to include this for 5.3 CR1 ?

Sanne:
yes better have it now than after a Final, since it changes the schema.

DavideD:
Cool, thanks

PR #959, opened 2018-02-12
OGM-1396 Improve the consistency of property naming in the pom files

gsmet:
https://hibernate.atlassian.net/browse/OGM-1396

(see the issue for the rationale of the changes)

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -23,7 +23,7 @@
 
         <!--
             These dependencies should be aligned with the ones from the WildFly version we support
-            See the version.wildfly property in the hibernate-ogm-parent pom
+            See the wildflyVersion property in the hibernate-ogm-parent pom
             See http://search.maven.org/#search|gav|1|g%3A"org.wildfly"%20AND%20a%3A"wildfly-parent"
         -->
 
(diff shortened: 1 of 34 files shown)


Sanne:
Hi @gsmet , sorry I appreciate the effort, but this is going in the opposite direction of what we need :(

We should eventually use the notation you can see used by WildFly as that's what productization needs. There's an open JIRA in HSEARCH to make Search consistent with that.. doesn't apply to OGM yet but if you like to see consistency, then please use the "right format".

Sanne:
More specifically about the `hibernate-orm.module.slot` as you raised that last week: in that case I don't actually like it to be consistent with artifact versions, as it's used in a very different context.
Would it be better to have their own format? The format I've been using is consistent with others `slot names` across various other projects.

gsmet:
Could you be more explicit about "what we need"? Because most of the issues we have with consistency is that there is no written rules of what it should be and everyone adds new properties with the case he likes at instant t.

Personnally, I'm all for switching to lowercase with dashes and dots (I really dislike the camel case for properties but being consistent is IMHO more important than my personal tastes) but, for instance, WF uses `version.<component>` whereas we use `<component>.version`.

So having a proper set of rules would definitely help instead of trying to guess "what we need" and end up with something that nobody likes.

gsmet:
> Would it be better to have their own format? The format I've been using is consistent with others slot names across various other projects.

Well, just take a look at our current BOM and ask yourself if it looks OK from an external point of view. It definitely looks bad with these properties being totally inconsistent with the rest.

And they are not used in different contexts: the `module.xml` files contain version properties (camel cased) and module slots (with dashes and dots).

So, if we need to follow rules for productization, let's get the rules straight and apply them cleanly everywhere.

Sanne:
There's no guidelines about naming of slots as nobody else uses them, nor they are exposed beyond the build.

So perhaps they shouldn't be in the BOM at all, move them to the parent pom or some other appropriate scope?

> And they are not used in different contexts: the module.xml files contain version properties (camel cased) and module slots (with dashes and dots).

I disagree, they are of different "type".

It gets really hairy when one mixes such things up; it's not very unusual for the content of the module slot property to match one of the versions, so such a mistake could be released undetected during our integration tests and still become a critical bug.

gsmet:
Let's close this one. +1 for the general POM to follow more closely the JBoss conventions.

PR #961, opened 2018-02-12
[PREVIEW] OGM-1390 A quick preview of the upgrade to ORM 5.3 dependency

DavideD:
It's a preview to comment some of the changes.
It should not be merged! 

The code is good enough to understand what's going on, though.

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
@@ -6,7 +6,6 @@
  */
 package org.hibernate.ogm.dialect.query.spi;
 
-import java.util.HashMap;
 import java.util.Map;
 
 import org.hibernate.engine.query.spi.NamedParameterDescriptor;
@@ -24,32 +23,13 @@ public abstract class RecognizerBasedParameterMetadataBuilder implements Paramet
 
 	@Override
 	public ParameterMetadataImpl buildParameterMetadata(String nativeQuery) {
-		ParamLocationRecognizer recognizer = new ParamLocationRecognizer();
+		ParamLocationRecognizer recognizer = new ParamLocationRecognizer( 0 );
 		parseQueryParameters( nativeQuery, recognizer );
 
-		final int size = recognizer.getOrdinalParameterLocationList().size();
-		final OrdinalParameterDescriptor[] ordinalDescriptors = new OrdinalParameterDescriptor[ size ];
-		for ( int i = 0; i < size; i++ ) {
-			final Integer position = recognizer.getOrdinalParameterLocationList().get( i );
-			ordinalDescriptors[i] = new OrdinalParameterDescriptor( i, null, position );
-		}
+		final Map<Integer, OrdinalParameterDescriptor> ordinalDescriptors = recognizer.getOrdinalParameterDescriptionMap();
+		final Map<String, NamedParameterDescriptor> namedDescriptors = recognizer.getNamedParameterDescriptionMap();
 
-		final Map<String, NamedParameterDescriptor> namedParamDescriptorMap = new HashMap<String, NamedParameterDescriptor>();
-		final Map<String, ParamLocationRecognizer.NamedParameterDescription> map = recognizer.getNamedParameterDescriptionMap();
-		for ( final String name : map.keySet() ) {
-			final ParamLocationRecognizer.NamedParameterDescription description = map.get( name );
-			namedParamDescriptorMap.put(
-					name,
-					new NamedParameterDescriptor(
-							name,
-							null,
-							description.buildPositionsArray(),
-							description.isJpaStyle()
-					)
-			);
-		}
-
-		return new ParameterMetadataImpl( ordinalDescriptors, namedParamDescriptorMap );
+		return new ParameterMetadataImpl( ordinalDescriptors, namedDescriptors );
 	}
 
 	/**
--- a/core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
@@ -25,6 +25,7 @@ public abstract class RecognizerBasedParameterMetadataBuilder implements Paramet
 	public ParameterMetadataImpl buildParameterMetadata(String nativeQuery) {
 		ParamLocationRecognizer recognizer = new ParamLocationRecognizer( 0 );
 		parseQueryParameters( nativeQuery, recognizer );
+		recognizer.complete();
 
 		final Map<Integer, OrdinalParameterDescriptor> ordinalDescriptors = recognizer.getOrdinalParameterDescriptionMap();
 		final Map<String, NamedParameterDescriptor> namedDescriptors = recognizer.getNamedParameterDescriptionMap();
--- a/core/src/main/java/org/hibernate/ogm/type/impl/EnumType.java
+++ b/core/src/main/java/org/hibernate/ogm/type/impl/EnumType.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.ogm.type.impl;
 
+import java.lang.invoke.MethodHandles;
 import java.sql.Types;
 
 import org.hibernate.HibernateException;
@@ -16,7 +17,6 @@ import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.ogm.model.spi.Tuple;
 import org.hibernate.ogm.util.impl.Log;
 import org.hibernate.ogm.util.impl.LoggerFactory;
-import java.lang.invoke.MethodHandles;
 import org.hibernate.type.CustomType;
 
 /**
@@ -87,7 +87,8 @@ public class EnumType extends GridTypeDelegatingToCoreType {
 
 	private void initEnumValues() {
 		if ( enumValues == null ) {
-			this.enumValues = coreEnumType.returnedClass().getEnumConstants();
+			Class<? extends Enum<?>> returnedClass = coreEnumType.returnedClass();
+			this.enumValues = returnedClass.getEnumConstants();
 			if ( enumValues == null ) {
 				throw new NullPointerException( "Failed to init enumValues" );
 			}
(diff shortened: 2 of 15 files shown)


core/src/test/java/org/hibernate/ogm/backendtck/inheritance/singletable/family/SingleTableInheritancePersistTest.java
```diff
@@ -45,7 +45,9 @@ public void setUp() {
 	@After
 	public void tearDown() {
 		try {
-			em.getTransaction().begin();
+			if ( !em.getTransaction().isActive() ) {
+				em.getTransaction().begin();
+			}
```
gsmet: Wondering if we should fix that or the test itself as it's not a good example of how things should be done.

But maybe we have all our tests like that and it's not worth fixing.

core/src/main/java/org/hibernate/ogm/type/impl/EnumType.java
```diff
@@ -87,7 +87,8 @@ public Object nullSafeGet(Tuple rs, String name, SharedSessionContractImplemento
 
 	private void initEnumValues() {
 		if ( enumValues == null ) {
-			this.enumValues = coreEnumType.returnedClass().getEnumConstants();
+			Class<Enum<?>> returnedClass = coreEnumType.returnedClass();
+			this.enumValues = returnedClass.getEnumConstants();
```
gsmet: I can't see a difference between the 2 codes. But either is OK to me.

core/src/test/java/org/hibernate/ogm/test/datastore/DatastoreProviderGeneratingSchema.java
```diff
@@ -51,32 +45,17 @@
...
-							String columnName = column.getCanonicalName();
-							// do something with column
-						}
-						//TODO handle unique constraints?
-					}
-				}
-			}
-
-			throw new RuntimeException( "STARTED!" );
+			schemaInitialized = true;
```
gsmet: Looks like a good simplification but what was the issue exactly?

core/src/test/java/org/hibernate/ogm/test/jpa/JPAStandaloneNoOGMTest.java
```diff
@@ -39,7 +39,7 @@ public void testJTAStandaloneNoOgm() throws Exception {
 			fail( "Expected exception was not raised" );
 		}
 		catch ( PersistenceException pe ) {
-			assertThat( Throwables.getRootCause( pe ).getMessage() ).contains( "hibernate.dialect" );
+			assertThat( Throwables.getRootCause( pe ).getMessage() ).contains( "No Persistence provider for EntityManager named noogm" );
```
gsmet: Wondering if the purpose of the test was not also to validate that the error message was explicit enought about the issue (i.e. the fact that `hibernate.dialect` was not defined). Maybe the message is in the `cause`?

core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
```diff
@@ -24,32 +23,13 @@
 
 	@Override
 	public ParameterMetadataImpl buildParameterMetadata(String nativeQuery) {
-		ParamLocationRecognizer recognizer = new ParamLocationRecognizer();
+		ParamLocationRecognizer recognizer = new ParamLocationRecognizer( 0 );
```
gsmet: So... this is the change in ORM 5.3 that I find the most concerning (the fact that the base for parameter numbering has changed). I'm really worried about it, even if we have a configuration knob to restore the old behavior.

I understand you chose to be entirely compatible with the old behavior of OGM? I must admit that I'm wondering if we should follow the ORM way or be more conservative. I'm a bit torn on this subject.

@Sanne what's your take on this?

core/src/test/java/org/hibernate/ogm/test/type/converter/JpaAttributeConverterGridTypeApplicationTest.java
```diff
@@ -81,11 +79,12 @@ protected void configure(Map<String, Object> cfg) {
...
 	@Override
 	protected Class<?>[] getAnnotatedClasses() {
-		return new Class<?>[] {
+		return new Class<?>[]{
 				Printer.class
 		};
 	}
 
+	@SuppressWarnings("serial")
 	public static class CustomMapDialect extends MapDialect {
```
gsmet: IIRC, we decided to ignore these warnings globally (at least that's what I did in my Eclipse configuration).

This commit is a lot of noise for such a small thing.

core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
DavideD: To be fair, this is one of the part where I need to investigate more, they now have compliance settings that should help use to decide. For now, it seemed easier to be conservative.

core/src/test/java/org/hibernate/ogm/test/datastore/DatastoreProviderGeneratingSchema.java
DavideD: The issue is that now in case of error during the factory creation no exception is thrown.
This is per J{PA specification.

This also caused some discussion on https://hibernate.atlassian.net/browse/HHH-12247

core/src/main/java/org/hibernate/ogm/type/impl/EnumType.java
DavideD: I think it's a leftover

core/src/test/java/org/hibernate/ogm/backendtck/inheritance/singletable/family/SingleTableInheritancePersistTest.java
DavideD: I think it might be because of the way the test is written, but I will check before sending a clean PR

core/src/test/java/org/hibernate/ogm/test/type/converter/JpaAttributeConverterGridTypeApplicationTest.java
DavideD: OK

core/src/test/java/org/hibernate/ogm/test/jpa/JPAStandaloneNoOGMTest.java
DavideD: I'm pretty sure the exception thrown was completely different but I will double check

core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
gsmet: Yes, that's what I suspected. But TBH, I'm unsure on how we should follow up on that. It seems annoying to break every applications using ordinal parameters by default. But on the other hand, I wouldn't want us to be inconsistent with ORM. I can see heated discussions coming :)

core/src/test/java/org/hibernate/ogm/test/jpa/JPAStandaloneNoOGMTest.java
gsmet: OK, that would be a bit unfortunate. Maybe it's worth digging to see if ORM forgot to add a cause or if it's totally impossible to have a precise error message.

DavideD: This change in behavior is due to the changes in `HibernatePersistenceProvider.createEntityManagerFactory`.

Per JPA specification this method doesn't have to throw exceptions. Meaning that, instead of throwing the exception with error message will return `org.hibernate.HibernateException: Access to DialectResolutionInfo cannot be null when 'hibernate.dialect' not set` it will return null

For this reason our test fails, because a different exception is thrown but the reason is the same.
I think this new exception explains bettere what happened then the old one. In fact in the cofiguration we left the persistence provider away on purpose.

I guess now we will have to update the OGM provider  to not throw exceptions anymore as well

core/src/main/java/org/hibernate/ogm/type/impl/EnumType.java
DavideD: So, as a quick experiment, I removed this commit and I have a compilation error but I don't see any difference with the classes involved. I think I'm too tired now, if you want to try to solve this Java puzzle, be my guest :)

DavideD: Otherwise I will check tomorrow

gsmet: It might be some weird generics issue with the compiler. It's not worth wasting time IMHO. Let's keep your new code.

core/src/test/java/org/hibernate/ogm/test/jpa/JPAStandaloneNoOGMTest.java
gsmet: OK, thanks for taking the time to explore this. Yes, we probably need to be in par with what is done in ORM. I have seen that the behavior might change depending on the compliance setting so, if it's indeed the case, we need to have a similar behavior.

core/src/test/java/org/hibernate/ogm/backendtck/inheritance/singletable/family/SingleTableInheritancePersistTest.java
DavideD: It seems this change is not actually necessary, I probably  got tricked by some other exceptions I hadn't solved yet

core/src/test/java/org/hibernate/ogm/test/jpa/JPAStandaloneNoOGMTest.java
Sanne: > Per JPA specification this method doesn't have to throw exceptions. 

Interesting, I'm familiar with that bootstrap process and Steve is right (of course?) on that HHH JIRA on which we were discussing this, but this doesn't look like necessary the case covered by the specification.

Not a blocker for OGM but I'll go see some clarifications.

core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
Sanne: that's another one of the annoying "JPA mandates it" cases... we'll just need to follow the example of ORM.

DavideD:
Closing this one for now

PR #968, opened 2018-02-21

QwerTech:
[OGM-1235](https://hibernate.atlassian.net/browse/OGM-1235) Added Testcase that reproducing a bug

Fix via adding registerColumnType binary in OgmDialect


Sanne:
@Hibernate-CI add to whitelist

DavideD:
Can you describe a bit how this fixes the issue?

QwerTech:
Main reason for this issue is that sql type binary (-2) cannot be found in registered TypeNames while generating create statmens.
I've looked in how H2 and ORM work with same scenario. And I've found in H2Dialect constructor a lot of registerColumnType method calls. So I decided look in OgmDialect and not found extra columnType registrations.
I add one for binary type and It's work.
Maybe we need to register some more types, but I'm not sure, I need a consultation.

core/src/test/java/org/hibernate/ogm/backendtck/id/TableIdGeneratorInheritanceTest.java
```diff
@@ -0,0 +1,99 @@
...
+		em.getTransaction().commit();
+
+		em.clear();
+
+		em.getTransaction().begin();
+
+		loadedTruck = em.find( Truck.class, loadedTruck.getId() );
+
+		assertThat( loadedTruck.getId() ).isNotNull();
+		assertThat( loadedTruck ).isNotNull();
```
fax4ever: Perhaps it is better to reverse the order of the assertions. Verifying that loadedTruck is not null before verify the id.

QwerTech: Yes, you right. Excuse for mistake. Fixed this.

QwerTech:
Fixed assertions

core/src/main/java/org/hibernate/ogm/dialect/impl/OgmDialect.java
```diff
@@ -31,6 +32,7 @@
 
 	public OgmDialect(GridDialect gridDialect) {
 		this.gridDialect = gridDialect;
+		registerColumnType( Types.BINARY,"binary" );
```
fax4ever: missing space between BINARY, and "binary".
Maybe also an in line comment like: "enabling support for binary types"

DavideD:
@QwerTech I'm going to squash together the commits related to the tests, there is no reason to keep them separated

DavideD:
Just a small comment about style.

The format of our commit masseges is always:

OGM-1235 <message>

DavideD:
Also it causes less problems during rebases if new files are added in a separate commit when possible/make sense

core/src/test/java/org/hibernate/ogm/backendtck/id/TableIdGeneratorInheritanceTest.java
```diff
@@ -0,0 +1,100 @@
...
+		em.clear();
+
+		em.getTransaction().begin();
+
+		Truck loadedTruck = em.find( Truck.class, truck.getId() );
+
+		assertThat( loadedTruck ).isNotNull().as( "truck not loaded" );
+		assertThat( loadedTruck.getId() ).isNotNull().as( "truck id not loaded" );
+		assertThat( loadedTruck.getId() ).isEqualTo( truck.getId() )
+				.as( "loaded truck id is not equals saved truck id" );
```
DavideD: Let's simplify the assertions:
```
		assertThat( loadedTruck ).isNotNull();
		assertThat( loadedTruck.getId() ).isEqualTo( truck.getId() );
```

No need to check that getId is not null, it's already done when you do the isEqualTo. The  messages aren't very helpful in this case and only make the code base bigger.

core/src/test/java/org/hibernate/ogm/backendtck/id/TableIdGeneratorInheritanceTest.java
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	@Entity
+	@Table(name = "TRUCK")
+	private static class Truck extends BaseCar {
+	}
+
+	@Entity
+	@Table(name = "PASSENGER_CAR")
+	private static class PassengerCar extends BaseCar {
```
DavideD: `PassengerCar` is not needed for this test, let's remove it.

core/src/main/java/org/hibernate/ogm/dialect/impl/OgmDialect.java
```diff
@@ -31,6 +32,8 @@
 
 	public OgmDialect(GridDialect gridDialect) {
 		this.gridDialect = gridDialect;
+		//enabling support for binary types
+		registerColumnType( Types.BINARY, "binary" );
```
DavideD: This solution seems to work but I'm not convinced it's the right approach.
The problem seems to be that OgmDialect doesn't override `getDefaultMultiTableBulkIdStrategy()` and therfore it is using the one in the `Dialect` superclass. This is doing things that it shouldn't leading to the exception.

If I change the implementation, this test passes without the need to `registerColumnType`

core/src/test/java/org/hibernate/ogm/backendtck/id/TableIdGeneratorInheritanceTest.java
QwerTech: I've added this class to make sure, that everything will be all right not only with a single successor class, but with two or more.

core/src/test/java/org/hibernate/ogm/backendtck/id/TableIdGeneratorInheritanceTest.java
QwerTech: If I understand isEqualTo right, then it would be true if all id will be null.
But it's not right, that's why I've added this assert.
If you stil think it's excess, then I remove it?

core/src/test/java/org/hibernate/ogm/backendtck/id/TableIdGeneratorInheritanceTest.java
DavideD: In general, a test aim to replicate the exception using the minimum set of elements.
If you remove this class the exception is still thrown so you have already replicated the problem the issue is about.

If you want to see that it works with two classes, you should add an additional test that does something with PassengerCar. The way it is now, you are not actually proving that it works with multiple classes.

QwerTech: It's initializing for now.
What do you advise to do? Remove or add tests?

core/src/test/java/org/hibernate/ogm/backendtck/id/TableIdGeneratorInheritanceTest.java
DavideD: If both Id are null then it works because they are, in fact, equals :)

But you raise a valid point about checking that the id has been generated, I would prefer to have a separate test method for this and you should probably check the truck.getId() is not null after session.persist(). Having it in a separate test method make it easier for other programmers to understand what's going on.

QwerTech: Ok, I'll devide test to two cases.

core/src/test/java/org/hibernate/ogm/backendtck/id/TableIdGeneratorInheritanceTest.java
DavideD: Let's keep it simple for now and remove it.

QwerTech:
Cleared test case, added it via separate commit

core/src/main/java/org/hibernate/ogm/dialect/impl/OgmDialect.java
QwerTech: @DavideD What in the end will we do? I'll look at the solution through an override or you yourself?

QwerTech:
@DavideD except comment about override do you have any other remarks?

core/src/main/java/org/hibernate/ogm/dialect/impl/OgmDialect.java
DavideD: Can you have a look at `getDefaultMultiTableBulkIdStrategy`? I started with an empty implementation and it seems to pass the tests you added.

But we need some additional tests where we make sure that when we delete the entity the elements in all the tables are deleted.  Probably we have other scenarios to test, but we can start with this.

A simple way to do this is to create an entity, delete it and check that the db is empty.

schernolyas:
Hi @DavideD !
What is status of the PR? Should @QwerTech change/fix something?

DavideD:
I've sent a message 7 days ago asking to @QwerTech  if he could fix this using a difference approach:

> Can you have a look at getDefaultMultiTableBulkIdStrategy? I started with an empty implementation and it seems to pass the tests you added.

The problem is that it's more complicated than I initially thought because it requires to test that changes during updates or deletions get propagated to all the tables involved.

QwerTech:
Thanks @DavideD. I look at this week for solution.

DavideD:
> Thanks @DavideD. I look at this week for solution.

Thanks, I'll close this in the meanwhile. Please, feel free to reopen this pull request or send a new one when you are ready.

PR #969, opened 2018-02-21
OGM-1359  ORM 5.3

DavideD:
https://hibernate.atlassian.net/browse/OGM-1390

I think this version is good to be integrated once we have a release of Search 5.10.

But et me know what you think.

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
@@ -26,6 +26,7 @@ public abstract class RecognizerBasedParameterMetadataBuilder implements Paramet
 	public ParameterMetadataImpl buildParameterMetadata(String nativeQuery) {
 		ParamLocationRecognizer recognizer = new ParamLocationRecognizer();
 		parseQueryParameters( nativeQuery, recognizer );
+		recognizer.complete();
 
 		final int size = recognizer.getOrdinalParameterLocationList().size();
 		final OrdinalParameterDescriptor[] ordinalDescriptors = new OrdinalParameterDescriptor[ size ];
--- a/core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
+++ b/core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
@@ -6,7 +6,6 @@
  */
 package org.hibernate.ogm.dialect.query.spi;
 
-import java.util.HashMap;
 import java.util.Map;
 
 import org.hibernate.engine.query.spi.NamedParameterDescriptor;
@@ -24,33 +23,21 @@ public abstract class RecognizerBasedParameterMetadataBuilder implements Paramet
 
 	@Override
 	public ParameterMetadataImpl buildParameterMetadata(String nativeQuery) {
-		ParamLocationRecognizer recognizer = new ParamLocationRecognizer();
+		// At the moment we don't care about JDBC base count for positional parameters because
+		// we don't have a dialect that supports them.
+		// After a quick inspeciton it seems that only with Neo4j would be easy to make it work
+		// and even in that scenario I'm not sure we should care.
+		// I think it makes sense to keep track of the JPA compliance problem in a separate issues.
+		// In this case after solving OGM-1407
+		int jdbcStyleOrdinalCountBase = 0;
+		ParamLocationRecognizer recognizer = new ParamLocationRecognizer( jdbcStyleOrdinalCountBase );
 		parseQueryParameters( nativeQuery, recognizer );
 		recognizer.complete();
 
-		final int size = recognizer.getOrdinalParameterLocationList().size();
-		final OrdinalParameterDescriptor[] ordinalDescriptors = new OrdinalParameterDescriptor[ size ];
-		for ( int i = 0; i < size; i++ ) {
-			final Integer position = recognizer.getOrdinalParameterLocationList().get( i );
-			ordinalDescriptors[i] = new OrdinalParameterDescriptor( i, null, position );
-		}
+		final Map<Integer, OrdinalParameterDescriptor> ordinalDescriptors = recognizer.getOrdinalParameterDescriptionMap();
+		final Map<String, NamedParameterDescriptor> namedDescriptors = recognizer.getNamedParameterDescriptionMap();
 
-		final Map<String, NamedParameterDescriptor> namedParamDescriptorMap = new HashMap<String, NamedParameterDescriptor>();
-		final Map<String, ParamLocationRecognizer.NamedParameterDescription> map = recognizer.getNamedParameterDescriptionMap();
-		for ( final String name : map.keySet() ) {
-			final ParamLocationRecognizer.NamedParameterDescription description = map.get( name );
-			namedParamDescriptorMap.put(
-					name,
-					new NamedParameterDescriptor(
-							name,
-							null,
-							description.buildPositionsArray(),
-							description.isJpaStyle()
-					)
-			);
-		}
-
-		return new ParameterMetadataImpl( ordinalDescriptors, namedParamDescriptorMap );
+		return new ParameterMetadataImpl( ordinalDescriptors, namedDescriptors );
 	}
 
 	/**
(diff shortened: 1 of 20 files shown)


core/src/main/java/org/hibernate/ogm/transaction/emulated/impl/EmulatedLocalTransactionCoordinatorBuilder.java
```diff
@@ -86,7 +86,7 @@ public void commit() {
 
 		@Override
 		public void rollback() {
-			status = TransactionStatus.NOT_ACTIVE;
+			status = TransactionStatus.ROLLED_BACK;
```
Sanne: I'm not familiar with the new status codes, but are you sure this shouldn't be `MARKED_ROLLBACK` ?

pom.xml
```diff
@@ -68,6 +69,12 @@
         <!-- Target version of WildFly (for testing and JipiJapa integration) -->
         <version.wildfly>11.0.0.Final</version.wildfly>
         <hibernateWildflyClassifier>wildfly-11-dist</hibernateWildflyClassifier>
+        <jpa22patch.version>1.0.0.Beta1</jpa22patch.version>
+        <jpa22patch.filename>hibernate-jpa-api-2.2-wildflymodules-${jpa22patch.version}-wildfly-${version.wildfly}-patch.zip</jpa22patch.filename>
+
+        <cdi20patch.version>3.0.2.Final</cdi20patch.version>
+        <cdi20patch.filename>wildfly-${version.wildfly}-weld-${cdi20patch.version}-patch.zip</cdi20patch.filename>
```
Sanne: does OGM need the CDI 2.0 implementation?

core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
```diff
@@ -24,32 +23,19 @@
 
 	@Override
 	public ParameterMetadataImpl buildParameterMetadata(String nativeQuery) {
-		ParamLocationRecognizer recognizer = new ParamLocationRecognizer();
+		// At the moment we don't care because neo4j doesn't support positional paramters
```
gsmet: s/paramters/parameters/

core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
```diff
@@ -24,32 +23,19 @@
...
+		// I'm not even sure it would make a difference from the Neo4j point of view.
+		// We can change this in a separate issue and check the right value using SessionFactoryImplementor
+		// or SessionFactoryOptions
```
gsmet: I'm a bit puzzled by this comment. We are in the core module and you are only talking about Neo4j?

As far as I can see, you decided against taking into account the ORM option in OGM. Why is that?

pom.xml
gsmet: I think it's probably ORM 5.3 needing it.

core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
DavideD: I thought the comment was clear, but it seems it's not.

Currently, positional parameters query don't work or are not supported. I was thinking about Neo4j because is the only dialect at the the moment that would be affected by this change.

DavideD: This issue is more connected to test positional parameters and  think we can work on it when solving this issue: https://hibernate.atlassian.net/projects/OGM/issues/OGM-1407

pom.xml
DavideD: New annotations have been introduced and you have this errors at start-up if you don't patch WildFly: 
```
Caused by: java.lang.NoClassDefFoundError: javax/persistence/TableGenerators
```

core/src/main/java/org/hibernate/ogm/transaction/emulated/impl/EmulatedLocalTransactionCoordinatorBuilder.java
DavideD: I don't think these are new, anyway I  will have a look.

core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
gsmet: Hmm, OK. So if it's not supported at all, let's be more generic in the comment and clearly states that it's not supported (and why not add a pointer to the issue).

That being said, does it take a lot of effort to mimic the ORM behavior and get ready for when it's supported? I'm a bit worried we could forget to update that later.

(If it's a lot of efforts, I understand we don't do the extra mile now though but we should add a note in the positional parameters issue description)

core/src/main/java/org/hibernate/ogm/transaction/emulated/impl/EmulatedLocalTransactionCoordinatorBuilder.java
DavideD: I think it's fine to use ROLLED_BACK in this scenario. There are some expectation around it and if the status is different the rollback is tried again causing event to get caught twice (and some of our tests fail).

core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
DavideD: I prefer to work on it when we work on the positional parameters issue because we need to add new tests to check that everything works as expected. I've updated the message and the issue, I'm confident we won't forget about it.

DavideD:
Jenkins, retest this please

pom.xml
fax4ever: JPA 2.2 => JEE 8 => CDI 2.0. Then it could make sense in term of specification

core/src/main/java/org/hibernate/ogm/dialect/query/spi/RecognizerBasedParameterMetadataBuilder.java
gsmet: OK, cool, thanks for updating the issue, that's all I asked for :).

pom.xml
gsmet: Yes, IIRC, an API added in CDI 2.0 was required for the work of Yoann and Steve on ORM/Search/CDI. @yrodiere could confirm without a doubt.

DavideD:
Jenkins, retest this please

DavideD:
I've updated this commit: https://github.com/hibernate/hibernate-ogm/pull/969/commits/4e0005a4e6ed75773fe7162c4cd1e5f9e25911f4

pom.xml
yrodiere: Yep, we definitely need CDI 2.0, because of a new method we used in the implementation, that is not available in CDI 1.x: namely `javax.enterprise.inject.spi.BeanManager#createInstance`.
But as far as I can tell, the goal has always been to support CDI 2.0 and nothing else.

Sanne: What if I'm not interested in CDI, now I have to have it on classpath because of the JPA spec?

That's horrible.

yrodiere: What I said is that *if* you have it on the classpath, it must be the right version, and this version is CDI 2.0.
Your original comment was about wildfly, so obviously in an EE application server where Hibernate ORM implements JPA 2.2, you need CDI, and the correct version at that. That seems reasonable.
Now, in Java SE mode, and if you don't have CDI on the classpath, I think you should be alright, because ORM has the appropriate checks in `org.hibernate.resource.beans.spi.ManagedBeanRegistryInitiator#resoveBeanContainer`.

yrodiere: > What I said is

Well, that's what I meant to say at least :) Sorry if it wasn't clear.

Sanne: ah, thanks I'm relieved.

DavideD:
I've updated the PR to use Hibernate Search 5.10.0.Beta1.

What do you think? Should we merge it?

gsmet:
@DavideD I don't see us releasing something before the ORM 5.3 release so I would say let's merge it.

PR #971, opened 2018-02-22
OGM-1347 Add configuration option for generated protobuf file

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1347

### DIFF
--- a/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/configuration/impl/InfinispanRemoteConfiguration.java
+++ b/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/configuration/impl/InfinispanRemoteConfiguration.java
@@ -102,6 +102,8 @@ public class InfinispanRemoteConfiguration {
 
 	private String schemaPackageName;
 
+	private String schemaFileName;
+
 	private Properties clientProperties;
 
 	private boolean createCachesEnabled;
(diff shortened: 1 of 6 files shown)


infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/configuration/impl/InfinispanRemoteConfiguration.java
```diff
@@ -173,14 +179,29 @@ public void initConfiguration(Map<?, ?> configurationMap, ServiceRegistryImpleme
 				.withDefault( InfinispanRemoteProperties.DEFAULT_SCHEMA_PACKAGE_NAME )
 				.getValue();
 
+		this.schemaFileName = propertyReader
+				.property( InfinispanRemoteProperties.SCHEMA_FILE_NAME, String.class )
+				.withDefault( InfinispanRemoteProperties.DEFAULT_SCHEMA_FILE_NAME )
```
DavideD: I will change it but, so that you know, you can apply a validation while reading the property.
Check the method `withValidator`

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/971 with some changes I squashed in the commit:

- I've created a validator for the property
- I've added a test to check the default name is used when one is not defined

Thanks

PR #970, opened 2018-02-22
Ogm 1405 Choosing a template configuration cache for Infinispan Remote dialect

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1405
https://hibernate.atlassian.net/browse/OGM-1329

PR #972, opened 2018-02-23
OGM-1405

DavideD:
Follows: https://github.com/hibernate/hibernate-ogm/pull/970
[OGM-1405](https://hibernate.atlassian.net/browse/OGM-1405) [OGM-1329](https://hibernate.atlassian.net/browse/OGM-1329)

@fax4ever I've added some commits on top of your PR, could you have a quick look and let me know if you agree with the changes?

I didn't like the property name: `use_cache_as_template`
and I replaced it with `new_cache_template`.
To be fair I don't like this one either, maybe we should call it `template_for_new_caches`?

@Sanne Any suggestion? It's a property to define the default template to use when we create new caches in infinispan and the user doesn't want to use the one defined as default in the infinispan configuration.


fax4ever:
@DavideD,
**template_for_new_caches** could be a very nice and talking name in my opinion.

I had a look, and I'm definitely agree with all your improvements.
Thank you

DavideD:
I'm going to merge this one, but I'm not convinced about the documentation parapgraph. I'm going to send a new PR with some changes

DavideD:
@fax4ever this only work for infinispan-remote, would it be possible to do it for embedded as well?

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/972

Thanks

PR #973, opened 2018-02-27
OGM-614  java.sql.Timestamp

DavideD:
https://hibernate.atlassian.net/browse/OGM-614


core/src/main/java/org/hibernate/ogm/type/impl/StringTimestampTypeDescriptor.java
```diff
@@ -0,0 +1,39 @@
...
+import org.hibernate.engine.spi.Mapping;
+import org.hibernate.ogm.type.descriptor.impl.TimestampDateTypeDescriptor;
+import org.hibernate.ogm.type.descriptor.impl.TimestampMappedAsStringGridTypeDescriptor;
+
+/**
+ * For {@link Timestamp} objects use a String representation.
+ *
+ * @author Davide D'Alto
+ */
+public class StringTimestampTypeDescriptor extends AbstractGenericBasicType<Date> {
```
fax4ever: Serializing a Date as a String: maybe the actual String written on datastore could be TimeZone dependent.
What could happen if the server, who did the put operation, had a different TimeZone from the server that perform a get on the same Entity? I'm asking myself if the same Date would be return in that case.

DavideD: Yes, that it is an issue right now, when I worked on that I didn't consider timezones. We have actually a conversion for dates that comes with the class Iso8601StringDateType. I would like to use these types instead.

We need to be careful though because it will change the mapping. We should create a separate issue.

Good quesiton

fax4ever: I see the point here. Maybe in another issue we could deep the topic.

DavideD:
I've rebased this one to the latest changes.
Is there any readon not to merge it?

core/src/main/java/org/hibernate/ogm/type/impl/TimestampGridType.java
```diff
@@ -0,0 +1,39 @@
...
+
+	public static final TimestampGridType INSTANCE = new TimestampGridType();
+
+	public TimestampGridType() {
+		super( TimestampGridTypeDescriptor.INSTANCE, JdbcTimestampTypeDescriptor.INSTANCE );
+	}
+
+	@Override
+	public String getName() {
+		return "mongo_timestamp";
```
gsmet: This is in core and it has `mongo_` in its name, is it expected?

DavideD: The javadoc as well

DavideD: Fixed

DavideD:
I think I applied all the remarks for this one

DavideD:
Anything else missing?

core/src/main/java/org/hibernate/ogm/type/impl/TimestampGridTypeDescriptor.java
```diff
@@ -0,0 +1,53 @@
...
+			public X extract(Tuple resultset, String name) {
+				Date document = (Date) resultset.get( name );
+
+				if ( document == null ) {
+					return null;
+				}
+
+				return javaTypeDescriptor.wrap( new Timestamp( document.getTime() ), null );
+			}
+		};	}
```
gsmet: The indentation is weird here.

gsmet:
@DavideD added one very minor comment, looks good otherwise.

DavideD:
Thanks @gsmet. I will fix the indentation and merge it

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/973

PR #977, opened 2018-02-28
OGM-1411 MongoDB native query aggregate pipeline: avoid to change original pipeline

fax4ever:

### DIFF
--- a/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBNativeQueryPaginationTest.java
+++ b/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBNativeQueryPaginationTest.java
@@ -8,6 +8,7 @@ package org.hibernate.ogm.datastore.mongodb.test.query.nativequery;
 
 import static org.fest.assertions.Assertions.assertThat;
 
+import java.util.Arrays;
 import java.util.List;
 
 import org.hibernate.Session;
@@ -25,21 +26,21 @@ import org.junit.Test;
  */
 public class MongoDBNativeQueryPaginationTest extends OgmTestCase {
 
-	public static final String QUERY_FIND = "db.WILDE_POEM.find( { 'author' : 'Oscar Wilde' } )";
-	public static final String QUERY_AGGREGATE_MATCH = "db.WILDE_POEM.aggregate( [ { '$match' : { 'author' : 'Oscar Wilde' } } ] )";
+	public static final String QUERY_FIND = "db.WILDE_POEM.find( { '$query' : { 'author' : 'Oscar Wilde' }, '$orderby' : { 'name' : 1 } } )";
+	public static final String QUERY_AGGREGATE_MATCH = "db.WILDE_POEM.aggregate( [ { '$match' : { 'author' : 'Oscar Wilde' } }, { '$sort': {'name': 1 } } ] )";
 
 	private final OscarWildePoem portia = new OscarWildePoem( 1L, "Portia", "Oscar Wilde", 1881 );
 	private final OscarWildePoem athanasia = new OscarWildePoem( 2L, "Athanasia", "Oscar Wilde", 1879 );
 	private final OscarWildePoem imperatrix = new OscarWildePoem( 3L, "Ave Imperatrix", "Oscar Wilde", 1882 );
 	private final OscarWildePoem intellectualis = new OscarWildePoem( 4L, "Amor Intellectualis", "Oscar Wilde", 1881 );
-	private final OscarWildePoem apologia = new OscarWildePoem( 5L, "Apologias ", "Oscar Wilde", 1881 );
+	private final OscarWildePoem apologia = new OscarWildePoem( 5L, "Apologias", "Oscar Wilde", 1881 );
 	private final OscarWildePoem easter = new OscarWildePoem( 6L, "Easter Day", "Oscar Wilde", 1881 );
 	private final OscarWildePoem rome = new OscarWildePoem( 7L, "Rome Unvisited", "Oscar Wilde", 1881 );
 	private final OscarWildePoem miniato = new OscarWildePoem( 8L, "San Miniato", "Oscar Wilde", 1881 );
 	private final OscarWildePoem liberty = new OscarWildePoem( 9L, "Sonnet to Liberty", "Oscar Wilde", 1881 );
 	private final OscarWildePoem vita = new OscarWildePoem( 10L, "Vita Nuova", "Oscar Wilde", 1881 );
 
-	private final OscarWildePoem[] poems = { portia, athanasia, imperatrix, intellectualis, apologia, easter, rome, miniato, liberty, vita };
+	private final OscarWildePoem[] poems = { intellectualis, apologia, athanasia, imperatrix, easter, portia, rome, miniato, liberty, vita };
 
 	@Before
 	public void init() {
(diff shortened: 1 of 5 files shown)


mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBNativeQueryPaginationTest.java
```diff
@@ -0,0 +1,202 @@
...
+				.setFirstResult( 5 )
+				.setMaxResults( 5 )
+				.list();
+
+			assertThat( result.size() ).isEqualTo( 5 );
+		} );
+	}
+
+	@Test
+	public void test_aggregation_match_page1_page2_5() {
```
DavideD: It's not critical, but when we create tests we use the capitalized approach. I will update it

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBNativeQueryPaginationTest.java
```diff
@@ -0,0 +1,202 @@
...
+
+			assertThat( result.size() ).isEqualTo( 5 );
+
+			result = session.createNativeQuery( QUERY_FIND )
+				.addEntity( OscarWildePoem.TABLE_NAME, OscarWildePoem.class )
+				.setFirstResult( 5 )
+				.setMaxResults( 5 )
+				.list();
+
+			assertThat( result.size() ).isEqualTo( 5 );
```
DavideD: Checking the result size is not enough because you don't know if the results are the one you expected.
It could return always the same 5 results for example.

Please, update the tests so that it checks that every single result is the one we expect. You can add an order by to make sure that there are no surprises.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBNativeQueryPaginationTest.java
```diff
@@ -0,0 +1,202 @@
...
+import org.hibernate.ogm.utils.OgmTestCase;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * The test pagination behaviour for native queries.
+ * Using find and aggregate pipeline.
+ *
```
DavideD: I would change this to `Test the pagination for native queries with MongoDB`

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBNativeQueryPaginationTest.java
```diff
@@ -0,0 +1,202 @@
...
+ * @author Fabio Massimo Ercoli
+ */
+public class MongoDBNativeQueryPaginationTest extends OgmTestCase {
```
DavideD: I think it would be nice to have a similar test in core as well for HQL queries.
Maybe you can add it reusing this code? Not sure.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBNativeQueryPaginationTest.java
```diff
@@ -0,0 +1,202 @@
...
+			assertThat( result.size() ).isEqualTo( 5 );
+		} );
+	}
+
+	@Override
+	protected Class<?>[] getAnnotatedClasses() {
+		return new Class<?>[] { OscarWildePoem.class };
+	}
+
+	public void inTransaction(Consumer<Session> consumer) {
```
DavideD: Nice

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBNativeQueryPaginationTest.java
```diff
@@ -0,0 +1,202 @@
...
+		}
+		catch (Throwable t) {
+			if ( transaction.isActive() ) {
+				transaction.rollback();
+			}
+			throw t;
+		}
+
+		session.clear();
+		session.close();
```
DavideD: Session is autoclosable you can use` try (OgmSession session = openSession) { ... }`, also the session.clear shouldn't be necessary (I think)

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBNativeQueryPaginationTest.java
fax4ever: No problem, I'll fix the name in another commit.

DavideD: This issue is quite simple, feel free to squash the changes and pushed force them... no need to have separate commits for this.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBNativeQueryPaginationTest.java
fax4ever: Could be nice! Maybe we could create another Issue for that?

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBNativeQueryPaginationTest.java
fax4ever: I've learned it from @dreab8 :)

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBNativeQueryPaginationTest.java
fax4ever: Could be a nice improvement. Thank you

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBNativeQueryPaginationTest.java
DavideD: We could, but it's sort of related to what you are doing and I don't feel the need to create a separate issue for it. But I don't have a strong opinion on this, feel free to d oas you prefer.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBNativeQueryPaginationTest.java
DavideD: Maybe you can put this method in core so that we can reuse it somewhere else? It doesn't seem specific to this class/dialect

fax4ever: move to OgmTestCase base class?

DavideD: I was thinking about  a static method in some utility class. Something like JPAUtils maybe. So that it can be used even  in those test not extending OgmTestCase

fax4ever:
@DavideD thank you for the review. I found interesting hints.

gsmet:
Jenkins retest this please

fax4ever:
Jenkins retest please this

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBNativeQueryPaginationTest.java
```diff
@@ -0,0 +1,253 @@
...
+			List result = session.createNativeQuery( QUERY_AGGREGATE_MATCH )
+				.addEntity( OscarWildePoem.TABLE_NAME, OscarWildePoem.class )
+				.setFirstResult( 0 )
+				.setMaxResults( 10 )
+				.list();
+
+			assertThat( result.size() ).isEqualTo( 10 );
+
+			new PoemResultValidator( poems )
+				.validateAndRemove( result )
```
DavideD: I don't think I like this approach. The problem is that I don't understand what's going on and which value you expect.
Also, I don't know what kind of validation you are applying. Don't you think this is simpler and easier to understand?

```
assertThat( result ).containsOnly(portia, athanasa, imperatrix);
```
This will also check the size.

I woudl also return less results so that it's easier to figure out what's missing

DavideD: There is another problem with this test, if the query returns the results from the wrong page I don't think it will fail. For example, it returns the results from the second page instead of the first one

fax4ever: There is no order defined on the query, any result sequence should be be ok.

gsmet: I'm +1 with Davide. Add an order and test the results with `containsOnly()`. It's the only way to be sure everything works OK.

fax4ever: You're right. I'm going to push the fix.

fax4ever:
@DavideD @gsmet  I've updated the PR and squashed the commits on it. Thank you for the review

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBNativeQueryPaginationTest.java
```diff
@@ -25,21 +26,21 @@
...
-	private final OscarWildePoem apologia = new OscarWildePoem( 5L, "Apologias ", "Oscar Wilde", 1881 );
+	private final OscarWildePoem apologia = new OscarWildePoem( 5L, "Apologias", "Oscar Wilde", 1881 );
 	private final OscarWildePoem easter = new OscarWildePoem( 6L, "Easter Day", "Oscar Wilde", 1881 );
 	private final OscarWildePoem rome = new OscarWildePoem( 7L, "Rome Unvisited", "Oscar Wilde", 1881 );
 	private final OscarWildePoem miniato = new OscarWildePoem( 8L, "San Miniato", "Oscar Wilde", 1881 );
 	private final OscarWildePoem liberty = new OscarWildePoem( 9L, "Sonnet to Liberty", "Oscar Wilde", 1881 );
 	private final OscarWildePoem vita = new OscarWildePoem( 10L, "Vita Nuova", "Oscar Wilde", 1881 );
 
-	private final OscarWildePoem[] poems = { portia, athanasia, imperatrix, intellectualis, apologia, easter, rome, miniato, liberty, vita };
+	private final OscarWildePoem[] poems = { intellectualis, apologia, athanasia, imperatrix, easter, portia, rome, miniato, liberty, vita };
```
DavideD: `intellectualis` before `apologias`?

fax4ever: yes the name of the poem intellectualis is "Amor Intellectualis"

DavideD:
Looks good but I'm going to change the name of the test methods. I don't want to have to rename the method if I change the starting postion or the maxResults in one of the tests :).

DavideD:
I've created a new pull request: https://github.com/hibernate/hibernate-ogm/pull/980

PR #979, opened 2018-03-01
OGM-359 Stored procedures for Mongo

DavideD:
[OGM-359](https://hibernate.atlassian.net/browse/OGM-359)

Follows #976 

While  I was checking the latest PR on stored procedures I noticed several issues and ended up fixing them and rebase the history. I think now it's in a state the we can merge. In particular,  this PR:

* Fix the logic by making it more similar to the one in ORM
* Removes commons-lang library to escape strings and created a utility method that hopefully will work
* Adds test cases for when an exception is raised
* NoSQLProcedureResultSetOutputImpl.java is now the same as the one in ORM. The problem with the one in ORM is that it is package private
* Clean up the exception messages, they are more user friendly now 

There are still a couple of tests to add:
* Check storeprocedures returning multiple entities
* Test case for stored procedures in MongoDB causing log.multipleDocumentReturnedByStoredProcedure

I'll create a separate issue for these latest couple of cases

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureCallMemento.java
+++ b/core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureCallMemento.java
@@ -1,0 +1,45 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.storedprocedure.impl;
+
+import java.util.Map;
+
+import org.hibernate.HibernateException;
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.procedure.ProcedureCall;
+import org.hibernate.procedure.ProcedureCallMemento;
+import org.hibernate.procedure.internal.NoSQLProcedureCallImpl;
+
+public class NoSQLProcedureCallMemento implements ProcedureCallMemento {
+
+	private final ProcedureCallMemento delegate;
+
+	public NoSQLProcedureCallMemento(ProcedureCallMemento delegate) {
+		this.delegate = delegate;
+	}
+
+	@Override
+	public ProcedureCall makeProcedureCall(SharedSessionContractImplementor session) {
+		return new NoSQLProcedureCallImpl( session, this );
+	}
+
+	@Override
+	public Map<String, Object> getHintsMap() {
+		return delegate.getHintsMap();
+	}
+
+	@SuppressWarnings("unchecked")
+	public <T> T unwrap(Class<T> cls) {
+		if ( cls.isInstance( delegate ) ) {
+			return (T) delegate;
+		}
+		if ( cls.isInstance( this ) ) {
+			return (T) this;
+		}
+		throw new HibernateException( "Cannot unwrap the following type: " + cls );
+	}
+}
--- a/core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureResultSetOutputImpl.java
+++ b/core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureResultSetOutputImpl.java
@@ -1,0 +1,54 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.storedprocedure.impl;
+
+import java.util.List;
+import java.util.function.Supplier;
+
+import org.hibernate.result.ResultSetOutput;
+
+/**
+ * Implementation of {@link ResultSetOutput} for OGM.
+ * <p>
+ * This is a copy of the implementation in ORM that is package private and therefore we cannot use.
+ *
+ * @see org.hibernate.result.internal.ResultSetOutputImpl
+ * @author Davide D'Alto
+ */
+class NoSQLProcedureResultSetOutputImpl implements ResultSetOutput {
+
+	private final Supplier<List<?>> resultSetSupplier;
+
+	public NoSQLProcedureResultSetOutputImpl(List<?> results) {
+		this.resultSetSupplier = () -> results;
+	}
+
+	public NoSQLProcedureResultSetOutputImpl(Supplier<List<?>> resultSetSupplier) {
+		this.resultSetSupplier = resultSetSupplier;
+	}
+
+	@Override
+	public boolean isResultSet() {
+		return true;
+	}
+
+	@Override
+	public List<?> getResultList() {
+		return resultSetSupplier.get();
+	}
+
+	@Override
+	public Object getSingleResult() {
+		final List<?> results = getResultList();
+		if ( results == null || results.isEmpty() ) {
+			return null;
+		}
+		else {
+			return results.get( 0 );
+		}
+	}
+}
--- a/core/src/main/java/org/hibernate/ogm/util/impl/StringHelper.java
+++ b/core/src/main/java/org/hibernate/ogm/util/impl/StringHelper.java
@@ -11,6 +11,7 @@ import java.util.Iterator;
 /**
  * Utility functions for dealing with strings.
  *
+ * @author Davide D'Alto
  * @author Emmanuel Bernard &lt;emmanuel@hibernate.org&gt;
  * @author Gunnar Morling
  */
@@ -77,5 +78,29 @@ public class StringHelper {
 			return sb.substring( separator.length() );
 		}
 		return "";
+	}
+
+	/**
+	 * If a text contains double quotes, escape them.
+	 *
+	 * @param text the text to escape
+	 * @return Escaped text or {@code null} if the text is null
+	 */
+	public static String escapeDoubleQuotesForJson(String text) {
+		if ( text == null ) {
+			return null;
+		}
+		StringBuilder builder = new StringBuilder( text.length() );
+		for ( int i = 0; i < text.length(); i++ ) {
+			char c = text.charAt( i );
+			switch ( c ) {
+				case '"':
+				case '\\':
+					builder.append( "\\" );
+				default:
+					builder.append( c );
+			}
+		}
+		return builder.toString();
 	}
 }
--- a/documentation/manual/src/main/asciidoc/modules/mongodb.asciidoc
+++ b/documentation/manual/src/main/asciidoc/modules/mongodb.asciidoc
@@ -3473,6 +3473,59 @@ List<Poem> result = session.createNativeQuery( nativeQuery ).addEntity( Poem.cla
 ----
 ====
 
+[[ogm-mongodb-stored-proc-native]]
+
+==== Server-side JavaScript and stored procedures
+
+[NOTE]
+====
+This is an experimental feature.
+====
+
+In MongoDB, it is possible to call server-side JavaScript as if it is a stored procedure.
+You can use the existing methods in JPA:
+
+.Calling server-side JavaScript with positional parameters
+====
+[source, JAVA]
+----
+  @Entity
+  public class Car {
+    @Id
+    private Integer id;
+
+    private String brand;
+
+    ...
+  }
+
+  EntityManager em = ...
+  StoredProcedureQuery storedProcedureQuery = em.createStoredProcedureQuery( "findMostExpensiveCars", Car.class );
+  storedProcedureQuery.registerStoredProcedureParameter( "year", Integer.class, ParameterMode.IN );
+  storedProcedureQuery.setParameter( "year", 1995 );
+  List<Car> cars = storedProcedureQuery.getResultList();
+----
+====
+
+This example will work assuming that there is a `findMostExpensiveCars` JavaScript function in MongoDB 
+and that the result of the function is a list of cars that can be mapped to the `Car` entity.
+
+.Calling server-side JavaScript with Hibernate OGM with positional parameters
+====
+[source, JSON]
+----
+{
+  "result" : [
+    { "id":1, "brand":"Bentley" },
+    { "id":2, "brand":"Maserati" },
+  ]
+}
+----
+====
+
+More details about server-side functions can be found in
+https://docs.mongodb.com/manual/core/server-side-JavaScript[the MongoDB reference documentation].
+
 ==== Hibernate Search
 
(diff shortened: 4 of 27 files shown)


DavideD:
@schernolyas maybe you can have a look and see if the changes I've applied make sense. Thanks

schernolyas:
Okey @DavideD . I will look to your version.

schernolyas:
@DavideD  I think ... we have to mark the feature as "experimental"

DavideD:
> @DavideD I think ... we have to mark the feature as "experimental"

Good point, I will

core/src/main/java/org/hibernate/ogm/util/impl/StringHelper.java
```diff
@@ -78,4 +79,31 @@ public static String join(Iterator<?> iterator, String separator) {
...
+
+	/**
+	 * Escape the following characters: ',",/,\
+	 * <p>
+	 * Example: ' becomes \'
+	 *
+	 * @param text the string to escape
+	 * @return the escaped text or {@code null} if text is {@code null}
+	 */
+	public static String escape(String text) {
```
gsmet: This is not a generic escaping utility so `StringHelper.escape()` is not a good choice.

So maybe I would keep it in the MongoDB module and at least call it `escapeJsonString()`.

By the way, are you sure we need to escape `/`? I find it a bit weird. I'm also wondering if we need to escape `"`, considering the string is wrapped into `'`. It might be the case, I'm not very familiar with MongoDB escaping rules but I think we should test it to be sure we are not escaping things too much. Unless it was already tested?

documentation/manual/src/main/asciidoc/modules/query.asciidoc
```diff
@@ -8,6 +8,7 @@ With Hibernate OGM, you have a few alternatives that should get you covered:
 * Use JPQL - only for simple queries for now
 * Use the NoSQL native query mapping the result as managed entities
 * Use Hibernate Search queries - primarily full-text queries
+* Use the stored procedure mapping the result as managed entities
```
gsmet: Maybe `Use stored procedures mapping the results to managed entities` would be better?

documentation/manual/src/main/asciidoc/modules/query.asciidoc
```diff
@@ -421,3 +422,73 @@ go check the https://docs.jboss.org/hibernate/search/{hibernate-search-major-min
 === Using the Criteria API
 
 At this time, we have not implemented support for the Criteria APIs (neither Hibernate native nor JPA).
+
+=== Using stored procedures
+
+Often you want the raw power of the underlying NoSQL query engine.
+Even if that costs you portability.
+
+Hibernate OGM addresses that problem by letting you express stored procedures (e.g. in Server-side JavaScript for MongoDB )
```
gsmet: * Let's drop the cap on `Server-side`, `server-side` it is.
 * `MongoDB )`, we should remove the space before the parenthesis

documentation/manual/src/main/asciidoc/modules/query.asciidoc
```diff
@@ -421,3 +422,73 @@ go check the https://docs.jboss.org/hibernate/search/{hibernate-search-major-min
...
+and map the result of these queries as mapped entities or return primitive result.
+
+In JPA, use `EntityManager.createStoredProcedureQuery` or `EntityManager.createNamedStoredProcedureQuery` .
+The first form accepts a result class if your result set maps the mapping definition of the entity.
+The second form accepts the name of a resultSetMapping
+and lets you customize how properties are mapped to columns by the query.
+You can also used a predefined named query which defines its result set mapping.
```
gsmet: s/used/use/

documentation/manual/src/main/asciidoc/modules/query.asciidoc
```diff
@@ -421,3 +422,73 @@ go check the https://docs.jboss.org/hibernate/search/{hibernate-search-major-min
...
+
+   // getters, setters ...
+
+}
+
+...
+
+javax.persistence.EntityManager em = ...
+
+StoredProcedureQuery storedProcedureQuery = em.createStoredProcedureQuery( "mostExpensiveCarsPerYear", Car.class );
```
gsmet: It's a minor thing but could we be consistent with the naming of the stored procedure? Either use underscores or camel case for both.

documentation/manual/src/main/asciidoc/modules/query.asciidoc
```diff
@@ -421,3 +422,73 @@ go check the https://docs.jboss.org/hibernate/search/{hibernate-search-major-min
...
+
+You can also use named parameters and `StoredProcedureQuery#getSingleResult()`.
+
+Check out each individual datastore chapter for more info
+on the specifics of the native query language mapping.
+
+[WARNING]
+====
+'OUT' and 'IN_OUT' parameters of stored procedures are not supported yet.
+Main reason of it is supported data storages are supports 'IN' parameters only.
```
gsmet: This sentence is weird.

Maybe `The main reason of it is that supported data storages only support 'IN' parameters` would be better?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1733,6 +1739,95 @@ private static ReadPreference getReadPreference(AssociationContext associationCo
...
+					response.getString( "codeName" ).getValue(), mce );
+		}
+	}
+
+	private StringBuilder createCallStoreProcedureCommand(String storedProcedureName, ProcedureQueryParameters params) {
+		StringBuilder commandLine = new StringBuilder( storedProcedureName ).append( "(" );
+		List<Object> positionalParameters = params.getPositionalParameters();
+		for ( Object paramValue : positionalParameters ) {
+			if ( paramValue instanceof String ) {
+				//need for escape char "'"
```
gsmet: Depending on what we decide needs escaping, we should update this comment.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/logging/impl/Log.java
```diff
@@ -134,4 +136,12 @@
...
 	HibernateException invalidGeoJsonType(String actualType, String expectedType);
 
+	@Message(id = 1238, value = "Unable to execute command \"%s\". Error message : %s. Code name: %s")
+	HibernateException unableToExecuteCommand(String command, String errorMessage,String codeMessage, @Cause Exception e);
+
+	@Message(id = 1239, value = "Dialect %s does not support named parameters when calling stored procedures")
+	HibernateException dialectDoesNotSupportNamedParametersForStoredProcedures( @FormatWith(ClassObjectFormatter.class) Class<?> dialectClass );
+
+	@Message(id = 1240, value = "Procedures returning muliple documents are not supported. Procedure '%1$s' returned %2$d results")
+	HibernateException multipleDocumentReturnedByStoredProcedure(String storedProcedureName, int size);
```
gsmet: We usually have a . at the end of the log messages (this remark is for all the messages above).

+ s/muliple/multiple/

documentation/manual/src/main/asciidoc/modules/mongodb.asciidoc
```diff
@@ -3473,6 +3473,55 @@ List<Poem> result = session.createNativeQuery( nativeQuery ).addEntity( Poem.cla
 ----
 ====
 
+[[ogm-mongodb-stored-proc-native]]
+
+==== Server-side JavaScript and stored procedures
+
+In MongoDB, it is possible to call server side javascripts as if it is a stored procedure.
```
gsmet: * In the main doc, you call it `server-side JavaScript`, let's be consistent about the case. Maybe use `server-side JavaScript scripts` here? It looks redundant but I think it's the way to go.
 * `as if they were stored procedures`

documentation/manual/src/main/asciidoc/modules/mongodb.asciidoc
```diff
@@ -3473,6 +3473,55 @@ List<Poem> result = session.createNativeQuery( nativeQuery ).addEntity( Poem.cla
...
+You can use the existing methods in JPA:
+
+.Calling server side javascript with Hibernate OGM with positional parameters
```
gsmet: same here `server-side JavaScript`

I would drop the `with Hibernate OGM`. It's 2 `with` and, well, we are in the Hibernate OGM documentation :).

documentation/manual/src/main/asciidoc/modules/mongodb.asciidoc
```diff
@@ -3473,6 +3473,55 @@ List<Poem> result = session.createNativeQuery( nativeQuery ).addEntity( Poem.cla
...
+    @Id
+    private Integer id;
+
+    private String brand;
+
+    ...
+  }
+
+  EntityManager em = ...
+  StoredProcedureQuery storedProcedureQuery = em.createStoredProcedureQuery( "findMostExpensiveCars", Car.class );
```
gsmet: Same here about the consistency remark below (sorry I reviewed the main doc first so the comments look a bit backward). Let's name the stored procedures either with underscores or camel case. (Mentioning it here so that it's also fixed if you choose the underscores).

documentation/manual/src/main/asciidoc/modules/mongodb.asciidoc
```diff
@@ -3473,6 +3473,55 @@ List<Poem> result = session.createNativeQuery( nativeQuery ).addEntity( Poem.cla
...
+  storedProcedureQuery.registerStoredProcedureParameter( "year", Integer.class, ParameterMode.IN );
+  storedProcedureQuery.setParameter( "year", 1995 );
+  List<Car> cars = storedProcedureQuery.getResultList();
+----
+====
+
+This example will work assuming that there is a javascript function in MongoDB defined
```
gsmet: "... there is a `findMostExpensiveCars` JavaScript function defined in MongoDB..." would be better

The word order at the end of the sentence could be improved too: "that can be mapped to the `Car` entity"

documentation/manual/src/main/asciidoc/modules/mongodb.asciidoc
```diff
@@ -3473,6 +3473,55 @@ List<Poem> result = session.createNativeQuery( nativeQuery ).addEntity( Poem.cla
...
+`findMostExpensiveCars` and that the result of the function is a list of cars that can be mapped
+by the entity `Car`.
+
+.Calling server side javascript with Hibernate OGM with positional parameters
```
gsmet: Same remark here about the title. But maybe the title should be changed anyway as it does not fit very well to what is there?

documentation/manual/src/main/asciidoc/modules/mongodb.asciidoc
```diff
@@ -3473,6 +3473,55 @@ List<Poem> result = session.createNativeQuery( nativeQuery ).addEntity( Poem.cla
...
+  "result" : [
+    { "id":1, "brand":"Bentley" },
+    { "id":2, "brand":"Maserati" },
+  ]
+}
+----
+====
+
+I will let the details about server side funtions to
+https://docs.mongodb.com/manual/core/server-side-javascript[the MongoDB reference documentation].
```
gsmet: Maybe "More details about server-side JavaScript functions can be found in the MongoDB reference documentation."

We usually avoid the `I` in documentation.

core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureCallMemento.java
```diff
@@ -0,0 +1,43 @@
...
+
+	public NoSQLProcedureCallMemento(ProcedureCallMemento delegate) {
+		this.delegate = delegate;
+	}
+
+	public ProcedureCall makeProcedureCall(SharedSessionContractImplementor session) {
+		return new NoSQLProcedureCallImpl( session, this );
+	}
+
+	public Map<String, Object> getHintsMap() {
```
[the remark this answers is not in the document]
DavideD: Right, thanks

core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureResultSetOutputImpl.java
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	@Override
+	public List<?> getResultList() {
+		return resultSetSupplier.get();
+	}
+
+	@Override
+	public Object getSingleResult() {
+		final List<?> results = getResultList();
```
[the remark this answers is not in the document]
DavideD: +1
Just, so you know, I copied and paste the class from ORM and just renamed it.

DavideD:
@gsmet 
I've applied the remarks and update the PR.
In the end I think we only need to escape double quotes. We can also add tests later if somebody has a use case. I also add a check to  verify if the string is a number or a text.
All this part would require more testing but I think it can be merged for now.

I also added a note in the documentation to highlight that this feature is experimental.

Cheers

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1733,6 +1742,114 @@ private static ReadPreference getReadPreference(AssociationContext associationCo
...
+	private void appendStoredProcedureParamValue(String storedProcedureName, StringBuilder commandLine, Object paramValue) {
+		if ( paramValue == null ) {
+			// I don't think this can happen but, just in case, null should be a valid JSON value
+			commandLine.append( "null" );
+		}
+		else {
+			if ( requiresQuotes( paramValue ) ) {
+				String escapedValue = StringHelper.escapeDoubleQuotesForJson( paramValue.toString() );
+				// I think we only have to worry about double quotes for String
+				commandLine.append( '"' ).append( escapedValue ).append( '"' );
```
gsmet: I think you also have to escape the `\` character. Otherwise a `\"` string will be changed to `"`.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/logging/impl/Log.java
gsmet: @DavideD the s/muliple/multiple/ change was not done.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
DavideD: I'm ont sure I follow, but I added a test and it seems to work
Feel free to add another one before making changes, please

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/979

I've refactored the code and squashed the commits to create a clean history.
Everything test works now.

gsmet:
@DavideD Nice, glad to see this in.

DavideD:
@gsmet Yes, it definitely took longer than expected.

I also added a test to return value without mapping the result and found another couple of problems :/

PR #980, opened 2018-03-05
OGM-1411 Test pagination

DavideD:
Folllows https://github.com/hibernate/hibernate-ogm/pull/977

@fax4ever can you look at this and if it works for you, merge it?

I started to rename the methods but I realize it would be nice to move the tests to core so that we can run the same tests for Neo4j and for JPQL.  I also refactored the code  a bit.

Thanks


fax4ever:
I like the idea to move the tests to core.
Thank you @DavideD. I'm going to check it.

core/src/test/java/org/hibernate/ogm/backendtck/queries/pagination/PaginationUseCases.java
```diff
@@ -0,0 +1,159 @@
...
+import org.hibernate.Session;
+import org.hibernate.ogm.utils.OgmTestCase;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+/**
+ * Test the pagination for native queries with MongoDB
```
fax4ever: this test is now generic, maybe we could remove the reference to the dialect

DavideD: Updated and pushed force

fax4ever:
I merge it. Thanks @DavideD

DavideD:
@fax4ever Thanks, did you merge it in the 5.3 branch as well?

fax4ever:
Of course. I'm going to backporting it to 5.3 branch.

PR #981, opened 2018-03-05
OGM-1411 Backporting MongoDB native query with pagination on aggregate

fax4ever:


mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -151,6 +145,12 @@
 import com.mongodb.client.model.UpdateOptions;
 import com.mongodb.client.result.DeleteResult;
 import com.mongodb.client.result.UpdateResult;
+import org.bson.Document;
+import org.bson.types.ObjectId;
+import org.parboiled.Parboiled;
+import org.parboiled.errors.ErrorUtils;
+import org.parboiled.parserunners.RecoveringParseRunner;
+import org.parboiled.support.ParsingResult;
```
DavideD: No need to change in this PR but make sure in the future to not reorder the imports unless for good reasons

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/981

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
DavideD: I've reverted it before merging

fax4ever: ok thank you

PR #982, opened 2018-03-07
OGM-1375 We shouldn't create partial entities from a native query

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1375
It should be back-portable to 5.3.1 (Hibernate OGM 5.3 branch).

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/util/impl/Log.java
+++ b/core/src/main/java/org/hibernate/ogm/util/impl/Log.java
@@ -304,4 +304,7 @@ public interface Log extends BasicLogger {
 
 	@Message(id = 89, value = "%1$s does not support queries on polymorphic entities using TABLE_PER_CLASS inheritance strategy. You should try using SINGLE_TABLE instead. Entities: %2$s")
 	HibernateException queriesOnPolymorphicEntitiesAreNotSupportedWithTablePerClass( String datastore, Collection<String> subclassEntityNames );
+
+	@Message(id = 90, value = "addEntity is not allowed in native queries using projection on root Entity")
+	HibernateException addEntityNotAllowedInNativeQueriesUsingProjection();
 }
--- a/core/src/main/java/org/hibernate/ogm/util/impl/Log.java
+++ b/core/src/main/java/org/hibernate/ogm/util/impl/Log.java
@@ -305,6 +305,6 @@ public interface Log extends BasicLogger {
 	@Message(id = 89, value = "%1$s does not support queries on polymorphic entities using TABLE_PER_CLASS inheritance strategy. You should try using SINGLE_TABLE instead. Entities: %2$s")
 	HibernateException queriesOnPolymorphicEntitiesAreNotSupportedWithTablePerClass( String datastore, Collection<String> subclassEntityNames );
 
-	@Message(id = 90, value = "addEntity is not allowed in native queries using projection on root Entity")
-	HibernateException addEntityNotAllowedInNativeQueriesUsingProjection();
+	@Message(id = 90, value = "Projection and addEntity are not allowed in the same query: table <%1$s>.")
+	HibernateException addEntityNotAllowedInNativeQueriesUsingProjection( String table );
 }
--- a/core/src/test/java/org/hibernate/ogm/backendtck/queries/projection/JPQLProjectionTest.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/queries/projection/JPQLProjectionTest.java
@@ -6,7 +6,6 @@
  */
 package org.hibernate.ogm.backendtck.queries.projection;
 
-
 import static org.fest.assertions.Assertions.assertThat;
 
 import java.util.Collections;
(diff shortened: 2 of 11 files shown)


core/src/main/java/org/hibernate/ogm/util/impl/Log.java
```diff
@@ -304,4 +304,7 @@
 
 	@Message(id = 89, value = "%1$s does not support queries on polymorphic entities using TABLE_PER_CLASS inheritance strategy. You should try using SINGLE_TABLE instead. Entities: %2$s")
 	HibernateException queriesOnPolymorphicEntitiesAreNotSupportedWithTablePerClass( String datastore, Collection<String> subclassEntityNames );
+
+	@Message(id = 90, value = "addEntity is not allowed in native queries using projection on root Entity")
```
DavideD: Is it possible to give some additional information related to the query and/or which entity caused the problem?

I'm thinking about a user that it is reading a Log and suddenly see this message without knowing for sure which of the hundred queries caused the problem.

If we don't have the additional info that's OK.

I would prefer if the message were something similar to:
```
Projection and addEntity are  not allowed in the same query. Entity <org.hib.Pippo> with query <select p from ... > 
```

core/src/test/java/org/hibernate/ogm/backendtck/queries/projection/Poem.java
```diff
@@ -0,0 +1,85 @@
...
+	private Integer year;
+
+	public Poem() {
+	}
+
+	public Poem(Long id, String name, String author, int year) {
+		this.id = id;
+		this.name = name;
+		this.author = author;
+		this.year = year;
```
DavideD: I prefer if we don't have fields that aren't used.
I might be wrong, but it seems you are only interested to `id` and `name` for the tests

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBProjectionTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+	@Test
+	@TestForIssue( jiraKey = "OGM-1375" )
+	public void testResultListWithAddEntityWithProjection() {
+		thrown.expect( PersistenceException.class );
+		thrown.expectMessage( "OGM000090: addEntity is not allowed in native queries using projection on root Entity" );
+
+		inTransaction( session -> {
+			session.createNativeQuery( NATIVE_QUERY_WITH_PROJECTION )
+				.addEntity( org.hibernate.ogm.backendtck.queries.projection.Poem.class )
```
DavideD: Any reason to use the package as well here?

core/src/test/java/org/hibernate/ogm/backendtck/queries/projection/JPQLProjectionTest.java
```diff
@@ -0,0 +1,69 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.backendtck.queries.projection;
+
+
```
DavideD: Just one empty line, please

core/src/test/java/org/hibernate/ogm/backendtck/queries/projection/JPQLProjectionTest.java
```diff
@@ -0,0 +1,69 @@
...
+import static org.fest.assertions.Assertions.assertThat;
+
+import java.util.Collections;
+import java.util.List;
+
+import org.junit.Test;
+
+
```
DavideD: Just one empty line, please

core/src/test/java/org/hibernate/ogm/backendtck/queries/projection/Poem.java
fax4ever: Other fields are present to remark the concept of partial Entity. I'd rather leave them.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBProjectionTest.java
fax4ever: no, it's a typo

core/src/test/java/org/hibernate/ogm/backendtck/queries/projection/Poem.java
DavideD: Make sense.

I wonder if we can find a way to make it a bit less obscure. Somebody might delete them in the future.

Maybe you can add a test to project only this two fields? So we have a test case that doesn't return the id. In the end the id is sort of a special field.

fax4ever: I do not know because actually all tests without projection work on equals and hashcode, that in turn use all fields.

DavideD: I can understand that **technicallly** they are used, But I could remove them in the Poem class and I wouldn't have any problem with our test suit. If somebody looks at this code how can he guess that this fields are actually part of the test suit?

At least leave a comment.

fax4ever: ok I'm making sure that some tests use some, other tests other fields, so you do not have fields not directly used

fax4ever:
With this second commit: error is thrown in the same way also using Neo4J embedded dialect. Unfortunately, in this second case, validation is done only after the query has been executed and not before: at result set extraction time. This depends on that native query on neo4j is never parsed by the dialect as it happens instead in MondoDB.

DavideD:
@fax4ever  I'm going to close this one since you are still applying some changes. Feel free to re-open i it when it's ready or send a new PR (whatever you prefer).

PR #983, opened 2018-03-08
OGM-1418 Fix README.md about how to run the tests on a Neo4j server

DavideD:
https://hibernate.atlassian.net/browse/OGM-1418

@fax4ever Easy one to review

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,9 +141,14 @@ prior to running the tests:
 For running the tests in the _neo4j_ and _integrationtest/neo4j_ modules, by default the
 embedded Neo4j configuration is used.
 
-If you want to run the tests on a remote server, you need to specify the profile `neo4j-remote`
+Hibernate OGM can connect to a Neo4j server remotely via the Bolt protocol or via HTTP.
+These tests are not executed by default, you can run them using the following commands:
 
-    mvn clean install -s settings-example.xml -Pneo4j-remote
+    mvn clean install -s settings-example.xml -Pneo4j-http
+
+or
+
+    mvn clean install -s settings-example.xml -Pneo4j-bolt
 
 This assumes Neo4j to be installed on `localhost`, using the default port and no authentication.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -141,9 +141,14 @@ prior to running the tests:
 For running the tests in the _neo4j_ and _integrationtest/neo4j_ modules, by default the
 embedded Neo4j configuration is used.
 
-If you want to run the tests on a remote server, you need to specify the profile `neo4j-remote`
+Hibernate OGM can connect to a Neo4j server remotely via the Bolt protocol or via HTTP.
+These tests are not executed by default, you can run them using the following commands:
```
fax4ever: I would prefer giving more emphasis to the concept:
"
Hibernate OGM can connect to a Neo4j server remotely via the Bolt protocol or via HTTP.

__Important:__ These tests are __not executed__ by default. 
But are executed by all Hibernate OGM DevOps pipelines.
So before contribute **make sure** that all these tests passes too.

You can run them using the following commands:
"

DavideD: I don't want to put off contributors making them think that they have to download and install a Neo4j server when they only want to contribute to MongoDB, for example.

If, in the future, people will send a lot of PR failing for the Neo4j remote server, we can update it.
Right now, this is more a place to remind people which profiles exist in the project. 

One thing that we should do, though, is to update the contribute page on hibernate.org and add a link in the README (if it's not there already). But we can do this as a separate issue

README.md
```diff
@@ -141,9 +141,14 @@ prior to running the tests:
...
 
-    mvn clean install -s settings-example.xml -Pneo4j-remote
+    mvn clean install -s settings-example.xml -Pneo4j-http
```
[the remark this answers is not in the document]
DavideD: I guess, but one still need to run the integration-test and the jboss modules project.

PR #984, opened 2018-03-09
OGM-1375 We shouldn't create partial entities from a native query

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1375

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/util/impl/Log.java
+++ b/core/src/main/java/org/hibernate/ogm/util/impl/Log.java
@@ -307,4 +307,7 @@ public interface Log extends BasicLogger {
 
 	@Message(id = 90, value = "Returning multiple entities is not supported. Procedure '%1$s' expects results of type %2$s")
 	HibernateException multipleEntitiesOutputNotSupported(String procedureName, Collection<?> synchronizedQuerySpaces);
+
+	@Message(id = 91, value = "Projection and addEntity are not allowed in the same query: table <%1$s>.")
+	HibernateException addEntityNotAllowedInNativeQueriesUsingProjection( String table );
 }
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
@@ -849,8 +849,13 @@ public class MongoDBDialect extends BaseGridDialect implements QueryableGridDial
 		MongoDBQueryDescriptor queryDescriptor = backendQuery.getQuery();
 
 		EntityKeyMetadata entityKeyMetadata =
-				backendQuery.getSingleEntityMetadataInformationOrNull() == null ? null :
-					backendQuery.getSingleEntityMetadataInformationOrNull().getEntityKeyMetadata();
+			( backendQuery.getSingleEntityMetadataInformationOrNull() == null ) ? null :
+				backendQuery.getSingleEntityMetadataInformationOrNull().getEntityKeyMetadata();
+
+		// Projections and addEntities are not allowed in the same query at the same time
+		if ( entityKeyMetadata != null && queryDescriptor.getProjection() != null ) {
+			throw log.addEntityNotAllowedInNativeQueriesUsingProjection( entityKeyMetadata.getTable() );
+		}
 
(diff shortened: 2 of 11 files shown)


neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/query/nativequery/Neo4jProjectionTest.java
```diff
@@ -0,0 +1,59 @@
...
+			List<Object> movies = session.createNativeQuery( getNativeQueryWithProjectionYearAuthor() )
+				.getResultList();
+
+			assertThat( movies ).hasSize( 1 );
+			Object[] movie = (Object[]) movies.get( 0 );
+
+			// actually Neo4J remote http return always Integer type
+			// while Neo4J remote bolt return always Long type
+			// regardless the fact that type declared on the Entity is an Integer or a Long
+			assertThat( ( (Number) movie[0] ).intValue() ).isEqualTo( 1968 );
```
fax4ever: @DavideD notice here that I have to change the test in order to work with problem of returning number type of Neo4J HTTP and Bolt. I believe we should explore it in another issue

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BoltNeo4jDialect.java
```diff
@@ -118,7 +119,14 @@ protected BoltNeo4jEntityQueries createNeo4jEntityQueries(EntityKeyMetadata enti
 				List<EntityKey> entityKeys = new ArrayList<>();
 				while ( result.hasNext() ) {
 					Record record = result.next();
-					Map<String, Object> recordAsMap = record.get( 0 ).asMap();
+					Value value = record.get( 0 );
+
+					if ( !( "NODE".equals( value.type().name() ) ) ) {
```
DavideD: I think you should use the constant here: InternalTypeSystem.TYPE_SYSTEM.NODE()

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -849,8 +849,13 @@ public void forEachTuple(ModelConsumer consumer, TupleTypeContext tupleTypeConte
 		MongoDBQueryDescriptor queryDescriptor = backendQuery.getQuery();
 
 		EntityKeyMetadata entityKeyMetadata =
-				backendQuery.getSingleEntityMetadataInformationOrNull() == null ? null :
-					backendQuery.getSingleEntityMetadataInformationOrNull().getEntityKeyMetadata();
+			( backendQuery.getSingleEntityMetadataInformationOrNull() == null ) ? null :
+				backendQuery.getSingleEntityMetadataInformationOrNull().getEntityKeyMetadata();
```
DavideD: It's not critical, but I prefer this format for this:

```
		backendQuery.getSingleEntityMetadataInformationOrNull() == null
                                ? null
				: backendQuery.getSingleEntityMetadataInformationOrNull().getEntityKeyMetadata();
```

I find it more readable

core/src/main/java/org/hibernate/ogm/util/impl/Log.java
```diff
@@ -307,4 +307,7 @@
 
 	@Message(id = 90, value = "Returning multiple entities is not supported. Procedure '%1$s' expects results of type %2$s")
 	HibernateException multipleEntitiesOutputNotSupported(String procedureName, Collection<?> synchronizedQuerySpaces);
+
+	@Message(id = 91, value = "Projection and addEntity are not allowed in the same query: table <%1$s>.")
+	HibernateException addEntityNotAllowedInNativeQueriesUsingProjection( String table );
```
DavideD: I'm not sure if this message is that helpful, in the end it's not the table the problem but the query

fax4ever: Unfortunately in the EmbeddedNeo4jBackendQueryResultIterator.java class we don't have the orginal query text.

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/984

In the end I slightly changed the message log so that if we can print the query we print it otherwise we don't.

@fax4ever Have a look, if you you see some issues with it we can add a commit to fix it

I've merged in 5.4 and 5.3.1

Thanks

PR #985, opened 2018-03-09

QwerTech:
Сontinuation of https://github.com/hibernate/hibernate-ogm/pull/968
added EmptyMultiTableBulkIdStrategy
switch dialect to EmptyMultiTableBulkIdStrategy
extend TableIdGeneratorInheritanceTest with delete and update tests


core/src/test/java/org/hibernate/ogm/backendtck/id/TableIdGeneratorInheritanceTest.java
```diff
@@ -0,0 +1,158 @@
...
+		doInTransaction( em -> em.persist( truck ) );
+
+		doInTransaction( em -> {
+			Truck loadedTruck = em.find( Truck.class, truck.getId() );
+			em.remove( loadedTruck );
+		} );
+
+		doInTransaction( em -> {
+			Truck loadedTruck = em.find( Truck.class, truck.getId() );
+			assertThat( loadedTruck ).isNull();
```
DavideD: This test is not enough.
The fact that the entity is null doesn't mean that all the collections involved are empty. I think you can only check this with native queries.

core/src/main/java/org/hibernate/ogm/dialect/impl/EmptyMultiTableBulkIdStrategy.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.hibernate.boot.spi.MetadataImplementor;
+import org.hibernate.boot.spi.SessionFactoryOptions;
+import org.hibernate.engine.jdbc.connections.spi.JdbcConnectionAccess;
+import org.hibernate.engine.jdbc.spi.JdbcServices;
+import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.hql.internal.ast.HqlSqlWalker;
+import org.hibernate.hql.spi.id.MultiTableBulkIdStrategy;
+
+public class EmptyMultiTableBulkIdStrategy implements MultiTableBulkIdStrategy {
```
DavideD: I can see that this is working but I would like to understand why. A bit of description would be helpful

core/src/test/java/org/hibernate/ogm/backendtck/id/TableIdGeneratorInheritanceTest.java
QwerTech: Do I understand correctly: should I check through a native request that the truck table is empty after removal?
Should I check that base table is empty and not exists?

DavideD: Mmmh .. let me check, maybe the mapping is not what I'm expecting

core/src/main/java/org/hibernate/ogm/dialect/impl/EmptyMultiTableBulkIdStrategy.java
QwerTech: Ok, I'll see deeper. But in the first look, MultiTableBulkIdStrategy is used while multi tables updates or deletions. But in my case there is single table.

core/src/test/java/org/hibernate/ogm/backendtck/id/TableIdGeneratorInheritanceTest.java
DavideD: Right, my bad. We don't create multiple tables in OGM for the same entity.
I will create a follow up issue to tihs one to add some test to check for the mapping in MongoDB for TABLE_PER_CLAS hierarchies, it seems that right now we only check for the single table per class case (SingleTableInheritanceTest)

core/src/main/java/org/hibernate/ogm/dialect/impl/EmptyMultiTableBulkIdStrategy.java
DavideD: After looking at it I realise that we don't create  in muliple tables in OGM for the same entity.
The reason is because we don't have join for all the datastores. Still, not everybody i sfamiliar with the codebase and there should be some javadoc in places where we have this kind of magic.

QwerTech: What do you mean?

core/src/test/java/org/hibernate/ogm/backendtck/id/TableIdGeneratorInheritanceTest.java
QwerTech: Is that enough of the test in a result?

DavideD:
@QwerTech Can can you rebase this PR so that it only has three commits? 

1st commit: Add core/src/main/java/org/hibernate/ogm/dialect/impl/EmptyMultiTableBulkIdStrategy.java
2nd commit: Changes to core/src/main/java/org/hibernate/ogm/dialect/impl/OgmDialect.java
3rd commit: Add core/src/test/java/org/hibernate/ogm/backendtck/id/TableIdGeneratorInheritanceTest.java


Or, if it's OK for you, I can do that but your user won't appear in the commit history?

QwerTech:
Ok, I will do it tomorrow

PR #987, opened 2018-03-13
OGM-1047 using database pool

schernolyas:
https://hibernate.atlassian.net/browse/OGM-1047

### DIFF
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
@@ -157,7 +157,7 @@ public class EmbeddedNeo4jGraphDatabaseFactory implements GraphDatabaseServiceFa
 						LOG.debugf( " currentTransaction: %s", spi.currentTransaction() );
 						//LOG.warnf( "close transaction %s", spi.currentTransaction().getTransactionId() );
 						LOG.warnf( "transaction  started %s", spi.currentTransaction().startTime() );
-						spi.currentTransaction().failure();
+						spi.currentTransaction().success();
 						spi.currentTransaction().closeTransaction();
 					}
 				}
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
@@ -159,16 +159,16 @@ public class EmbeddedNeo4jGraphDatabaseFactory implements GraphDatabaseServiceFa
 					if ( spi.isInOpenTransaction() ) {
 						LOG.warnf( "currentTransaction: %s", spi.currentTransaction() );
 						LOG.warnf( "transaction started %s", spi.currentTransaction().startTime() );
-						LOG.warn( "hang transactions  will commited and closed .." );
-						spi.currentTransaction().success();
+						spi.currentTransaction().failure();
 						spi.currentTransaction().closeTransaction();
+						throw LOG.existsOpenTransactionsFornDatabasePathException( key );
 					}
 				}
 				catch ( IllegalAccessException | NoSuchFieldException e) {
 					throw new HibernateException( "Can't get value of private variable by reflection!",e );
 				}
 				catch (TransactionFailureException e) {
-					throw new HibernateException( "Can't close transaction!",e );
+					throw new HibernateException( "Can't close transction!",e );
 				}
 
 
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
@@ -7,16 +7,22 @@
 package org.hibernate.ogm.datastore.neo4j.embedded.impl;
 
 import java.io.File;
+import java.lang.invoke.MethodHandles;
 import java.net.URL;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.hibernate.ogm.datastore.neo4j.Neo4jProperties;
+import org.hibernate.ogm.datastore.neo4j.logging.impl.Log;
+import org.hibernate.ogm.datastore.neo4j.logging.impl.LoggerFactory;
 import org.hibernate.ogm.datastore.neo4j.spi.GraphDatabaseServiceFactory;
 import org.hibernate.ogm.util.configurationreader.spi.ConfigurationPropertyReader;
+
 import org.neo4j.graphdb.GraphDatabaseService;
 import org.neo4j.graphdb.factory.GraphDatabaseBuilder;
 import org.neo4j.graphdb.factory.GraphDatabaseFactory;
+import org.neo4j.kernel.internal.GraphDatabaseAPI;
 
 /**
  * Contains methods to create a {@link GraphDatabaseService} for the embedded Neo4j.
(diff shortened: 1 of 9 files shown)


neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
```diff
@@ -64,32 +66,33 @@ public GraphDatabaseService create() {
 		AtomicBoolean isNew = new AtomicBoolean( false );
 		FactoryHolder factoryHolder = GRAPH_DATABASE_SERVICE_MAP.computeIfAbsent( dbLocationAbsolutePath, ( String dbPath ) -> {
 			LOG.infof( " Create new service instance for dbPath  %s", dbLocationAbsolutePath );
-			GraphDatabaseService service = buildGraphDatabaseService();
+			GraphDatabaseFacade service = buildGraphDatabaseService();
 			FactoryHolder holder = new FactoryHolder();
 			holder.setCounter( 1 );
-			holder.setGraphDatabaseService( service );
+			holder.setGraphDatabaseFacade( service );
```
DavideD: When you send a PR don't rename a method and add changes in the same commit. It makes it harder to figure out where the important changes in this commit are

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
```diff
@@ -64,32 +66,33 @@ public GraphDatabaseService create() {
...
 			isNew.set( true );
 			return holder;
 		} );
 		if ( !isNew.get() ) {
 			synchronized (GRAPH_DATABASE_SERVICE_MAP) {
-				boolean isAvailable = factoryHolder.getGraphDatabaseService().isAvailable( 100 );
+				factoryHolder.getGraphDatabaseFacade();
+				boolean isAvailable = factoryHolder.getGraphDatabaseFacade().isAvailable( 100 );
 				if ( isAvailable ) {
```
DavideD: What's 100?

DavideD:
Could you explain in the description what's changes compared to the last time? How did you close the transactions? Why do you consider this a preview? What's missing?

schernolyas:
Hi @DavideD !

I found reason why tests failed. I commit all uncommitted transactions in 'shutdown' phase. Why test have uncommitted transactions is separate question. Now all tests works fine.

About why the PR has status 'preview'. The PR has some developer's log :-(. When I fix all your comments I will prepare 'ready for merge' PR (or correct the PR) with clear commit history.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
schernolyas: Ok. I will try.

DavideD:
> I found reason why tests failed. I commit all uncommitted transactions in 'shutdown' phase

Mmh... that doesn't seem right, for real applications using the library we need to throw an exception if something is not commited. The library cannot just commit transaction because we need to close the factory. Also It's suspicious that the build works without this PR.

schernolyas:
Hibernate OGM update neo4j version. May be ... it affected. I not looked to history.

About "for real applications using the library we need to throw an exception if something is not commited". You right. it is exception. Now I can found tests what leave transactions by log.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
schernolyas: It is time of checking availability in milliseconds.

DavideD: It's better not to have Magic numbers like this in the codebase. This should be a constant called something like WAITING_TIME_MS

schernolyas: fixed in next commit. A lot of thanks!

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jSchemaDefiner.java
```diff
@@ -59,6 +64,42 @@ protected void createUniqueConstraintsIfMissing(DatastoreProvider provider, List
 		EmbeddedNeo4jDatastoreProvider neo4jProvider = (EmbeddedNeo4jDatastoreProvider) provider;
 		GraphDatabaseService neo4jDb = neo4jProvider.getDatabase();
 		Transaction tx = neo4jDb.beginTx();
+		if ( tx instanceof TopLevelTransaction ) {
```
schernolyas: The code can be removed. It is for debug and code investigations only.

DavideD: I don't think we should care about the Transaction specific implementation.

schernolyas: Yes, @DavideD . I will remove it. The code is not needs for 'ready for merge' PR. The code is for debug and investigation only.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jSchemaDefiner.java
```diff
@@ -64,42 +59,6 @@ protected void createUniqueConstraintsIfMissing(DatastoreProvider provider, List
...
-			}
-			catch (IllegalStateException ise) {
-				log.warn( "New transction created!" );
-			}
-			catch (Exception e) {
-				log.error( "ERROR!", e );
-			}
-		}
-
-
```
DavideD: Are you telling me that all this code was to just log something? Maybe next time, at least, put it in a separate method so that the purpose is clear, please.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
```diff
@@ -151,20 +153,24 @@ static void shutdownGraphDatabaseService(GraphDatabaseService neo4jDb) {
...
 					LOG.debugf( " isInOpenTransaction: %s", spi.isInOpenTransaction() );
 					if ( spi.isInOpenTransaction() ) {
-						LOG.debugf( " currentTransaction: %s", spi.currentTransaction() );
-						//LOG.warnf( "close transaction %s", spi.currentTransaction().getTransactionId() );
-						LOG.warnf( "transaction  started %s", spi.currentTransaction().startTime() );
+						LOG.warnf( "currentTransaction: %s", spi.currentTransaction() );
+						LOG.warnf( "transaction started %s", spi.currentTransaction().startTime() );
+						LOG.warn( "hang transactions  will commited and closed .." );
 						spi.currentTransaction().success();
 						spi.currentTransaction().closeTransaction();
```
DavideD: This code is still closing a transaction when the factory is shut down. You are  coimmitting a transaction under the hood without a user knowing about it. What if the transaction is supposed to be rolled back because something else needs to be done for the user workflow to be correct? Also the transaction could already be in a `failure()` state but you are resetting it with `success()`. I mean, maybe I'm missing something here but this does not seem correct.

schernolyas: Okey @DavideD . I will throw Exception .

schernolyas: fixed.

DavideD:
@schernolyas Before I start the review, could you squash together commits in a more meaningful way?

Specfically, I would like not to have a commit like c116975 and then see two commits later that it gets removed by b9edc27.

I'm talking also about  cc0b755, 1b602d2  and so on

If the purpose of a commit is to amend the previous commit, please, just squash them together.

Let me know when you have done.

Thanks

schernolyas:
Hi @DavideD !
I prepared new clear PR https://github.com/hibernate/hibernate-ogm/pull/1005 .

PR #990, opened 2018-03-16

QwerTech:
Сontinuation of #968 and #985
added EmptyMultiTableBulkIdStrategy
switch dialect to EmptyMultiTableBulkIdStrategy
extend TableIdGeneratorInheritanceTest with delete and update tests


DavideD:
Thanks, this is much better

DavideD:
I've applied some changes and merged https://github.com/hibernate/hibernate-ogm/pull/990

Thanks a lot.

Please @QwerTech have a look at the commits on master to have an idea of what I mean better commit messages. When one reads the history log needs to get an idea of what the changes are about not which files get inlcuded. I've also refactored the tests a bit and moved the doTransaction method in the superclass so that other classes can use it.

PR #994, opened 2018-03-26
OGM-1434 Test case for MongoDB: ObjectId passed as parameter to query

DavideD:
https://hibernate.atlassian.net/browse/OGM-1434

### DIFF
--- a/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/id/objectid/Bar.java
+++ b/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/id/objectid/Bar.java
@@ -7,6 +7,7 @@
 package org.hibernate.ogm.datastore.mongodb.test.id.objectid;
 
 import java.util.HashSet;
+import java.util.Objects;
 import java.util.Set;
 
 import javax.persistence.CascadeType;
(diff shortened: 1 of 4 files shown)


mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/id/objectid/Bar.java
```diff
@@ -18,6 +18,7 @@
 import javax.persistence.OneToMany;
 
 import org.bson.types.ObjectId;
+import org.jboss.logging.processor.util.Objects;
```
gsmet: I don't think this is the one you want to use :)

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/id/objectid/DoorMan.java
```diff
@@ -12,6 +12,7 @@
 import javax.persistence.Id;
 
 import org.bson.types.ObjectId;
+import org.jboss.logging.processor.util.Objects;
```
gsmet: Same here.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/id/objectid/MusicGenre.java
```diff
@@ -16,6 +16,7 @@
 import javax.persistence.OneToMany;
 
 import org.bson.types.ObjectId;
+import org.jboss.logging.processor.util.Objects;
```
gsmet: Ditto.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/id/objectid/Bar.java
DavideD: Damn! :)
Thanks

Sanne: Add it to the checkstyle blacklist? We did similar things for e.g. to ban the deprecated junit helpers.

DavideD:
Fixed the imports, Is it OK if I merge it @gsmet?

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/id/objectid/Bar.java
```diff
@@ -72,4 +73,31 @@ public void setMusicGenre(MusicGenre musicGenre) {
...
 		this.doorMen = doorMen;
 	}
+
+	@Override
+	public int hashCode() {
+		final int prime = 31;
+		int result = 1;
+		result = prime * result + ( ( doorMen == null ) ? 0 : doorMen.hashCode() );
+		result = prime * result + ( ( musicGenre == null ) ? 0 : musicGenre.hashCode() );
+		result = prime * result + ( ( name == null ) ? 0 : name.hashCode() );
```
gsmet: Note that in the tests, you can also use `Objects#hash(Object...)`. Not that I require you to change it here, just for next time.

It's not always recommended in our internal code as it allocates an array but for the tests, I think it's good enough.

gsmet:
Merged, thanks!

PR #1000, opened 2018-04-06
OGM-1429 Support run java code for embedded mode (#42)

ntviet18:


infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/InfinispanDialect.java
```diff
@@ -228,6 +247,104 @@ public void forEachTuple( ModelConsumer consumer, TupleTypeContext tupleTypeCont
 		return (KeyProvider<EK, AK, ISK>) provider.getKeyProvider();
 	}
 
+	@Override
+	public ClosableIterator<Tuple> callStoredProcedure( String storedProcedureName, ProcedureQueryParameters queryParameters, TupleContext tupleContext ) {
+		validate( queryParameters );
+		EmbeddedCacheManager embeddedCacheManager = getCacheManager().getCacheManager();
+		Cache<String, String> cache = embeddedCacheManager.getCache( "___stored_procedures", true );
```
fax4ever: Maybe we can extract "___stored_procedures" constants, used in API to register the alias procedure name.

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/InfinispanDialect.java
```diff
@@ -228,6 +247,104 @@ public void forEachTuple( ModelConsumer consumer, TupleTypeContext tupleTypeCont
...
+
+	private void validate(ProcedureQueryParameters queryParameters) {
+		List<Object> positionalParameters = queryParameters.getPositionalParameters();
+		if ( positionalParameters != null && positionalParameters.size() > 0 ) {
+			throw log.dialectDoesNotSupportPositionalParametersForStoredProcedures( getClass() );
+		}
+	}
+
+	private Object execute(String storedProcedureName, EmbeddedCacheManager embeddedCacheManager, Callable<?> callable) {
+		AtomicReference<Object> ref = new AtomicReference<>();
```
fax4ever: Why use an atomic reference without compare and set. Maybe it is my miss :)

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/InfinispanDialect.java
```diff
@@ -228,6 +247,104 @@ public void forEachTuple( ModelConsumer consumer, TupleTypeContext tupleTypeCont
 		return (KeyProvider<EK, AK, ISK>) provider.getKeyProvider();
 	}
 
+	@Override
+	public ClosableIterator<Tuple> callStoredProcedure( String storedProcedureName, ProcedureQueryParameters queryParameters, TupleContext tupleContext ) {
```
fax4ever: Maybe we can create a new class (stateless and thread safe) to delegate stored procedure invocation. 
Put in there callStoredProcedure, validate, execute, instantiate, setParams, execute, extractTuples, extractTuple, methos and ExceptionWrapper class. The collaborator class can be created in the constructor as final blank filed.

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/InfinispanDialect.java
fax4ever: @Sanne WDYT? thank you

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/util/ReflectionHelper.java
```diff
@@ -0,0 +1,102 @@
...
+import java.lang.reflect.Method;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Reflection utilities.
+ *
+ * @author The Viet Nguyen &amp;ntviet18@gmail.com&amp;
+ */
+public class ReflectionHelper {
```
Sanne: There is a class with the same name in ogm core: `org.hibernate.ogm.util.impl.ReflectionHelper`

Could you add new methods to that one maybe? I have some additional comments I'll add below.

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/util/ReflectionHelper.java
```diff
@@ -0,0 +1,102 @@
...
+	 * @param className full qualified class name
+	 *
+	 * @return new object of given class.
+	 *
+	 * @throws ClassNotFoundException if the class cannot be located
+	 * @throws IllegalAccessException if class or its no-arg constructor are not accessible
+	 * @throws InstantiationException if the class cannot be instantiated, e.g. abstract class, interface, etc.
+	 */
+	public static Object instantiate(String className) throws ClassNotFoundException, IllegalAccessException, InstantiationException {
+		Class<?> clazz = Class.forName( className );
```
Sanne: This would prevent OGM to work correctly in modular environments. When needing to create a new instance of a user domain model by name, we should use `ClassLoaderService`.

You can find several examples of using the service in the code base. For example `org.hibernate.ogm.service.impl.SessionFactoryEntityNamesResolver` does something similar.

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/InfinispanDialect.java
```diff
@@ -228,6 +247,104 @@ public void forEachTuple( ModelConsumer consumer, TupleTypeContext tupleTypeCont
...
+			return (Callable<?>) ReflectionHelper.instantiate( className );
+		}
+		catch (ClassCastException | ClassNotFoundException | IllegalAccessException | InstantiationException e) {
+			throw log.cannotInstantiateStoredProcedure( storedProcedureName, className, e );
+		}
+	}
+
+	private static void setParams( String storedProcedureName, ProcedureQueryParameters queryParameters, Callable<?> callable ) {
+		try {
+			ReflectionHelper.setFields( callable, queryParameters.getNamedParameters() );
```
Sanne: It would be very nice for people using it if the error being thrown could say exactly wich "setter" is expected, when a NoSuchMethodException is thrown. Could we improve the exception errors?

ntviet18:
@fax4ever, @Sanne thank you, please see the updates

fax4ever:
Thank you @ntviet18. For me it's ok. Wating for other feedbacks.

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/impl/InfinispanEmbeddedStoredProceduresDialect.java
```diff
@@ -0,0 +1,152 @@
...
+			Map<String, Object> introspect = ReflectionHelper.introspect( obj );
+			introspect.forEach( tuple::put );
+			return tuple;
+		}
+		catch (IntrospectionException | InvocationTargetException | IllegalAccessException e) {
+			throw log.cannotExtractStoredProcedureResultSet( storedProcedureName, obj, e );
+		}
+	}
+
+	private static class ExceptionWrapper extends RuntimeException {
```
DavideD: Why do you need an ExceptionWrapper? Can't you throw an HibernateException?

infinispan-embedded/src/test/java/org/hibernate/ogm/datastore/infinispan/test/storedprocedures/NotReallyAProcedure.java
```diff
@@ -0,0 +1,19 @@
...
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+
+package org.hibernate.ogm.datastore.infinispan.test.storedprocedures;
+
+/**
+ * @author The Viet Nguyen &amp;ntviet18@gmail.com&amp;
+ */
+public class NotReallyAProcedure implements Runnable {
```
DavideD: `NoOpProcedure` would be a better name

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/impl/InfinispanEmbeddedStoredProceduresDialect.java
```diff
@@ -0,0 +1,152 @@
...
+import org.hibernate.ogm.util.impl.CollectionHelper;
+import org.hibernate.ogm.util.impl.ReflectionHelper;
+
+import org.infinispan.Cache;
+import org.infinispan.manager.EmbeddedCacheManager;
+
+/**
+ * @author The Viet Nguyen &amp;ntviet18@gmail.com&amp;
+ */
+public class InfinispanEmbeddedStoredProceduresDialect {
```
DavideD: Let's not use the suffix `Dialect`  for this class. Dialect has already a defined meaning in OGM and `*Manager` or `*Caller` could be valid alternatives I can think on top of my head.

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/impl/InfinispanEmbeddedStoredProceduresDialect.java
ntviet18: Thank you, we can. I forgot cleanup this workaround from my previous implementation

infinispan-embedded/src/test/java/org/hibernate/ogm/datastore/infinispan/test/storedprocedures/NotReallyAProcedure.java
ntviet18: Should we support Runnable as a StoredProcedure? Currently, we don't support it. But it will make a valid procedure which return Void.

infinispan-embedded/src/main/java/org/hibernate/ogm/datastore/infinispan/impl/InfinispanEmbeddedStoredProceduresDialect.java
ntviet18: sure

infinispan-embedded/src/test/java/org/hibernate/ogm/datastore/infinispan/test/storedprocedures/InfinispanNamedParametersStoredProcedureCallTest.java
```diff
@@ -0,0 +1,119 @@
...
+import static org.hibernate.ogm.backendtck.storedprocedures.Car.RESULT_SET_PROC_ID_PARAM;
+import static org.hibernate.ogm.backendtck.storedprocedures.Car.RESULT_SET_PROC_TITLE_PARAM;
+
+/**
+ * Testing call of stored procedures
+ *
+ * @author The Viet Nguyen &amp;ntviet18@gmail.com&amp;
+ */
+@TestForIssue(jiraKey = { "OGM-1429" })
+public class InfinispanNamedParametersStoredProcedureCallTest extends OgmJpaTestCase {
```
DavideD: I don't think these test are Infinipan specific, couldn't you update `NamedParametersStoredProcedureCallTest`.
This way we can also check othere dialects.

infinispan-embedded/src/test/java/org/hibernate/ogm/datastore/infinispan/test/storedprocedures/NotReallyAProcedure.java
DavideD: I thought that the `NotReally` part referred to the fact that the StoreProcedure doesn't do anything.

No need to support Runnable, but let's call it `InvalidStoredProcedure`. I think it's clearer that way. Also a bit of javadoc can go a long way in these cases.

infinispan-embedded/src/test/java/org/hibernate/ogm/datastore/infinispan/test/storedprocedures/InfinispanNamedParametersStoredProcedureCallTest.java
```diff
@@ -0,0 +1,119 @@
...
+		thrown.expect( PersistenceException.class );
+		thrown.expectMessage( "org.hibernate.HibernateException: OGM001115: Dialect org.hibernate.ogm.datastore.infinispan.impl.InfinispanEmbeddedStoredProceduresDialect does not support positional parameters when calling stored procedures" );
+		StoredProcedureQuery storedProcedureQuery = em.createStoredProcedureQuery( Car.SIMPLE_VALUE_PROC, Car.class );
+		storedProcedureQuery.registerStoredProcedureParameter( 0, Integer.class, ParameterMode.IN );
+		storedProcedureQuery.setParameter( 0, 1 );
+		storedProcedureQuery.getSingleResult();
+	}
+
+	@Test
+	public void testExceptionWhenInvalidProcedureName() throws Exception {
```
DavideD: Invalid is a very vague statement, is it invalid because it doesn't exists on the server?

DavideD: In that case we should explain the we haven't found the store procedure on the server or something like that

infinispan-embedded/src/test/java/org/hibernate/ogm/datastore/infinispan/test/storedprocedures/NotReallyAProcedure.java
DavideD: You could also throw an exception in the run method to make sure that no test can execute it

infinispan-embedded/src/test/java/org/hibernate/ogm/datastore/infinispan/test/storedprocedures/InfinispanNamedParametersStoredProcedureCallTest.java
ntviet18: We also run the common backendtck test defined in core for Infinispan. In this test we add new case when calling procedure by full qualified class name.

ntviet18:
@DavideD thank you, please see the updates

DavideD:
Thanks, this looks good. I'll merge it in a bit as I want to take the chance to update some existing code

DavideD:
Thanks @ntviet18 I want to take the chances to refactor some existing code before merging it. But I will do it shortly.

DavideD:
Followed by: https://github.com/hibernate/hibernate-ogm/pull/1006

PR #1001, opened 2018-04-06
 OGM-1428 Feature pack distribution for Hibernate OGM WildFly modules

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1428
https://hibernate.atlassian.net/browse/OGM-1439


documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
```diff
@@ -523,6 +504,77 @@ Add a `WEB-INF/jboss-deployment-structure.xml` in your archive with the followin
 More information about the descriptor can be found in the
 https://docs.jboss.org/author/display/WFLY10/Class+Loading+in+WildFly[WildFly documentation].
 
+===== Server provisioning via other build tools
+
+The provisioning tools have been developed mainly for Maven users, but there are options for users of other build tools too.
```
DavideD: I think I will shorten this to:
```
If you are not using Maven in your project, there is also a Gradle plugin [link] available on the official portal;
```

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
```diff
@@ -523,6 +504,77 @@ Add a `WEB-INF/jboss-deployment-structure.xml` in your archive with the followin
...
+
+Gradle users can use the link:https://plugins.gradle.org/plugin/org.wildfly.build.provision[org.wildfly.build.provision] plugin, available on the official Gradle plugin portal.
+
+Both the Maven and the Gradle plugins are just thin wrappers invoking a standalone provisioning library, so you could also integrate this library with any other JVM based tool.
+
+See also https://github.com/wildfly/wildfly-build-tools[WildFly provisioning build tools].
```
DavideD: I might be wrong but aren't this maven plugins?

I would change the title of this paragraph to `Build tools for server provisioning`. This way you can talk first about the Maven plugin and then about the Gradle one

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
```diff
@@ -523,6 +504,77 @@ Add a `WEB-INF/jboss-deployment-structure.xml` in your archive with the followin
 More information about the descriptor can be found in the
 https://docs.jboss.org/author/display/WFLY10/Class+Loading+in+WildFly[WildFly documentation].
 
+===== Server provisioning via other build tools
+
+The provisioning tools have been developed mainly for Maven users, but there are options for users of other build tools too.
+
+Gradle users can use the link:https://plugins.gradle.org/plugin/org.wildfly.build.provision[org.wildfly.build.provision] plugin, available on the official Gradle plugin portal.
+
+Both the Maven and the Gradle plugins are just thin wrappers invoking a standalone provisioning library, so you could also integrate this library with any other JVM based tool.
```
DavideD: I'm not sure I understand what you are suggesting here. I would add it at the end of the paragraph if needed

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
```diff
@@ -523,6 +504,77 @@ Add a `WEB-INF/jboss-deployment-structure.xml` in your archive with the followin
...
+
+See also https://github.com/wildfly/wildfly-build-tools[WildFly provisioning build tools].
+
+[[hibernate-ogm-jboss-modules-feature-packs]]
+==== List of the Hibernate OGM WildFly/JBoss feature packs
+
+===== Core feature pack
+
+This is the base feature pack, extended by all Hibernate OGM dialect feature packs.
+It includes the main module:
```
DavideD: I would change it to: `It contains the core of Hibernate OGM and all the dialects feature packs extend it`

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
```diff
@@ -523,6 +504,77 @@ Add a `WEB-INF/jboss-deployment-structure.xml` in your archive with the followin
...
+
+* _org.hibernate.ogm_, containing Hibernate OGM core module
```
DavideD: Shouldn't this be a groupId + artifactId?
Doesn't this one also include Search and the WildFly feature pack?

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
```diff
@@ -523,6 +504,77 @@ Add a `WEB-INF/jboss-deployment-structure.xml` in your archive with the followin
...
+
+It includes Hibernate ORM {hibernate-orm-version} modules, required by Hibernate OGM {hibernate-ogm-version}.
+Several more modules are included but you should not need to refer to the other ones directly.
```
DavideD: I don't find this sentence very clear, which other modules. why mention them if the user doesn't have to refer to them.

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
```diff
@@ -523,6 +504,77 @@ Add a `WEB-INF/jboss-deployment-structure.xml` in your archive with the followin
...
+
+The feature pack is published on the JBoss Nexus repository and Maven Central as
+link:$$https://repository.jboss.org/nexus/index.html#nexus-search;gav~org.hibernate.ogm~hibernate-ogm-featurepack-core~{hibernate-ogm-version}~~$$[org.hibernate.ogm:hibernate-ogm-featurepack-core:{hibernate-ogm-version}:zip].
+
+===== Infinispan Remote feature pack
+
+This is the Infinispan Remote dialect feature pack.
+It includes the main module:
+
+* _org.hibernate.ogm.infinispan-remote_, containing Hibernate OGM Infinispan Remote module
```
DavideD: It took me a while to understand that this is the module on WIldfly after you install the feature packs.
I think that for a user it would be more important to have the cooridnates of the feature pack first, possibly with the syntax: 
```
		<feature-pack
			groupId="org.hibernate.ogm"
			artifactId="hibernate-ogm-featurepack-infinispan-embedded"
			version="${project.version}" />

```

This way, they can copy and paste it in their provisioning file. Then, you can describe the rest. But maybe I'm missing something, let me know if what I'm saying doesn't make sense.

DavideD: I think we should also add the feature pack configuration line in the section about sa single datastore. So that the user can find it when looking at a specific datastore as well

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
```diff
@@ -577,3 +629,7 @@ Download the Infinispan modules pack for WildFly {wildfly-short-version} from he
 
 Then similarly to what you did with the Hibernate OGM modules zip, unpack this one too in your `modules` directory within the application server.
 
+[NOTE]
+====
+This distribution is already provided by Hibernate OGM Infinispan Remote and Embedded feature packs.
```
DavideD: I think it's hard to understand what you mean with `This distribution`.

Is this paragraph still relevant? It refers to Hibernate OGM modules and they don't exist anymore.

I would change the note to something like:
```
If you are using the Hibernate OGM  Infinispan feature packs, you don't have to worry about this. Infinispan ... is already 
inlcuded in them.
```

featurepack/test/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+    <parent>
+        <groupId>org.hibernate.ogm</groupId>
+        <artifactId>hibernate-ogm-featurepack</artifactId>
+        <version>5.4.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>hibernate-ogm-featurepack-test</artifactId>
+    <packaging>pom</packaging>
+
+    <name>Hibernate OGM WildFly Feature Pack - Demonstration module</name>
+    <description>This is module shows how to provision a WildFly distribution with Hibernate OGM</description>
```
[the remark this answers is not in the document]
fax4ever: Sure is not a "test" right. Maybe we can call this module "example" ?
If we move it into integration testing, we could use it as a basis for integration testing. 
At that point we could call it "wildfly-provisioned" or something.

featurepack/base-infinispan/pom.xml
```diff
@@ -0,0 +1,163 @@
...
+
+    <parent>
+        <groupId>org.hibernate.ogm</groupId>
+        <artifactId>hibernate-ogm-featurepack</artifactId>
+        <version>5.4.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>hibernate-ogm-featurepack-base-infinispan</artifactId>
+    <packaging>pom</packaging>
+
+    <!-- This module can be removed after solving issue: https://issues.jboss.org/browse/ISPN-8689 -->
```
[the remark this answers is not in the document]
fax4ever: I created the issue: https://hibernate.atlassian.net/browse/OGM-1442.

featurepack/test/pom.xml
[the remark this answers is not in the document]
DavideD: BUt doi you need this in the integration testing? Aren't you already creating the wildfly with all the dialects to run the existing tests?

I'm fine with renaming it `example` and keep it in the feature pack module. An alternative could be to add it in the `documentation/examples` with the gettingstarted one. If the purpose is to show the users how one can use feature packs, maybe this last one is the best approach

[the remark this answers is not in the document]
fax4ever: No we don't need it.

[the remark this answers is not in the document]
fax4ever: In documentation/examples could even better. Maybe we can call it "*-wildfly-quickstart" ?

featurepack/test/wildfly-server-provisioning.xml
```diff
@@ -0,0 +1,30 @@
...
+			groupId="org.hibernate.ogm"
+			artifactId="hibernate-ogm-featurepack-mongodb"
+			version="${project.version}" />
+
+		<feature-pack
+			groupId="org.hibernate.ogm"
+			artifactId="hibernate-ogm-featurepack-neo4j"
+			version="${project.version}" />
+
+	</feature-packs>
```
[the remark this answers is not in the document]
fax4ever: Maybe I can extract a property for the version.

featurepack/test/pom.xml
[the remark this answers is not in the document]
fax4ever: In the -wildfly-quickstart documentation can I add sample code and test as in gettingstarted?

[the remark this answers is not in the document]
DavideD: I don't know, I guess you can. Maybe keep all this part in a separate commit so that you can work on it and we can merge it without affecting the rest

featurepack/core/pom.xml
```diff
@@ -0,0 +1,326 @@
...
+    </parent>
+    <artifactId>hibernate-ogm-featurepack-core</artifactId>
+    <packaging>pom</packaging>
+
+    <name>Hibernate OGM WildFly Feature Pack - Hibernate OGM Core</name>
+    <description>Hibernate OGM WildFly Feature Pack - Hibernate OGM Core</description>
+
+    <properties>
+        <hibernate-orm.module.slot>${hibernateOrmFamilyVersion}</hibernate-orm.module.slot>
+        <hibernate.ogm.jipijapa.module.slot>${project.version}</hibernate.ogm.jipijapa.module.slot>
```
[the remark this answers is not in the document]
fax4ever: I taken the properties from the old module project. I would prefer to use dotted version, the dash should be limited to the artifact name.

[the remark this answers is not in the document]
DavideD: Should be alright then

infinispan-remote/pom.xml
```diff
@@ -24,11 +24,13 @@
...
-        <!-- Skipping enforcer plug-in as WildFly / Infinispan Server artifact
-             causes several dependency convergence issues which are beyond our control -->
-        <enforcer.skip>true</enforcer.skip>
+
         <infinispanServerName>infinispan-server</infinispanServerName>
         <infinispan-server.home>${project.build.directory}/${infinispanServerName}</infinispan-server.home>
+
+        <!-- Skipping enforcer dependency convergence check as WildFly / Infinispan Server artifact
+             causes several dependency convergence issues which are beyond our control -->
+        <enforcer.dependencyconvergence.skip>true</enforcer.dependencyconvergence.skip>
```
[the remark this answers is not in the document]
fax4ever: The property now inhibits only dependencyconvergence not the full enforcer. I steal the pattern from Hibernate Search.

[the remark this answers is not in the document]
DavideD: Got it, nice one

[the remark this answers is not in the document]
DavideD: I usually keep this changes in a separate commit though:
1. they could be applied as a separate issue
2. It make it easier for the reviewer to figure out why something has changed (because it is described in the commit message)
3. It's not required to solve the issue

Anyway, don't change it. It is a nice example of what I consider when I need to decide if something should be in a separate commit (we had this conversation in the past).

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
```diff
@@ -372,20 +372,79 @@ for a full list of properties you can explicitly set.
...
+		<feature-pack
+			groupId="org.hibernate.ogm"
+			artifactId="hibernate-ogm-featurepack-neo4j"
+			version="{hibernate-ogm-version}" /> # <2>
+
+	</feature-packs>
+</server-provisioning>
+----
+<1> Add Hibernate Search feature pack, if you need it.
+See https://docs.jboss.org/hibernate/search/5.10/reference/en-US/html_single/#hibernatesearch-jboss-modules-feature-packs[list of available Hibernate Search feature packs].
```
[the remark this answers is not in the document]
fax4ever: I'll move the link out of the boxes

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
fax4ever: WildFly yes, Search is optional

fax4ever: org.hibernate:hibernate-search-jbossmodules-hibernate-orm is not Search feature pack, but the repackaging of Hibernate ORM temporary distributed by Hiberante Search

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
fax4ever: We can remove it, I copied from Search docs, but I agree with you is useless.

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
fax4ever: No the paragraph is not relevant anymore. Feature pack solving the problem of provision of Infinispan module too. I'll remove it.

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
fax4ever: I copy this kind of description from Hibernate Search documentation: https://docs.jboss.org/hibernate/search/5.10/reference/en-US/html_single/#hibernatesearch-jboss-modules-feature-packs.

fax4ever:
@DavideD thank you for the review. I made some changes, based on your directions. Tell me if it is the case to add an extra commit to fill *-wildfly-quickstart module with a sample code.

documentation/examples/wildfly-quickstart/pom.xml
```diff
@@ -6,14 +6,13 @@
 
     <parent>
         <groupId>org.hibernate.ogm</groupId>
-        <artifactId>hibernate-ogm-featurepack</artifactId>
+        <artifactId>hibernate-ogm-documentation-examples</artifactId>
```
DavideD: A suggestion, try to be consistent when you choose a name.
If we have decided to call it wildfly quickstart let's use the same name consistently. This is also helpful for us to check where the name doesn't fit right. 

Right now you have used:
1. `Wildfly-quickstart` for the folder name
2. `documentation-examples` for the artifact id
3. `Demonstration module` for the artifact name

featurepack/core/pom.xml
```diff
@@ -322,5 +322,4 @@
             </plugin>
         </plugins>
     </build>
-
```
DavideD: You can squash this kind of changes in the previous commits, no need to add them in a follow up commit

DavideD:
You r last commit is called `Doc review`, but you are also changing several poms and moving a module around, Can you see how misleading this is?

DavideD:
> Sorry @DavideD but I don't understand the problem here :P

Sorry, my bad, I didn't realize that that was the parent artifact id and not the artifact id of the module.
Forget about it :)

DavideD:
By the way, let me know when you have finished with this PR, please.

Thanks

fax4ever:
@DavideD If there is no anything else PR could be ok.

fax4ever:
@DavideD rebased on master

featurepack/base-infinispan/pom.xml
```diff
@@ -0,0 +1,163 @@
...
+    <artifactId>hibernate-ogm-featurepack-base-infinispan</artifactId>
+    <packaging>pom</packaging>
+
+    <!-- At the time of writing, Infinispan does not provide a feature pack client modules distribution -->
+    <!-- This module can be removed after solving issue: https://issues.jboss.org/browse/ISPN-8689 -->
+    <name>Hibernate OGM WildFly Feature Pack - Temporary Infinispan modules Repackaging</name>
+    <description>Temporary repackaging of Infinispan modules as a Wildfly feature pack</description>
+
+    <properties>
+        <infinispan-client-azure-version>6.1.0</infinispan-client-azure-version>
```
DavideD: This is strange name for the properties, shouldn't be something like infinispanClientAzureVersion?

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
```diff
@@ -372,20 +372,80 @@ for a full list of properties you can explicitly set.
...
-and Hibernate Search {hibernate-search-version}.
+Maven users can use the `wildfly-server-provisioning-maven-plugin`
+to create a custom WildFly server including the Hibernate OGM modules:
+====
+[source, XML]
+[subs="verbatim,attributes"]
+----
+<plugins>
+	<plugin>
+	<groupId>org.wildfly.build</groupId>
```
DavideD: groupId shouldn be in the same level as plugins, it's missing an indentation

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
```diff
@@ -372,20 +372,80 @@ for a full list of properties you can explicitly set.
...
+			<goal>build</goal>
+		</goals>
+		<phase>compile</phase>
+		<configuration>
+			<config-file>server-provisioning.xml</config-file>
+			<server-name>wildfly-with-hibernate-ogm</server-name>
+		</configuration>
+		</execution>
+	</executions>
+</plugin>
```
DavideD: It's missing the closure of the <plugins>

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
```diff
@@ -577,3 +623,7 @@ Download the Infinispan modules pack for WildFly {wildfly-short-version} from he
 
 Then similarly to what you did with the Hibernate OGM modules zip, unpack this one too in your `modules` directory within the application server.
 
+[NOTE]
+====
+If you are using the Hibernate OGM  Infinispan feature packs, you don't have to worry about this. Infinispan client is already inlcuded in them.
```
DavideD: s/inlcuded/included/

DavideD: 2 spaces after Hibernate OGM

integrationtest/server-provisioning.xml
```diff
@@ -11,5 +11,25 @@
...
+		<feature-pack
+				groupId="org.hibernate.ogm"
+				artifactId="hibernate-ogm-featurepack-mongodb"
+				version="${project.version}" />
+
+		<feature-pack
+				groupId="org.hibernate.ogm"
+				artifactId="hibernate-ogm-featurepack-neo4j"
+				version="${project.version}" />
+
```
DavideD: The indentation here is different from the first two feature packs

fax4ever:
Thank you @DavideD  I'll do the last changed and I'll merge it ;)

PR #1005, opened 2018-04-13
OGM-1047 using database pool

schernolyas:
https://hibernate.atlassian.net/browse/OGM-1047

### DIFF
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
@@ -7,16 +7,27 @@
 package org.hibernate.ogm.datastore.neo4j.embedded.impl;
 
 import java.io.File;
+import java.lang.invoke.MethodHandles;
+import java.lang.reflect.Field;
 import java.net.URL;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
 
+import org.hibernate.HibernateException;
 import org.hibernate.ogm.datastore.neo4j.Neo4jProperties;
+import org.hibernate.ogm.datastore.neo4j.logging.impl.Log;
+import org.hibernate.ogm.datastore.neo4j.logging.impl.LoggerFactory;
 import org.hibernate.ogm.datastore.neo4j.spi.GraphDatabaseServiceFactory;
 import org.hibernate.ogm.util.configurationreader.spi.ConfigurationPropertyReader;
+
 import org.neo4j.graphdb.GraphDatabaseService;
 import org.neo4j.graphdb.factory.GraphDatabaseBuilder;
 import org.neo4j.graphdb.factory.GraphDatabaseFactory;
+import org.neo4j.kernel.api.exceptions.TransactionFailureException;
+import org.neo4j.kernel.impl.factory.GraphDatabaseFacade;
+import org.neo4j.kernel.internal.GraphDatabaseAPI;
 
 /**
  * Contains methods to create a {@link GraphDatabaseService} for the embedded Neo4j.
@@ -24,6 +35,10 @@ import org.neo4j.graphdb.factory.GraphDatabaseFactory;
  * @author Davide D'Alto &lt;davide@hibernate.org&gt;
  */
 public class EmbeddedNeo4jGraphDatabaseFactory implements GraphDatabaseServiceFactory {
+	public static final int WAITING_TIME_MS = 100;
+
+	private static final Map<String, FactoryHolder> GRAPH_DATABASE_SERVICE_MAP = new ConcurrentHashMap<>();
+	private static Log LOG = LoggerFactory.make( MethodHandles.lookup() );
 
 	private File dbLocation;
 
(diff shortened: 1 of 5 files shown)


schernolyas:
The PR replace https://github.com/hibernate/hibernate-ogm/pull/987

DavideD:
I remember you introduced this idea to speed up the build. After running it, I didn't notice any improvement and sometimes it seems slightly slower. I was wondering, is it faster the build for you?

schernolyas:
HI @DavideD !

My implementation saving 4 seconds per test. 
Previous implementation starts new instance of Neo4j.  It required 7 seconds in my machine.

schernolyas:
Hi @DavideD !

Could you please review the PR ?

schernolyas:
Hi @DavideD !

Could you please review the PR ?

DavideD:
@fax4ever @gsmet Could you try to build with this PR and let me know if you see some improvement in the building? You should see them just by building the  neo4j-embedded project. Thanks

schernolyas:
Hi @fax4ever and @gsmet !

Main idea of the PR is: using one instance of Neo4j instead start and create new instance.
Instance reusing save 4 seconds for each test (on my machine).

fax4ever:
@DavideD sure, I'm going to try. Hi @schernolyas, I'll try the same test on our environment.

fax4ever:
@schernolyas  On my local environment sometimes some tests fail on this branch.
The tests not fail always. Maybe there is a concurrency issue.

fax4ever:
`[ERROR] Failures: 
[ERROR]   OneToOneTest.testUnidirectionalManyToOne:45->OgmTestCase.checkCleanCache:105 [Entity cache should be empty] expected:<[0]> but was:<[3]>
[INFO] 
[ERROR] Tests run: 454, Failures: 1, Errors: 0, Skipped: 36
`

Sanne:
Often that means that some test didn't cleanup its previous state.

So it might be a problem of existing tests, likely one of those specific to Neo4J.

My guess is we didn't see such problems before as we'd be starting a new database.

gsmet:
We had a very similar failure on Travis for a while and we never got to the bottom of it. Might be interesting to try to understand what's going on if you can reproduce it.

fax4ever:
Sometimes the list of failure tests changes, another output could be: `[ERROR] Failures: 
[ERROR]   ListTest.testOrderedList:62->OgmTestCase.checkCleanCache:105 [Entity cache should be empty] expected:<[0]> but was:<[3]>
[ERROR]   ListTest.testOrderedListAndCompositeId:172->OgmTestCase.checkCleanCache:105 [Entity cache should be empty] expected:<[0]> but was:<[3]>`

fax4ever:
I'm going to spend some time to investigate the problem of tests that fails. I'll keep you posted.

DavideD:
@fax4ever an additional test that you can try is to remove the tests and then check with Travis if the build pass. Note that you will need to launch the build several times becase the first time the error doesn't appear

DavideD:
I've changed the Label to on `ice` but the idea is to fix the problem with our test suite first

fax4ever:
Changing the clauses to choose if rollback the transaction. See [OGM-1471](https://hibernate.atlassian.net/browse/OGM-1471). This PR tests work.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
```diff
@@ -7,23 +7,38 @@
...
 
 /**
  * Contains methods to create a {@link GraphDatabaseService} for the embedded Neo4j.
  *
  * @author Davide D'Alto &lt;davide@hibernate.org&gt;
  */
 public class EmbeddedNeo4jGraphDatabaseFactory implements GraphDatabaseServiceFactory {
+	public static final int WAITING_TIME_MS = 100;
+
+	private static final Map<String, FactoryHolder> GRAPH_DATABASE_SERVICE_MAP = new ConcurrentHashMap<>();
```
fax4ever: It is really necessary to synchronize use:
1. ConcurrentHashMap
2. AtomicBoolean
3. synchronized 
together ?

schernolyas: HI @fax4ever !
AtomicBoolean needs for return result from lambda expression
Method "computeIfAbsent" of ConcurentMap needs for return (create or just return) FactoryHolder for required path

Synchronization by ConcurentMap ...It can be excessively ...need think

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
```diff
@@ -86,4 +139,74 @@ private void setConfigurationFromLocation(GraphDatabaseBuilder builder, URL cfgL
...
+		}
+	}
+
+	private static Object getPrivateField(Field privateField, Object targetObject) throws IllegalAccessException {
+		privateField.setAccessible( true );
+		return privateField.get( targetObject );
+	}
+
+
+	private static class FactoryHolder {
```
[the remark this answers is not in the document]
schernolyas: it is not counter only. It is holder of information about instance. And, of course, it contains counter of connections to database instance

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/EmbeddedGraphDatabaseFactoryTest.java
```diff
@@ -68,6 +66,51 @@ public void testLoadMalformedPropertiesLocation() throws Exception {
 		factory.create().shutdown();
 	}
 
+	@Test
+	public void testInstantiationTwoFactoriesForOnePath() {
+		//create first factory
```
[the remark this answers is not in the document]
schernolyas: May be. What is your offer?
The test checks that two instances for one database path is same.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
fax4ever: Let me understand the scope of the issue. This issue is made to avoid re-creating a service if I need a second session factory? Let's talk more about the real use case. Please

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
```diff
@@ -86,4 +142,83 @@ private void setConfigurationFromLocation(GraphDatabaseBuilder builder, URL cfgL
...
+		private int counter;
+		private GraphDatabaseFacade graphDatabaseFacade;
+		private boolean newHolder;
```
fax4ever: are you sure you need this field?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
```diff
@@ -86,4 +142,83 @@ private void setConfigurationFromLocation(GraphDatabaseBuilder builder, URL cfgL
 			builder.loadPropertiesFromURL( cfgLocation );
 		}
 	}
+
+	static void shutdownGraphDatabaseService(GraphDatabaseService neo4jDb) {
+		GraphDatabaseAPI sq = (GraphDatabaseAPI) neo4jDb;
+		String key = sq.getStoreDir().getAbsolutePath();
+		if ( GRAPH_DATABASE_SERVICE_MAP.containsKey( key ) ) {
+			synchronized (GRAPH_DATABASE_SERVICE_MAP) {
```
fax4ever: Are you should you need to a nested synchronized?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
schernolyas: I need mark new instance.
If the instance is not new I need increment counter of clients.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
schernolyas: I try organize queue. You right. Thanks!

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jGraphDatabaseFactory.java
fax4ever: Maybe you can use the counter, it's an idea. WDYT?

schernolyas:
Hi !

What is verdict for the PR ? 
Last activity for the PR was at May. Now is August. 
Any ideas ?
Thanks!

DavideD:
I'm  not going to include this solution because I am afraid it might cause weird problems hard too debug,

PR #1008, opened 2018-04-14
OGM-1407 Neo4j support for queries with jpa positional parameters

vk23:


DavideD:
ok to test

DavideD:
Thanks @vk23, I'll have a look soon

schernolyas:
Hi @vk23 !

Please, fix comments of checkstyle.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BaseNeo4jDialect.java
```diff
@@ -170,6 +171,20 @@ private void applyMaxRows(QueryParameters queryParameters, StringBuilder nativeQ
 		}
 	}
 
+	private void applyPositionalParameters(QueryParameters queryParameters, StringBuilder nativeQuery) {
+		for ( String key : queryParameters.getNamedParameters().keySet() ) {
+			removePositionalQuestionMark( nativeQuery, key );
+		}
+	}
```
gsmet: Quick question: what's exactly the purpose of this method? Because AFAICS, it does not "apply" anything, it just removes question marks with a rather fragile algorithm.

Thanks for the explanation.

vk23: Probably this is not an appropriate name for the method because it just removes symbols so it must renamed. But what is the right way to remove question marks from the resulting query?

gsmet: OK, so, yes, the name of the method is definitely not the best we could choose. Let's change that.

Frankly, it's a bit of a mess to have to do that, considering we have parsed the query with a real parser before and we have the position of the parameters (or at least should have with the parser).

Having played with it a bit, I'm unfortunately not sure we can do better, considering how messy the parameter code is in ORM 5.2. We probably will have to wait for 6 to have something better.

It's really annoying that the Neo4j client does not use the same syntax as JDBC and more annoying even that they don't have a proper marker for variable as replacing `0` in the query might lead to weird behavior (typically if you have the 0 value in your query, I wonder what it would make).

I'm wondering if maybe we shouldn't try to rewrite the query by using the parser information but our parser is not very advanced so it won't be perfect either and I don't remember us having a way to rewrite the query.

To explain what I have in mind, if possible, I would have rewritten the `?0` into `__parameter_0` (something compatible with Neo4j) but to summarize, we would need:
* a more solid parser
* a way to rewrite the query

Two things we don't have right now. It will probably be something to consider when we will move to 6 and try to also implement named parameters.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BaseNeo4jDialect.java
```diff
@@ -170,6 +171,20 @@ private void applyMaxRows(QueryParameters queryParameters, StringBuilder nativeQ
...
+
+	private void removePositionalQuestionMark(StringBuilder nativeQuery, String key) {
+		int i = nativeQuery.indexOf( "?" + key );
```
gsmet: To be more precise, when I was talking about this being fragile, I'm thinking about having a `?13` string somewhere in the query and trying to replace `?1`.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BaseNeo4jDialect.java
DavideD: When I created this issue I was thinking more about JPQL queries and being able to support positional parameters for that kind of queries. I don't know if in Cypher you can specify a position instead of a name for a parameter, but I'm not sure if I want to start the path where we need to parse a native Cypher query.

vk23: @DavideD 
But org.hibernate.ogm.dialect.query.spi.RecognizerBasedParameterMetadataBuilder#buildParameterMetadata is never called for JPQL in neo4j, only for native queries.

DavideD: So, it seems I made a mistake. When you asked me about the issue I made confusion about two separate problems.  When I answered you question I was thinking about JPA but I thin kwhen I created it I was thinking about positional parameters in Cypher queries.

Anyway the solution I had in mind was much simpler that what you are proposing.
Cypher accepts parameters in the form of `{0}`, `{1}` and so on, for native queries we could consider this as positional parameters. 

I think right now we are treating them as named parameter so I'm not sure if we should make this work for both named and positional.

DavideD: This make sense because what we are talking about is native queries, so we don't have to follow the JPA syntax, for JPQL queries instead we have to. Sorry for the confusion

DavideD:
@vk23 Small advice, 4 commits with the same description is bad practice.

For future reference, make sure to split the fix with self contained commits. Basically, a single commit should focus on one aspect of the PR and the message should describe shortly whats happening (I don't want to have to check the code to understand what's different between the commits).

Here some of things I consider when sending a PR: 

- New files in separate commits: This make it simpler to rebase the commit and the reviewer can focus on file that have changed compared to the one that have been added

- New utility functions in separate commits: Utility functions can potentially be used in other situations, introducing them as separate commits, make it easier to apply the commit to other branches if needed

- Tests in separate commits: Make it possible for me to check the test and see they fail without the fix

- Trying to keep changes to different modules in different commit, in particular the one in the documentation

- Refactoring and style changes of existing code in separate commits: these are usually not critical to pass the tests and might not get included

Basically a commit should contain the minimum amount of code to solve a problem. Keep in mind that these are guidelines and not strict rules. It depends from the issue one is trying to solve and the aim is to make life easier for the people working on the project.

Thanks

vk23:
@DavideD 
Ok, I got it. So what should I do now? Close this PR and create new one according these rules?

DavideD:
> Ok, I got it. So what should I do now? Close this PR and create new one according these rules?

You can choose what's easier for you, I usually rebase the branch and push force using:

    git rebase -i master

PR #1011, opened 2018-04-19
Ogm 1407 Neo4j support for queries with jpa positional parameters

vk23:


neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/query/impl/QueryParser.java
```diff
@@ -41,21 +41,24 @@ public Rule QueryParts() {
...
-		return Sequence(
-			ParameterBeginDelimiter(),
-			ZeroOrMore( WhiteSpace() ),
-			OneOrMore( Alphanumeric() ),
-			name.set( match() ),
-			ZeroOrMore( WhiteSpace() ),
-			ParameterEndDelimiter(),
-			adapter.addNamedParameter( name.get(), currentIndex() )
+		return OneOrMore(
+				TestNot( OrdinalParameter() ),
```
DavideD: I'm not following this approach.
This rule is called `NamedParameter` but inside you are  checking if it's a positional parameter.
Wouldn't be easier to change queryParst to:
```
 @SkipNode
 public Rule QueryPart() {
    return FirstOf( Quoted(), Escaped(), FirstOf(OrdinalParameterAction(), NamedParameter()), Other() );
 }
```

Note that I would rename `OrdinalParameterAction` to `Ordinalparameter`

vk23: 1.  I check if it's NOT ordinal parameter because anytime we need to test against named parameter we need to be sure that it's not ordinal. I believe it's not enough to just change priority only in QueryPart() because NamedParameter() is tested not only here.
2. There is a difference between sequence and action rules. When you want to capture value you use action (which parses and adds parameter)  but when you want just to check if this is namedParameter or any other you just test against appropriate sequence. Probably there is a bug: any time NamedParameter() is called it captures value and adds named parameter to result (eg in Rule Other()).

DavideD: > believe it's not enough to just change priority only in QueryPart() because NamedParameter() is tested not only here.

You can still rename NamedParameter() to Parameter() and execute the FirstOf(...). This way you know that it is tested in all the places NamedParameter is tested.

> Probably there is a bug: any time NamedParameter() is called it captures value and adds named parameter to result (eg in Rule Other()).

That might be a good point, still I don't think the rule NamedParameter() should contain a TestNot( OrdinalParameter ),  I would expect to have a NamedParameterAction() as well to keep the code consistent.
I would also prefer not to have duplicated logic between rules. What about this:
```
	@SuppressSubnodes
	public Rule OrdinalParameterAction() {
		StringVar identifier = new StringVar( "" );
		return Sequence( OrdinalParameter( identifier ), adapter.addPositionalParameter( identifier.get(), currentIndex() ) );
	}

	@SuppressSubnodes
	public Rule OrdinalParameterTest() {
		return OrdinalParameter( new StringVar( "" ) );
	}

	@SuppressSubnodes
	public Rule OrdinalParameter(StringVar identifier) {

		return Sequence(
				ParameterBeginDelimiter(),
				ZeroOrMore( WhiteSpace() ),
				OneOrMore( Digit() ),
				identifier.set( match() ),
				ZeroOrMore( WhiteSpace() ),
				ParameterEndDelimiter() );
	}

```

Note that you would need the same for NamedParamater.

I haven't tried it, but an additional imporovement could be to have a Parameter rule that reads the parameter name and then let the adapter check if it's a number or named. Something like:
```
addToTheAdapter( String paramName) {
    if isNumeric( paramName ) {
        adapter.addPositionalParameter( identifier.get(), currentIndex() );
    }
    else {
        adapter.addNamedParameter( name.get(), currentIndex() )
    }
}
```

DavideD:
Merged with some additional commits: https://github.com/hibernate/hibernate-ogm/pull/1011

DavideD:
and rebased

PR #1012, opened 2018-04-24
OGM-1426 Infinispan remote: enable support to JPQL

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1426
https://hibernate.atlassian.net/browse/OGM-1457

Currently we have the same limitation of Neo4j and MongoDb dialects:
* They don't support queries on polymorphic entities using TABLE_PER_CLASS inheritance strategy
* Selecting from associations is not yet implemented
* Query on Byte filed is not currently supported

### DIFF
--- a/core/src/test/java/org/hibernate/ogm/backendtck/inheritance/tableperclass/family/TablePerClassInheritancePersistTest.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/inheritance/tableperclass/family/TablePerClassInheritancePersistTest.java
@@ -7,6 +7,7 @@
 package org.hibernate.ogm.backendtck.inheritance.tableperclass.family;
 
 import static org.fest.assertions.Assertions.assertThat;
+import static org.hibernate.ogm.datastore.impl.DatastoreProviderType.INFINISPAN_REMOTE;
 import static org.hibernate.ogm.datastore.impl.DatastoreProviderType.MONGODB;
 import static org.hibernate.ogm.datastore.impl.DatastoreProviderType.NEO4J_BOLT;
 import static org.hibernate.ogm.datastore.impl.DatastoreProviderType.NEO4J_EMBEDDED;
@@ -68,7 +69,7 @@ public class TablePerClassInheritancePersistTest extends OgmJpaTestCase {
 	@Test
 	@TestForIssue(jiraKey = "OGM-1294")
 	@SkipByDatastoreProvider(
-			value = { MONGODB, NEO4J_BOLT, NEO4J_HTTP, NEO4J_EMBEDDED },
+			value = { MONGODB, NEO4J_BOLT, NEO4J_HTTP, NEO4J_EMBEDDED, INFINISPAN_REMOTE },
 			comment = "They don't support queries on polymorphic entities using TABLE_PER_CLASS inheritance strategy; requires multiple queries")
 	public void testJpqlWithSingleResult() {
 		initDB();
@@ -82,7 +83,7 @@ public class TablePerClassInheritancePersistTest extends OgmJpaTestCase {
 	@Test
 	@TestForIssue(jiraKey = "OGM-1294")
 	@SkipByDatastoreProvider(
-			value = { MONGODB, NEO4J_BOLT, NEO4J_HTTP, NEO4J_EMBEDDED },
+			value = { MONGODB, NEO4J_BOLT, NEO4J_HTTP, NEO4J_EMBEDDED, INFINISPAN_REMOTE },
 			comment = "They don't support queries on polymorphic entities using TABLE_PER_CLASS inheritance strategy; requires multiple queries")
 	public void testJpql() {
 		initDB();
@@ -96,7 +97,7 @@ public class TablePerClassInheritancePersistTest extends OgmJpaTestCase {
 	@Test
 	@TestForIssue(jiraKey = "OGM-1294")
 	@SkipByDatastoreProvider(
-			value = { MONGODB, NEO4J_BOLT, NEO4J_HTTP, NEO4J_EMBEDDED },
+			value = { MONGODB, NEO4J_BOLT, NEO4J_HTTP, NEO4J_EMBEDDED, INFINISPAN_REMOTE },
 			comment = "They don't support queries on polymorphic entities using TABLE_PER_CLASS inheritance strategy; requires multiple queries")
 	public void testJpqlReturnPropertiesForMan() {
 		initDB();
@@ -112,7 +113,7 @@ public class TablePerClassInheritancePersistTest extends OgmJpaTestCase {
 	@Test
 	@TestForIssue(jiraKey = "OGM-1294")
 	@SkipByDatastoreProvider(
-			value = { MONGODB, NEO4J_BOLT, NEO4J_HTTP, NEO4J_EMBEDDED },
+			value = { MONGODB, NEO4J_BOLT, NEO4J_HTTP, NEO4J_EMBEDDED, INFINISPAN_REMOTE },
 			comment = "They don't support queries on polymorphic entities using TABLE_PER_CLASS inheritance strategy; requires multiple queries")
 	public void testJpqlReturnPropertiesForChildren() {
 		initDB();
@@ -132,7 +133,7 @@ public class TablePerClassInheritancePersistTest extends OgmJpaTestCase {
 	@Test
 	@TestForIssue(jiraKey = "OGM-1294")
 	@SkipByDatastoreProvider(
-			value = { MONGODB, NEO4J_BOLT, NEO4J_HTTP, NEO4J_EMBEDDED },
+			value = { MONGODB, NEO4J_BOLT, NEO4J_HTTP, NEO4J_EMBEDDED, INFINISPAN_REMOTE },
 			comment = "They don't support queries on polymorphic entities using TABLE_PER_CLASS inheritance strategy; requires multiple queries")
 	public void testJpqlReturnPropertiesForSubClasses() {
 		initDB();
@@ -158,7 +159,7 @@ public class TablePerClassInheritancePersistTest extends OgmJpaTestCase {
 	@Test
 	@TestForIssue(jiraKey = "OGM-1294")
 	@SkipByDatastoreProvider(
-			value = { MONGODB, NEO4J_BOLT, NEO4J_HTTP, NEO4J_EMBEDDED },
+			value = { MONGODB, NEO4J_BOLT, NEO4J_HTTP, NEO4J_EMBEDDED, INFINISPAN_REMOTE },
 			comment = "They don't support queries on polymorphic entities using TABLE_PER_CLASS inheritance strategy; requires multiple queries")
 	public void testPolymorphicAssociation() {
 		initDB();
--- a/documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
+++ b/documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
@@ -2709,12 +2709,13 @@ We always recommend using transactional caches, at least unless you are confiden
 ==== Remote Query Capability
 
 Hibernate OGM backend is able to run the queries it needs to materialize relations.
-From version __5.4.0.Alpha2__ the dialect is capable to translate JPQL or Criteria queries to the Infinispan remote queries.
+From version __5.4.0.Alpha2__ the dialect is capable to translate JPQL to the **corresponding** Infinispan remote queries
+or to execute native queries.
 But there are still some limitations:
 
 * We don't support queries on polymorphic entities using TABLE_PER_CLASS inheritance strategy
-* Selecting from associations is not yet implemented
-* Query on Byte filed is not currently supported
+* Selecting from associations is not yet implemented (example: select h.author.name from Hypothesis h where author is a one-to-one association)
+* Query on Byte filed is currently unsupported
 
 ==== Known Limitations & Future improvements
 
--- a/documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
+++ b/documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
@@ -2577,24 +2577,10 @@ Hibernate OGM by convention will write to several named ``Cache``s, mapping each
 to a "cache name". In the above example, when having an `Hypothesis` entity this will
 write to a Cache named `Hypothesis`.
 
-The benefit is that you can tune each cache (each "table") independently; for example you could
+The benefit is that you can tune, or query, each cache (each "table") independently; for example you could
 configure the caches for the most important data to have a synchronous CacheStore which replicates
 data to a relational database, and have less important entries use an asynchronous CacheStore,
 or none at all, to favour performance over redundancy.
-
-The drawback of this design choice is that each named cache must be pre-defined in the Infinispan
-Server configuration: at this point, the Hot Rod protocol is not allowed to start missing caches
-so Hibernate OGM can not define the missing tables automatically.
-It generates the encoding protobuf, but you have to list the Cache names in the server configuration.
-
-[WARNING]
-====
-For each "table name" your model would generate on a relational database, you have to define
-a matching Cache on the Infinispan Server.
-
-If any Cache is missing, Hibernate OGM will fail to start and list which table names were
-expected, but not found. We plan to automate the creation of missing caches in the future.
-====
 
 ==== Caches creation and Configuration
 
@@ -2720,6 +2706,16 @@ __OGM default configuration__ is transactional.
 We always recommend using transactional caches, at least unless you are confident that data consistency is not important for your use case.
 ====
 
+==== Remote Query Capability
+
+Hibernate OGM backend is able to run the queries it needs to materialize relations.
+From version __5.4.0.Alpha2__ the dialect is capable to translate JPQL or Criteria queries to the Infinispan remote queries.
+But there are still some limitations:
+
+* We don't support queries on polymorphic entities using TABLE_PER_CLASS inheritance strategy
+* Selecting from associations is not yet implemented
+* Query on Byte filed is not currently supported
+
 ==== Known Limitations & Future improvements
 
 The Infinispan Remote dataprovider has some known limitations, some of which are
@@ -2728,11 +2724,6 @@ unsolvable without further development of Infinispan itself.
 Transaction Support::
 We're eagerly waiting for Infinispan to support transactions over Hot Rod, as it
 already provides this feature in Embedded Mode.
-
-Queries::
-At this point the Hibernate OGM backend is able to run the queries it needs to materialize
-relations, but does not yet translate JPQL queries nor Criteria queries to the
-Infinispan remote queries.
 
 Indexing::
 Infinispan supports Hibernate Search annotations directly embedded within its protobuf
--- a/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
+++ b/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
@@ -28,6 +28,7 @@ import org.hibernate.ogm.datastore.infinispanremote.logging.impl.Log;
 import org.hibernate.ogm.datastore.infinispanremote.logging.impl.LoggerFactory;
 import org.hibernate.ogm.datastore.infinispanremote.query.impl.InfinispanRemoteQueryDescriptor;
 import org.hibernate.ogm.datastore.infinispanremote.query.impl.InfinispanRemoteQueryHandler;
+import org.hibernate.ogm.datastore.infinispanremote.query.parsing.impl.InfinispanRemoteNativeQueryParser;
 import org.hibernate.ogm.datastore.map.impl.MapAssociationSnapshot;
 import org.hibernate.ogm.datastore.map.impl.MapHelpers;
 import org.hibernate.ogm.datastore.map.impl.MapTupleSnapshot;
@@ -179,7 +180,7 @@ public class InfinispanRemoteDialect<EK, AK, ISK> extends AbstractGroupingByEnti
 
 	@Override
 	public InfinispanRemoteQueryDescriptor parseNativeQuery(String nativeQuery) {
-		throw new UnsupportedOperationException( "Native Query not supported by Infinispan Remote Dialect" );
+		return new InfinispanRemoteNativeQueryParser( nativeQuery ).parse();
 	}
 
 	/**
--- a/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
+++ b/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.ogm.datastore.infinispanremote;
 
+import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -25,7 +26,8 @@ import org.hibernate.ogm.datastore.infinispanremote.impl.protostream.Protostream
 import org.hibernate.ogm.datastore.infinispanremote.impl.protostream.ProtostreamPayload;
 import org.hibernate.ogm.datastore.infinispanremote.logging.impl.Log;
 import org.hibernate.ogm.datastore.infinispanremote.logging.impl.LoggerFactory;
-import java.lang.invoke.MethodHandles;
+import org.hibernate.ogm.datastore.infinispanremote.query.impl.InfinispanRemoteQueryDescriptor;
+import org.hibernate.ogm.datastore.infinispanremote.query.impl.InfinispanRemoteQueryHandler;
 import org.hibernate.ogm.datastore.map.impl.MapAssociationSnapshot;
 import org.hibernate.ogm.datastore.map.impl.MapHelpers;
 import org.hibernate.ogm.datastore.map.impl.MapTupleSnapshot;
@@ -36,7 +38,12 @@ import org.hibernate.ogm.dialect.batch.spi.Operation;
 import org.hibernate.ogm.dialect.batch.spi.RemoveAssociationOperation;
 import org.hibernate.ogm.dialect.impl.AbstractGroupingByEntityDialect;
 import org.hibernate.ogm.dialect.multiget.spi.MultigetGridDialect;
+import org.hibernate.ogm.dialect.query.spi.BackendQuery;
 import org.hibernate.ogm.dialect.query.spi.ClosableIterator;
+import org.hibernate.ogm.dialect.query.spi.NoOpParameterMetadataBuilder;
+import org.hibernate.ogm.dialect.query.spi.ParameterMetadataBuilder;
+import org.hibernate.ogm.dialect.query.spi.QueryParameters;
+import org.hibernate.ogm.dialect.query.spi.QueryableGridDialect;
 import org.hibernate.ogm.dialect.spi.AssociationContext;
 import org.hibernate.ogm.dialect.spi.AssociationTypeContext;
 import org.hibernate.ogm.dialect.spi.DuplicateInsertPreventionStrategy;
@@ -60,6 +67,7 @@ import org.hibernate.ogm.model.spi.AssociationOperation;
 import org.hibernate.ogm.model.spi.AssociationOperationType;
 import org.hibernate.ogm.model.spi.Tuple;
 import org.hibernate.ogm.model.spi.Tuple.SnapshotType;
+
 import org.infinispan.client.hotrod.MetadataValue;
 import org.infinispan.client.hotrod.RemoteCache;
 import org.infinispan.client.hotrod.Search;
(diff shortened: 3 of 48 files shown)


infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemoteQueryRendererDelegate.java
```diff
@@ -0,0 +1,175 @@
...
+		}
+
+		queryBuilder.append( discriminatorColumnName );
+
+		if ( subclassEntityNames.size() == 1 ) {
+			queryBuilder.append( " = " );
+			appendValue( queryBuilder, discriminatorValue );
+		}
+		else {
+			ArrayList<Object> discriminatorValues = new ArrayList<>();
```
DavideD: Not critical but... ArrayList? Let's keep List

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemoteQueryRendererDelegate.java
```diff
@@ -0,0 +1,175 @@
...
+			Set<String> subclassEntityNames = entityPersister.getEntityMetamodel().getSubclassEntityNames();
+			throw LOG.queriesOnPolymorphicEntitiesAreNotSupportedWithTablePerClass( "Infinispan Server", subclassEntityNames );
+		}
+	}
+
+	private void addConditionOnDiscriminatorValue(OgmEntityPersister entityPersister, StringBuilder queryBuilder, String discriminatorColumnName) {
+		Object discriminatorValue = entityPersister.getDiscriminatorValue();
+		Set<String> subclassEntityNames = entityPersister.getEntityMetamodel().getSubclassEntityNames();
+
+		boolean hasWhere = queryBuilder.toString().contains( "where" );
```
DavideD: You cannot use this approach (I think), what if one of the column in the projection is called `where_are_you`. 
Basically, there could be a `where` hidden somewhere. 
Parsing a String this way is error prone and there is always a case where it won't work.
A better approach is to keep track of the projection and conditions separately and then merge them at the end.
See what I've done in Neo4jQueryRendererDelegate.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemoteQueryResolverDelegate.java
```diff
@@ -0,0 +1,139 @@
...
+import org.hibernate.hql.ast.origin.hql.resolve.path.PathedPropertyReferenceSource;
+import org.hibernate.hql.ast.origin.hql.resolve.path.PropertyPath;
+import org.hibernate.hql.ast.spi.QueryResolverDelegate;
+import org.hibernate.ogm.datastore.infinispanremote.logging.impl.Log;
+import org.hibernate.ogm.datastore.infinispanremote.logging.impl.LoggerFactory;
+
+import org.antlr.runtime.tree.Tree;
+
+/**
+ * Query resolver delegate targeting Infinispan server queries. Very basic implementation atm.
```
DavideD: Let's remove the second sentence

DavideD: This class looks very similar to the other ones. Would it make sense to have a `BaseQueryResolverDelegate` and keep there all the common implementations?

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemoteQueryStringBuilderHelper.java
```diff
@@ -0,0 +1,49 @@
...
+package org.hibernate.ogm.datastore.infinispanremote.query.parsing.impl;
+
+import java.util.List;
+
+/**
+ * Helper class to append single or multiple values to a {@link StringBuilder}
+ *
+ * @author Fabio Massimo Ercoli
+ */
+public class InfinispanRemoteQueryStringBuilderHelper {
```
DavideD: This class is not used anywhere

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/predicate/impl/InfinispanRemoteComparisonPredicate.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+package org.hibernate.ogm.datastore.infinispanremote.query.parsing.impl.predicate.impl;
+
+import static org.hibernate.ogm.datastore.infinispanremote.query.parsing.impl.InfinispanRemoteQueryStringBuilderHelper.appendValue;
+
+import org.hibernate.hql.ast.spi.predicate.ComparisonPredicate;
+import org.hibernate.hql.ast.spi.predicate.NegatablePredicate;
+
+/**
+ * Infinispan remote-based implementation of {@link ComparisonPredicate}.
```
DavideD: I would remove the `-based`

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/predicate/impl/InfinispanRemoteDisjunctionPredicate.java
```diff
@@ -0,0 +1,54 @@
...
+		int counter = 1;
+		builder.append( "(" );
+		for ( Predicate<StringBuilder> child : children ) {
+			builder.append( ( (NegatablePredicate<StringBuilder>) child ).getNegatedQuery() );
+			builder.append( ")" );
+			if ( counter++ < children.size() ) {
+				builder.append( " and (" );
+			}
+		}
+		return builder;
```
DavideD: There is a lot of duplicated code here

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemoteBasedQueryParserService.java
```diff
@@ -0,0 +1,67 @@
...
+		throw new UnsupportedOperationException( "The Infinispan Remote query parser supports parameterized queries" );
+	}
+
+	@Override
+	public QueryParsingResult parseQuery(SessionFactoryImplementor sessionFactory, String queryString) {
+		QueryParser queryParser = new QueryParser();
+		InfinispanRemoteProcessingChain processingChain = createProcessingChain( sessionFactory );
+		InfinispanRemoteQueryParsingResult result = queryParser.parseQuery( queryString, processingChain );
+
+		log.createdQuery( queryString, result );
```
DavideD: Is it going to be any real difference between queryString and result? Wouldn't be easier to return the queryString?

DavideD:
I just realized that you didn't add any test related to the query "conversion". It would be nice to have something similar to `CypherQueryParsingTest`.

core/src/test/java/org/hibernate/ogm/backendtck/inheritance/tableperclass/family/TablePerClassInheritancePersistTest.java
```diff
@@ -68,7 +69,7 @@ public void testPersistEntititesWithoutErrors() {
 	@Test
 	@TestForIssue(jiraKey = "OGM-1294")
 	@SkipByDatastoreProvider(
-			value = { MONGODB, NEO4J_BOLT, NEO4J_HTTP, NEO4J_EMBEDDED },
+			value = { MONGODB, NEO4J_BOLT, NEO4J_HTTP, NEO4J_EMBEDDED, INFINISPAN_REMOTE },
```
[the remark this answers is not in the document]
fax4ever: I chose to add test skips at the beginning in order to avoid having commits that would not pass test. But as you prefer.

core/src/main/java/org/hibernate/ogm/query/parsing/impl/SingleEntityQueryRendererDelegateKeepParameterPlaceholders.java
```diff
@@ -0,0 +1,177 @@
...
+import org.hibernate.hql.ast.spi.SingleEntityQueryRendererDelegate;
+import org.hibernate.hql.ast.spi.predicate.ComparisonPredicate;
+
+/**
+ * Temporary class overrides {@link SingleEntityQueryRendererDelegate} only to update query parameter binding.
+ * Parameters are keep as they are, no values are applied on them.
+ * Field {@link SingleEntityQueryRendererDelegate#namedParameters} are useless here.
+ *
+ * This class overrides 12 methods, to change only the private method: {@link SingleEntityQueryRendererDelegate#parameterValue(String)}.
+ * This is too bad. But we can be removed it, after that issue [https://hibernate.atlassian.net/browse/HQLPARSER-79] has been solved.
```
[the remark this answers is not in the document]
fax4ever: Much better as you suggested

core/src/test/java/org/hibernate/ogm/backendtck/inheritance/tableperclass/family/TablePerClassInheritancePersistTest.java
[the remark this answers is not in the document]
DavideD: > I chose to add test skips at the beginning in order to avoid having commits that would not pass test. But as you prefer.

There is a misunderstanding, when I tried t orun the build for this commit there was a failure. Maybe it wasn't caused by this particular change but something in this commit is not right.

It's not about what I prefer, it just that it seems there is a problem, if you think that I'm wrong and the build pass with this commit I will check again.

core/src/main/java/org/hibernate/ogm/query/parsing/impl/SingleEntityQueryRendererDelegateKeepParameterPlaceholders.java
```diff
@@ -0,0 +1,177 @@
...
+ *
+ * @param <Q> Builder class
+ * @param <R> Output class
+ */
+public abstract class SingleEntityQueryRendererDelegateKeepParameterPlaceholders<Q, R> extends SingleEntityQueryRendererDelegate<Q, R> {
```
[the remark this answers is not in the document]
fax4ever: I'm completely agree about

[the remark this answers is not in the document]
fax4ever: This was the very idea I had when I created the new base class, using this for all dialects that support named parameters

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
```diff
@@ -156,6 +166,26 @@ else if ( operation instanceof RemoveAssociationOperation ) {
...
+	}
+
+	@Override
+	public ParameterMetadataBuilder getParameterMetadataBuilder() {
+		return null;
+	}
+
+	@Override
+	public InfinispanRemoteQueryDescriptor parseNativeQuery(String nativeQuery) {
+		throw new UnsupportedOperationException( "Native Query not supported by Infinispan Remote Dialect" );
```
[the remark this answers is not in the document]
fax4ever: In my first implementation I had used a String as query descriptor, like in Neo4j. But then I had to introduce the class InfinispanRemoteQueryDescriptor, because the dialect needs to know also the cache on which perform the query and the exact projection rows.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/impl/ProtostreamPayloadClosableIterator.java
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	@Override
+	public boolean hasNext() {
+		return index < queryResult.size();
+	}
+
+	@Override
+	public Tuple next() {
+		return queryResult.get( index++ ).toTuple( Tuple.SnapshotType.UPDATE );
```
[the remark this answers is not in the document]
fax4ever: It does not return null if the method was invoked only after hasNext. Checking the index could be defensive programming, I think. In my opinion defensive programming should be applied only to API, or SPI, not for internal collaboration. But as you prefer.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemoteBasedQueryParserService.java
```diff
@@ -0,0 +1,67 @@
...
+import org.hibernate.ogm.service.impl.SessionFactoryEntityNamesResolver;
+import org.hibernate.ogm.util.impl.Log;
+import org.hibernate.ogm.util.impl.LoggerFactory;
+
+/**
+ * A {@link QueryParserService} implementation which creates Infinispan server queries in form of {@link String}s.
+ *
+ * @author Fabio Massimo Ercoli
+ */
+public class InfinispanRemoteBasedQueryParserService extends BaseQueryParserService {
```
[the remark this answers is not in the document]
fax4ever: I'm agree with you, but I just applied what was done with MongoDB and Neo4j, where the similar classes are named MongoDBBasedQueryParserService and InfinispanRemoteBasedQueryParserService.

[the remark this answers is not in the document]
fax4ever: If you want if I change it, I think we have change MongoDBBasedQueryParserService and InfinispanRemoteBasedQueryParserService too.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemoteQueryRendererDelegate.java
```diff
@@ -0,0 +1,175 @@
...
+import org.hibernate.ogm.datastore.infinispanremote.logging.impl.LoggerFactory;
+import org.hibernate.ogm.persister.impl.OgmEntityPersister;
+import org.hibernate.ogm.query.parsing.impl.SingleEntityQueryRendererDelegateKeepParameterPlaceholders;
+
+/**
+ * Parser delegate which creates Infinispan Remote queries in form of {@link StringBuilder}s.
+ *
+ * @author Fabio Massimo Ercoli
+ */
+public class InfinispanRemoteQueryRendererDelegate extends SingleEntityQueryRendererDelegateKeepParameterPlaceholders<StringBuilder, InfinispanRemoteQueryParsingResult> {
```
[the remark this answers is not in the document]
fax4ever: I was thought about it too. Maybe we can delegate this activity to another issue. WDYT? And in this new issue we could perform the same optimization both for InfinispanRemote and Neo4j.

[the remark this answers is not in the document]
fax4ever: I was thought the same

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemoteQueryResolverDelegate.java
fax4ever: +1

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemoteQueryStringBuilderHelper.java
fax4ever: I believe that it is used in the classes:
* InfinispanRemoteQueryRendererDelegate
* InfinispanRemoteComparisonPredicate
* InfinispanRemoteLikePredicate
* InfinispanRemoteRangePredicate

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/impl/ProtostreamPayloadClosableIterator.java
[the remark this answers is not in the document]
DavideD: What I'm asking is if the result could be 10 null elements? I don't think it would happen but I'm checking with you.
Again, it is not about what I prefer, I'm asking you if that use case can happen

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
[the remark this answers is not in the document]
DavideD: What I mean is, why do you throw an exception. Can't we return the an InfinispanRemoteQueryDescriptor that represents the query instead of throwing an exception?  Does it take a lot of work? What do we need to check?

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemoteBasedQueryParserService.java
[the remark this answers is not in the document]
DavideD: OK, I didn't see that MongoDB already does that, let's keep it with the Base then. We can change both later. This one got my attention because the name is longer than the MongoDB one

core/src/main/java/org/hibernate/ogm/query/parsing/impl/SingleEntityQueryRendererDelegateKeepParameterPlaceholders.java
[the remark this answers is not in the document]
DavideD: > I'm completely agree about

Should be: "I completely agree"

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemoteBasedQueryParserService.java
fax4ever: For instance "SELECT p FROM org.java.package.Person p where p.username = 'Fabio'" became "FROM HibernateOGMGenerated.Person where user = 'Fabio'. Where the Entity field "username" is mapped as "user" as datastore column name.

DavideD: Thanks, that's what I wanted to know. Adding the tests will make this clearer.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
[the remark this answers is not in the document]
fax4ever: Sorry I did not understand the question. Yes we can

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/impl/ProtostreamPayloadClosableIterator.java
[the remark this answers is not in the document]
fax4ever: Sorry I see only now the point. I'm going to check if it possible with a test.

core/src/test/java/org/hibernate/ogm/backendtck/inheritance/tableperclass/family/TablePerClassInheritancePersistTest.java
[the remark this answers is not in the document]
fax4ever: @DavideD, you were right. I'm going to fix the commit.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/impl/ProtostreamPayloadClosableIterator.java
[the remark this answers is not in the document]
fax4ever: Yes it is possible to store null value. But is not possible having its in query result set, even if it was a bulk get. See [Test|https://github.com/fax4ever/infinispan-play/blob/master/infinispan-query-play/src/test/java/it/redhat/demo/test/QueryIT.java]::getBulk_simple.

fax4ever:
@DavideD I'm going to rebase commits. I would like to refactor in something like these:

1. OGM-1426 Support JPQL - remove unsupported cases from tests
2. OGM-1426 Support JPQL - update core components
3. OGM-1426 Support JPQL - implementing queries for Infinispan Remote
4. OGM-1426 Support JPQL - update docs
5. OGM-1463 Support Native Query

WDYT? Thanks!

DavideD:
Looks good

fax4ever:
@DavideD I rebased thank you!

documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
```diff
@@ -2720,6 +2706,16 @@ __OGM default configuration__ is transactional.
 We always recommend using transactional caches, at least unless you are confident that data consistency is not important for your use case.
 ====
 
+==== Remote Query Capability
+
+Hibernate OGM backend is able to run the queries it needs to materialize relations.
+From version __5.4.0.Alpha2__ the dialect is capable to translate JPQL or Criteria queries to the Infinispan remote queries.
```
gsmet: It was preexisting but it should be either:
 `to the **corresponding** Infinispan remote queries`
or
 `to Infinispan remote queries`

documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
```diff
@@ -2720,6 +2706,16 @@ __OGM default configuration__ is transactional.
...
+But there are still some limitations:
+
+* We don't support queries on polymorphic entities using TABLE_PER_CLASS inheritance strategy
+* Selecting from associations is not yet implemented
+* Query on Byte filed is not currently supported
```
gsmet: s/filed/fields/
s/is not currently supported/is currently unsupported/

documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
DavideD: Criteria?

documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
DavideD: What do you mean when you say `selecting from association`?

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/impl/InfinispanRemoteQueryHandler.java
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	public ClosableIterator<Tuple> executeBackendQuery(BackendQuery<InfinispanRemoteQueryDescriptor> backendQuery, QueryParameters queryParameters) {
+		InfinispanRemoteQueryDescriptor queryDescriptor = backendQuery.getQuery();
+		RemoteCache<ProtostreamId, ProtostreamPayload> cache = provider.getCache( queryDescriptor.getCache() );
+
+		QueryFactory queryFactory = Search.getQueryFactory( cache );
+		Query query = queryFactory.create( queryDescriptor.getText() );
+
+		applyNamedParamters( queryParameters, query );
```
DavideD: should be `applyNamedParameters`

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/impl/ProtostreamPayloadClosableIterator.java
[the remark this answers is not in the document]
DavideD: OK, thanks. Just wanted to double check

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemotePropertyHelper.java
```diff
@@ -0,0 +1,65 @@
...
+
+	@Override
+	public Object convertToBackendType(String entityType, List<String> propertyPath, Object value) {
+		if ( value instanceof InfinispanRemoteQueryParameter ) {
+			return value;
+		}
+
+		if ( value instanceof String ) {
+			return value;
+		}
```
DavideD: Any particular reason to check String? I guess it will work but it seems an optimization. Maybe I'm missing something but I'm concerned about use cases where the type should be converted to something else. What if it's a true/false that has to be stored as boolean? Will it work?

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemoteQueryBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+
+	public void append(InfinispanRemoteQueryBuilder child) {
+		builder.append( child.builder );
+	}
+
+	public void appendValue(Object value) {
+		boolean isConstant = value instanceof String;
+
+		if ( isConstant ) {
+			builder.append( "'" );
```
DavideD: Don't you need to check if the string contains `'` and escape them?

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/predicate/impl/InfinispanRemoteConjunctionPredicate.java
```diff
@@ -0,0 +1,55 @@
...
+		builder.append( "(" );
+		for ( Predicate<InfinispanRemoteQueryBuilder> child : children ) {
+			builder.append( ( (NegatablePredicate<InfinispanRemoteQueryBuilder>) child ).getNegatedQuery() );
+			builder.append( ")" );
+			if ( counter++ < children.size() ) {
+				builder.append( " or (" );
+			}
+		}
+		return builder;
+	}
```
DavideD: You removed the duplicated code in `InfinispanRemoteDisjunctionPredicate` but not here

DavideD:
@fax4ever I've added some comments. Thanks a lot.

There are some minor typos, but mainly a couple of comments about escaping values during the creation of the infinispan queries. Also, I don't see a test equivalent to `CypherQueryParsingTest` (or MongoDBQueryParsingTest). We need them to check how a HQL appear as Infinispan Remote query.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemoteQueryBuilder.java
DavideD: I'm not sure about the `isCostant` can you give me an example? I would expect something like isText instead. Maybe I'm missing something?

documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
DavideD: > What do you mean when you say `selecting from association`?

I understand now what you mean, I think a quick example will make it clearer. Something like:
Selecting from associations is not yet implemented (example: select h.author.name from Hypothesis h where author is a one-to-one association)

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemotePropertyHelper.java
fax4ever: Not anymore, it had been useful before we introduced the class InfinispanRemoteQueryParameter. So I'm going to remove it.

fax4ever:
@DavideD I changed **InfinispanRemoteJPQLParsingTest**. thank you

fax4ever:
I rebased on top of master, in order to be sure that there are no conflicts with recent commits. @DavideD or @gsmet, if you think that there are no more changes to do, you can merge it, if you want. Thank you guys

DavideD:
Checking it ...

DavideD:
I've sent a new PR on top of this one: https://github.com/hibernate/hibernate-ogm/pull/1022

PR #1013, opened 2018-05-03
OGM-1460 In OgmSessionImpl, don't override the getTransaction() method

DavideD:
https://hibernate.atlassian.net/browse/OGM-1460
  
When we upgraded to ORM 5.2 we had some failures because it wasn't possible anymore to
use the getTransaction() method in a JTA environment. We updated our tests since then and it
seems that this issue has been solved.

@fax4ever , this should be a quick one, let's just wait if it passes all the builds on CI.


fax4ever:
Merged, thank you.

gsmet:
@DavideD nice, that's one thing that bugged me in the ORM 5.2 upgrade.

fax4ever:
The overridden method do the same that overriding did

PR #1004, opened 2018-04-13
OGM-1430 Support scripting stored procedures for infinispan remote (#43)

ntviet18:
Hi @fax4ever @Sanne @DavideD could you please review this pull request? It's basically the same as #1000. But for Infinitispan Remote module.

PR #1014, opened 2018-05-04
 OGM-1431 Support server task stored procedures for infinispan remote

ntviet18:
@DavideD , @fax4ever  please review infinispan server task support by @QwerTech and me. It's based on #1004


DavideD:
Hi @ntviet18, should we close the other PR and only consider this one? Did you just added a commit or did you change something in the previous PR?

THanks

ntviet18:
Hi @DavideD , I think we can close the former request. Because it turned out my previous implementation already supports Infinispan server task execution. So, in this request I just add some test cases to make sure it works with deployables.

DavideD:
One question, how the jar get uploaded to the server? I'm talking about simple-value-test.jar, result-set-procedure.jar and so on.

infinispan-remote/infinispan-server-testconfig/wildfly-trimmed-config.xml
```diff
@@ -93,6 +93,9 @@
                 <pattern-formatter pattern="%d{yyyy-MM-dd HH:mm:ss,SSS} %-5p [%c] (%t) %s%e%n"/>
             </formatter>
         </subsystem>
+        <subsystem xmlns="urn:jboss:domain:deployment-scanner:2.0">
+            <deployment-scanner path="deployments" relative-to="jboss.server.base.dir" scan-interval="5000" runtime-failure-causes-rollback="${jboss.deployment.scanner.rollback.on.failure:false}"/>
```
ntviet18: @DavideD , we added scanner to make sure the jars are deployed to Infinispan server

infinispan-remote/pom.xml
```diff
@@ -181,6 +181,15 @@
     </dependencies>
 
     <build>
+        <testResources>
+            <testResource>
+                <directory>src/test/resources</directory>
+            </testResource>
+            <testResource>
+                <directory>src/test/resources/storedprocedures/servertask</directory>
+                <targetPath>${infinispan-server.home}/standalone/deployments/</targetPath>
```
ntviet18: @DavideD , and then we copy the jars from test resources to deployment directory

ntviet18: great idea, let's me try it

DavideD: OK, Go for it

infinispan-remote/infinispan-server-testconfig/wildfly-trimmed-config.xml
```diff
@@ -93,6 +93,9 @@
                 <pattern-formatter pattern="%d{yyyy-MM-dd HH:mm:ss,SSS} %-5p [%c] (%t) %s%e%n"/>
             </formatter>
         </subsystem>
+        <subsystem xmlns="urn:jboss:domain:deployment-scanner:2.0">
+            <deployment-scanner path="deployments" relative-to="jboss.server.base.dir" scan-interval="1" runtime-failure-causes-rollback="${jboss.deployment.scanner.rollback.on.failure:false}"/>
```
ntviet18: @DavideD , do we use this configuration in CI? It works for me locally, but somehow fails to rescan deployments in Jenkins.

DavideD: I will check but we don't do anything specific on Jenkins for the Infinispan Server. It just run the build the same way you'd run it locally

DavideD: If this branch works for you I can try it on my machine and let you know

DavideD: Can you access this link?
http://ci.hibernate.org/view/Pull%20Requests/job/hibernate-ogm-PR/ws/infinispan-remote/target/infinispan-server/

DavideD:
If you can access this link: http://ci.hibernate.org/view/Pull%20Requests/job/hibernate-ogm-PR/ws/infinispan-remote/target/infinispan-server/

You should be able to check the content of the server used for the tests.

ntviet18:
yes, I can. Let's me re-run the full build with the exact parameters

infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/utils/RemoteHotRodServerRule.java
```diff
@@ -69,6 +69,7 @@ public void before() throws Exception {
 		// e.g. the usage as JUnit Rule concurrently with the usage as global test listener in Surefire.
 		synchronized ( running ) {
 			if ( running.compareAndSet( false, true ) ) {
+				RemoteHotRodServerHelper.beforeRun();
```
fax4ever: this class is also used in integration tests module, where the server is in a different file system position

infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/test/storedprocedures/ResultSetProcedure.java
```diff
@@ -0,0 +1,47 @@
...
+ * @author The Viet Nguyen
+ */
+public class ResultSetProcedure implements ServerTask<List<Car>> {
+
+	private int id;
+	private String title;
+
+	@Override
+	public void setTaskContext(TaskContext taskContext) {
+		id = taskContext.getParameters()
```
fax4ever: Maybe could be nice having a similar behaviour of the corresponding js file: __resultSetProcedure.js__.
In term of write and read data from a real cache.

ntviet18: sure, added similar operations

infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/utils/InfinispanRemoteJpaServerRunner.java
```diff
@@ -32,6 +32,7 @@ public InfinispanRemoteJpaServerRunner(Class<?> klass) throws InitializationErro
 	@Override
 	public void run(RunNotifier notifier) {
 		try {
+			InfinispanRemoteTestHelper.deployJavaStoredProcedures();
```
ntviet18: @fax4ever , I moved the method call here. WDYT?

fax4ever: nice!

core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureOutputsImpl.java
```diff
@@ -54,6 +54,13 @@ public NoSQLProcedureOutputsImpl(NoSQLProcedureCallImpl procedureCall) {
 		this.procedureCall = procedureCall;
 		this.sessionFactory = procedureCall.getSession().getFactory();
 		this.gridDialect = this.sessionFactory.getServiceRegistry().getService( StoredProcedureAwareGridDialect.class );
+		//TODO: Check this "if"
```
fax4ever: I think we could remove TODO. If the service was not available gridDialect sould be null, from what I tested.

ntviet18: thank you, I will remove soon

fax4ever:
I tried to rebase from master, solving conflicts, on my local branch. But executing the test suite I get stuck on **ServerTaskInfinispanNamedParametersStoredProcedureCallTest**. Could someone else verify if the tests worked after the rebase? Thank you

ntviet18:
@fax4ever , mine also hangs when run `verify` with maven. I will try to figure it out soon.

ntviet18:
@DavideD , @fax4ever , I have fixed the jar deployment problems. Please review.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/InfinispanRemoteStoredProceduresManager.java
```diff
@@ -0,0 +1,90 @@
...
+import org.infinispan.client.hotrod.exceptions.HotRodClientException;
+
+/**
+ * @author The Viet Nguyen &amp;ntviet18@gmail.com&amp;
+ */
+public class InfinispanRemoteStoredProceduresManager {
+
+	private static final Log log = LoggerFactory.make( MethodHandles.lookup() );
+
+	private static final Pattern REFERENCE_ERROR_REGEXP = Pattern.compile( ".*\"([a-zA-Z_$]+)\" is not defined in <eval>.*" );
```
fax4ever: It is fine relying on regex on **ISPN 027002**, but I think we can not rely on message text. In general it could change or be different in other languages.

Sanne: That's right, logger IDs should be maintained stable across versions (and not reused). But the message text might be updated and could be translated.

infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/utils/InfinispanRemoteTestHelper.java
```diff
@@ -143,6 +171,50 @@ public long getNumberOfEntities(SessionFactory sessionFactory) {
 
 	// Various static helpers below:
 
+	public static void registerScriptStoredProcedures(SessionFactory sessionFactory) {
```
fax4ever: @ntviet18 what you think about the idea of moving new helper methods in a dedicated helper class? It is not critical, it is just to walk about it :P

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/InfinispanRemoteStoredProceduresManager.java
fax4ever: A general workaround if you can't parse it is to report the original datastore error message into upfront Hibernate OGM message.

fax4ever:
@ntviet18, just for curiosity, the problem was caused by the fact that javascript and java procedures had the same names? Otherwise can you explain it to us? thanks

core/src/test/java/org/hibernate/ogm/backendtck/storedprocedures/NamedParametersStoredProcedureCallTest.java
```diff
@@ -62,7 +62,7 @@
 	protected EntityManager em;
```
fax4ever: Can you add comment on top of the class, arguing that even if the *INFINISPAN_REMOTE* dialect is skipped in this test, the test is provided and run by the dialect module test, using both java and javascript procedure?

infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/utils/InfinispanRemoteTestHelper.java
```diff
@@ -178,23 +187,32 @@ public static void clearScriptStoredProcedures(SessionFactory sessionFactory) {
...
+				.exportTo( new File( SERVER_TASK_RESULT_SET_PROCEDURE_JAR ), true );
+		ShrinkWrap.create( JavaArchive.class, "exceptional-procedure.jar" )
 				.addClass( ExceptionalProcedure.class )
-				.addAsResource( asResource( SERVER_TASK_META_INF_RESOURCE_DIRECTORY + "/exceptional-procedure" ), SERVER_TASK_META_INF_TARGET_FILE )
+				.addAsResource( asResource( SERVER_TASK_EXCEPTIONAL_PROCEDURE_META_INF ), SERVER_TASK_META_INF_TARGET_FILE )
 				.as( ZipExporter.class )
-				.exportTo( new File( INFINISPAN_DEPLOYMENTS_DIRECTORY + "/exceptional-procedure.jar" ), true );
+				.exportTo( new File( SERVER_TASK_EXCEPTIONAL_PROCEDURE_JAR ), true );
+		while ( !javaStoredProceduresDeployed() ) {
+			TimeUnit.SECONDS.sleep( 1 );
```
DavideD: What happens if the storedProcedures are never deployed? 
Have a lookt at HotRodServerRule to see an example of how we keep track of the number of attempts so that we don't end in an infinite loop.

infinispan-remote/infinispan-server-testconfig/wildfly-trimmed-config.xml
```diff
@@ -94,7 +94,7 @@
             </formatter>
         </subsystem>
         <subsystem xmlns="urn:jboss:domain:deployment-scanner:2.0">
-            <deployment-scanner path="deployments" relative-to="jboss.server.base.dir" scan-interval="50" runtime-failure-causes-rollback="${jboss.deployment.scanner.rollback.on.failure:false}"/>
+            <deployment-scanner path="deployments" relative-to="jboss.server.base.dir" scan-interval="1" auto-deploy-exploded="true" runtime-failure-causes-rollback="${jboss.deployment.scanner.rollback.on.failure:false}"/>
```
DavideD: Why do you need this change? Some explanation in the commit message or in the file would help. I assume is because you need to deploy the stored procedures

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/InfinispanRemoteStoredProceduresManager.java
```diff
@@ -64,16 +61,9 @@ private Object execute( RemoteCache<Object, Object> executor, String storedProce
...
+				if ( msg.contains( UNKNOWN_TASK_ERROR_ID ) ) {
 					throw log.procedureWithResolvedNameDoesNotExist( storedProcedureName, e );
 				}
-				Matcher matcher = REFERENCE_ERROR_REGEXP.matcher( msg );
-				if ( matcher.find() ) {
-					String param = matcher.group( 1 );
-					if ( param != null && !param.isEmpty() ) {
-						throw log.cannotSetStoredProcedureParameter( storedProcedureName, param, namedParameters.get( param ), e );
-					}
-				}
```
DavideD: I would assume that this case has a specific error code. isn't it?

schernolyas: Yes. But I and @fax4ever think that it is excess.

ntviet18:
@fax4ever , the problem was caused by me forgetting `auto-deploy-exploded="true"`, in macOs it works well with or without this flag. However, in other systems we need to add `*.dodeploy` file to make a deployment. But, then I later decided to test both mechanisms separately, so the future maintainers won't have the frustration that I have had.

infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/utils/InfinispanRemoteTestHelper.java
ntviet18: sure, I think it's a good idea. I will move it

core/src/test/java/org/hibernate/ogm/backendtck/storedprocedures/NamedParametersStoredProcedureCallTest.java
ntviet18: sure, I will add that

fax4ever: thank you

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/InfinispanRemoteStoredProceduresManager.java
DavideD: Don't we have a test for this?

ntviet18: @DavideD , yes, we have a test for it. And it's failing because it wants `invalid parameters` but got `cannot execute` exception. Should we ignore it?

DavideD: I wouldn't skip it, but change so that it can catch the different exception. Does the error message should contain some information about what caused the error?

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/InfinispanRemoteStoredProceduresManager.java
ntviet18: @fax4ever , @Sanne , @schernolyas , I think it's the only way we know whether we used an undefined parameters or not. Because Infinispan at the moment doesn't seem to parse Nashorn errors properly.

> Caused by: org.infinispan.client.hotrod.exceptions.HotRodClientException:Request for messageId=5718 returned server error (status=0x85): javax.script.ScriptException: ReferenceError: "param" is not defined in <eval> at line number 2
> <eval>:2 ReferenceError: "param" is not defined
> 	at org.infinispan.client.hotrod.impl.protocol.Codec20.checkForErrorsInResponseStatus(Codec20.java:332)
> 	at org.infinispan.client.hotrod.impl.protocol.Codec20.readHeader(Codec20.java:178)
> 	at org.infinispan.client.hotrod.impl.transport.netty.HeaderDecoder.decode(HeaderDecoder.java:138)
> 	at 

According to `int SERVER_ERROR_STATUS = 0x85;`. The status tell us nothing about the actual cause.

fax4ever: Unfortunately Hot Rod return the same code in both situations. Anyway a possible wrong interpretation of the error would not cause unexpected errors, because the parsing in inside the catch. Temporary I think that you can put TODO, just for the moment. WDYT @DavideD?

DavideD: It's annoying that we cannot get the exact error but I don't think this issue is blocking. Let's just make sure that the tests we have check that an exception is thrown and then we can work it out later how to make it user friendly

Sanne: @ntviet18 thanks for the explanation, I didn't know that. Ok then let's keep this for now - we can ask the Infinispan team to improve on this and then we will be able to improve OGM.

Sanne: (I agree with you all)

DavideD:
OK, please, don't apply any more changes to this PR, I'm going to rebase it a bit and apply it to master

DavideD:
@ntviet18 Please, next time you send a PR, try to follow this rules when you write the commit messages: https://chris.beams.io/posts/git-commit/

ntviet18:
@DavideD , sure. Thank you

fax4ever:
great contribution!

Sanne:
This is awesome! Thanks, I'm eager to try it.

ntviet18:
Thank you everyone :)

PR #1015, opened 2018-05-07
OGM-1462 Support composite indexes in Neo4j

ntviet18:
https://hibernate.atlassian.net/browse/OGM-1462

Hi @DavideD , @gsmet , could you please review?

### DIFF
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jSchemaDefiner.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jSchemaDefiner.java
@@ -7,22 +7,28 @@
 package org.hibernate.ogm.datastore.neo4j.embedded.impl;
 
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 
 import org.hibernate.boot.model.relational.Sequence;
 import org.hibernate.cfg.Environment;
 import org.hibernate.ogm.datastore.neo4j.impl.BaseNeo4jSchemaDefiner;
+import org.hibernate.ogm.datastore.neo4j.index.impl.Neo4jIndexSpec;
 import org.hibernate.ogm.datastore.neo4j.logging.impl.Log;
 import org.hibernate.ogm.datastore.neo4j.logging.impl.LoggerFactory;
 import java.lang.invoke.MethodHandles;
 import org.hibernate.ogm.datastore.spi.DatastoreProvider;
 import org.hibernate.ogm.model.key.spi.IdSourceKeyMetadata;
+import org.hibernate.ogm.util.impl.CollectionHelper;
 import org.hibernate.tool.hbm2ddl.UniqueConstraintSchemaUpdateStrategy;
+
 import org.neo4j.graphdb.GraphDatabaseService;
 import org.neo4j.graphdb.Label;
 import org.neo4j.graphdb.Transaction;
 import org.neo4j.graphdb.schema.ConstraintDefinition;
 import org.neo4j.graphdb.schema.ConstraintType;
+import org.neo4j.graphdb.schema.IndexCreator;
+import org.neo4j.graphdb.schema.IndexDefinition;
 
 /**
  * Initialize the schema for the Neo4j database:
@@ -55,6 +61,18 @@ public class EmbeddedNeo4jSchemaDefiner extends BaseNeo4jSchemaDefiner {
 	}
 
 	@Override
+	protected void createIndexesIfMissing(DatastoreProvider provider, List<Neo4jIndexSpec> indexes ) {
+		EmbeddedNeo4jDatastoreProvider neo4jProvider = (EmbeddedNeo4jDatastoreProvider) provider;
+		GraphDatabaseService neo4jDb = neo4jProvider.getDatabase();
+		try ( Transaction tx = neo4jDb.beginTx() ) {
+			for ( Neo4jIndexSpec index : indexes ) {
+				createIndex( neo4jDb, index );
+			}
+			tx.success();
+		}
+	}
+
+	@Override
 	protected void createUniqueConstraintsIfMissing(DatastoreProvider provider, List<UniqueConstraintDetails> constraints) {
 		EmbeddedNeo4jDatastoreProvider neo4jProvider = (EmbeddedNeo4jDatastoreProvider) provider;
 		GraphDatabaseService neo4jDb = neo4jProvider.getDatabase();
@@ -70,6 +88,32 @@ public class EmbeddedNeo4jSchemaDefiner extends BaseNeo4jSchemaDefiner {
 				tx.close();
 			}
 		}
+	}
+
+	private void createIndex(GraphDatabaseService neo4jDb, Neo4jIndexSpec index) {
+		Label label = index.getLabel();
+		if ( isIndexMissing( neo4jDb, index ) ) {
+			log.tracef( "Creating composite index for nodes labeled as %1$s on properties %2$s", label, index.getProperties() );
+			IndexCreator indexCreator = neo4jDb.schema().indexFor( label );
+			for ( String property : index.getProperties() ) {
+				indexCreator = indexCreator.on( property );
+			}
+			indexCreator.create();
+		}
+		else {
+			log.tracef( "Composite index already exists for nodes labeled as %1$s on properties %2$s", label, index.getProperties() );
+		}
+	}
+
+	private boolean isIndexMissing(GraphDatabaseService neo4jDb, Neo4jIndexSpec indexDetails) {
+		Iterable<IndexDefinition> indexes = neo4jDb.schema().getIndexes( indexDetails.getLabel() );
+		for ( IndexDefinition index : indexes ) {
+			List<String> indexProperties = CollectionHelper.toList( index.getPropertyKeys() );
+			if ( Objects.equals( indexProperties, indexDetails.getProperties() ) ) {
+				return false;
+			}
+		}
+		return true;
 	}
 
 	private void createUniqueConstraint(GraphDatabaseService neo4jDb, UniqueConstraintDetails constraint) {
--- a/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/BaseNeo4jSchemaDefiner.java
+++ b/neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/BaseNeo4jSchemaDefiner.java
@@ -14,6 +14,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.hibernate.boot.model.relational.Database;
 import org.hibernate.boot.model.relational.Namespace;
@@ -23,18 +24,23 @@ import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.mapping.Column;
 import org.hibernate.mapping.Constraint;
 import org.hibernate.mapping.ForeignKey;
+import org.hibernate.mapping.Index;
 import org.hibernate.mapping.PrimaryKey;
 import org.hibernate.mapping.Table;
 import org.hibernate.mapping.UniqueKey;
+import org.hibernate.ogm.datastore.neo4j.index.impl.Neo4jIndexSpec;
 import org.hibernate.ogm.datastore.neo4j.logging.impl.Log;
 import org.hibernate.ogm.datastore.neo4j.logging.impl.LoggerFactory;
 import java.lang.invoke.MethodHandles;
 import org.hibernate.ogm.datastore.spi.BaseSchemaDefiner;
 import org.hibernate.ogm.datastore.spi.DatastoreProvider;
 import org.hibernate.ogm.model.key.spi.IdSourceKeyMetadata;
+import org.hibernate.ogm.util.impl.CollectionHelper;
 import org.hibernate.service.spi.ServiceRegistryImplementor;
 import org.hibernate.tool.hbm2ddl.UniqueConstraintSchemaUpdateStrategy;
+
 import org.jboss.logging.Logger.Level;
+
 import org.neo4j.graphdb.Label;
 
 /**
@@ -81,6 +87,13 @@ public abstract class BaseNeo4jSchemaDefiner extends BaseSchemaDefiner {
 	}
 
 	private static final Log log = LoggerFactory.make( MethodHandles.lookup() );
+
+	private List<Neo4jIndexSpec> indexSpecs = new ArrayList<>();
+
+	@Override
+	public void validateMapping(SchemaDefinitionContext context) {
+		validateIndexSpecs( context );
+	}
 
 	protected void createUniqueConstraints(DatastoreProvider provider, Database database) {
 		List<UniqueConstraintDetails> constraints = new ArrayList<>();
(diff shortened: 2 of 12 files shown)


neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/CompositeIndexDetails.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.List;
+
+import org.hibernate.ogm.datastore.neo4j.query.parsing.cypherdsl.impl.CypherDSL;
+
+import org.neo4j.graphdb.Label;
+
+/**
+ * @author The Viet Nguyen &amp;ntviet18@gmail.com&amp;
+ */
+public class CompositeIndexDetails {
```
gsmet: AFAICS, it also create mono-column indexes. I would call it `NeojIndexSpec` to be consistent with `MongoDBIndexSpec`.

(I'll let @DavideD do a proper review as he's more proficient in Neo4j than me)

gsmet:
@DavideD I suspect the Neo4j version we have on CI is too old for this.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/CompositeIndexDetails.java
DavideD: I think it's consistent with the existing classes in Neo4j, I will keep this in mind and change it later for the classes involved.

ntviet18: yes, it's aligned with `UniqueConstraintDetails` in Neo4j module

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/BaseNeo4jSchemaDefiner.java
```diff
@@ -23,20 +25,25 @@
...
-import java.lang.invoke.MethodHandles;
 import org.hibernate.ogm.datastore.spi.BaseSchemaDefiner;
 import org.hibernate.ogm.datastore.spi.DatastoreProvider;
 import org.hibernate.ogm.model.key.spi.IdSourceKeyMetadata;
 import org.hibernate.service.spi.ServiceRegistryImplementor;
 import org.hibernate.tool.hbm2ddl.UniqueConstraintSchemaUpdateStrategy;
+
 import org.jboss.logging.Logger.Level;
+
 import org.neo4j.graphdb.Label;
```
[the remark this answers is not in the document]
ntviet18: sorry, I forgot to disable auto organize imports

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/CompositeIndexDetails.java
DavideD: I just realized that this is different than unique constraints. Le't call it `Neo4jIndexSpec` as suggested by @gsmet. This way we also highlight that they are two different concepts

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/mapping/CompositeIndexTest.java
```diff
@@ -0,0 +1,80 @@
...
+		}
+
+		public void setLastname(String lastname) {
+			this.lastname = lastname;
+		}
+	}
+
+	@Test
+	public void test() {
+		// TODO: create test plans
```
DavideD: In MongoDB there are already a lot of test cases for this issue that you can copy and adapt.  MongoDB is more powerful in term of indexes you will need to pick the one that make sense for Neo4j.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/impl/EmbeddedNeo4jSchemaDefiner.java
```diff
@@ -72,6 +97,33 @@ protected void createUniqueConstraintsIfMissing(DatastoreProvider provider, List
 		}
 	}
 
+	private void createCompositeIndex(GraphDatabaseService neo4jDb, CompositeIndexDetails index) {
```
gsmet: Unless if I'm mistaken (and I could be :)), I don't think we should call these methods `Composite` as if it targets only one index, it's a regular index, not a composite one.

So let's rename them to `createIndex()`/`isIndexMissing()`.

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/utils/Neo4jTestHelper.java
```diff
@@ -137,6 +151,50 @@ private static DatastoreProvider getDatastoreProvider(SessionFactory sessionFact
...
+			List<Neo4jIndexSpec> indexes = toStream( neo4jDb.schema().getIndexes() )
+					.map( i -> new Neo4jIndexSpec( i.getLabel(), toList( i.getPropertyKeys() ), i.isConstraintIndex() ) )
+					.collect( Collectors.toList() );
+			tx.success();
+			return indexes;
+		}
+	}
+
+	private static List<Neo4jIndexSpec> getHttpNeo4jIndexes(DatastoreProvider provider) {
+		// todo: neo4jHttpClient does not return graph values
```
ntviet18: @DavideD , I'm trying to get indexes by CALL db.indexes via HttpNeo4jClient. However it returns
```
{
  "results" : [ {
    "columns" : [ "description", "label", "properties", "state", "type", "provider" ],
    "data" : [ {
      "row" : null,
      "graph" : {
        "nodes" : [ ],
        "relationships" : [ ]
      },
      "meta" : null
    },... ],
  "errors" : [ ]
}
```

in contrast of direct rest response

```
{
    "results": [
        {
            "columns": [ "description",  "label", "properties", "state", "type", "provider" ],
            "data": [
                {
                    "row": [
                        "INDEX ON :Neo4jIndexTest$Person(age)",
                        "Neo4jIndexTest$Person",
                        [
                            "age"
                        ],
                        "ONLINE",
                        "node_label_property",
                        {
                            "version": "1.0",
                            "key": "lucene+native"
                        }
                    ]
                }, ...
    ],
    "errors": []
}
```

do we have any filter somewhere?

ntviet18: Never mind, I have found my mistake. There are two modes for the query to work. I changed to `row` mode and everything works as expected.

ntviet18:
@DavideD , @gsmet , could you please re-check? I have added some positive and negative test cases for all Neo4j provider types. It works locally for me with Neo4j v3.2.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/Neo4jIndexSpec.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Objects;
+
+import org.hibernate.ogm.datastore.neo4j.query.parsing.cypherdsl.impl.CypherDSL;
+
+import org.neo4j.graphdb.Label;
+
+/**
+ * @author The Viet Nguyen
+ */
+public class Neo4jIndexSpec {
```
fax4ever: could you add implements Serializable?

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/index/Neo4jIndexTest.java
```diff
@@ -0,0 +1,216 @@
...
+import org.neo4j.graphdb.Label;
+
+import static org.fest.assertions.Assertions.assertThat;
+import static org.hibernate.ogm.datastore.neo4j.test.util.ExceptionHelper.extract;
+
+/**
+ * @author The Viet Nguyen
+ */
+@TestForIssue(jiraKey = "OGM-1462")
+public class Neo4jIndexTest extends Neo4jJpaTestCase {
```
fax4ever: Could you add a test where the index has applied on column, where the **column** name it is different from **field** name?

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/index/Neo4jIndexTest.java
```diff
@@ -0,0 +1,216 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import org.neo4j.graphdb.Label;
+
+import static org.fest.assertions.Assertions.assertThat;
+import static org.hibernate.ogm.datastore.neo4j.test.util.ExceptionHelper.extract;
+
+/**
```
fax4ever: Could you add some text on JavaDoc expanding the **scope** of the test?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/http/impl/HttpNeo4jSchemaDefiner.java
```diff
@@ -53,6 +54,19 @@ protected void createSequences(List<Sequence> sequences, Set<IdSourceKeyMetadata
...
 
+	@Override
+	protected void createIndexesIfMissing(DatastoreProvider provider, List<Neo4jIndexSpec> indexes) {
+		Statements statements = new Statements();
+		for ( Neo4jIndexSpec index : indexes ) {
+			log.tracef( "Creating composite index for nodes labeled as %1$s on properties %2$s", index.getLabel(), index.getProperties() );
+			statements.addStatement( new Statement( index.asCypherQuery() ) );
+		}
+		log.debug( "Creating missing indexes" );
+		HttpNeo4jClient remoteClient = ( (HttpNeo4jDatastoreProvider) provider ).getClient();
```
fax4ever: When we close the instance of HttpNeo4jClient?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/Neo4jIndexSpec.java
```diff
@@ -0,0 +1,88 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+
+package org.hibernate.ogm.datastore.neo4j.impl;
```
fax4ever: Maybe you can move this class from *.neo4j.impl to *.neo4j.index.impl to be consistent with the same class of MongoDB?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/BaseNeo4jSchemaDefiner.java
```diff
@@ -82,6 +87,35 @@ public String toString() {
 
 	private static final Log log = LoggerFactory.make( MethodHandles.lookup() );
 
+	protected void createIndexes(DatastoreProvider provider, Database database) {
+		List<Neo4jIndexSpec> indexes = new ArrayList<>();
```
fax4ever: Maybe we can refactor the **indexes** local variable as a Class field, like in MongoDBSchemaDefiner class.
You could fill the list on validateMapping and create the indexes on initializeSchema, like in MongoDBSchemaDefiner. WDYT?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/Neo4jIndexSpec.java
ntviet18: sure, I moved the tests but for got the main class 😃

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/Neo4jIndexSpec.java
ntviet18: sure, but may I ask why should we do that? We don't move this anywhere outside the JVM yet. And, I don't see `MongoDBIndexSpec` implements `Serialzable` either

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/http/impl/HttpNeo4jSchemaDefiner.java
ntviet18: I think it's when we close the `HttpNeo4jDatastoreProvider`

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/index/Neo4jIndexTest.java
ntviet18: sure

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/index/Neo4jIndexTest.java
ntviet18: interesting case, I sure will add it

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/Neo4jIndexSpec.java
fax4ever: It's an optional, sometimes it is better to mark a class Serializable, it a way to describe the nature of the class itself.

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/index/Neo4jIndexTest.java
fax4ever: Thank you

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/http/impl/HttpNeo4jSchemaDefiner.java
fax4ever: you're right :P

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/impl/Neo4jIndexSpec.java
ntviet18: I get it. Will add in the next push

fax4ever: Now we are more focus on what is Serializable or not, more than in the past, for cloud compliance :P

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/index/Neo4jIndexTest.java
ntviet18: I have checked, this case is already covered by hibernate-core

ntviet18:
Hello everyone, could you please tell me what is the version we are using on CI?

DavideD:
> Hello everyone, could you please tell me what is the version we are using on CI?

Hi, I'm upgrading to 3.3.5 but I'm having some issues with the buid right now. Working on it. If you see some failure, it's likely not because of this PR.

DavideD:
Thanks a lot @ntviet18 ,
I'm still finishing the review but overall it looks good.

The only real problem I have with this PR is that you squashed all changes in a single commit. We always try to send PR as a collection of commits containing logical changes. So far I can see at least 3 commits: one for the creation of the ExceptionHjelper class,  one for the fix and one for the changes related to the tests used to check that the indexes actually exist no the datastore.

Having these changes separated has several advantages:
* it makes it easier to understand what the changes are about and therefore it's easier to review them;
* if the tests are in a separate commit I can run them on master and verify that they really fail without the fix;
* if a group of changes are not right or not exactly what we need I can merge the rest of the fix quite easily meaning that it takes me less time to merge a PR, or just update the commit that is not right;
* having new classes in separate commits (if possible) helps to highlight what's the new stuff that has been introduced and make it easier to rebase when needed;
* if I want to apply some of the changes to a different branch I can just cherry-pick the related commits;
* if we realize that something is not right in the future we might be able to simply revert the commit invoved;
* we might decide that some changes can be promoted to a new JIRA;
* it's easier to squashed commits together if needed than split an existing commit.

We still want the project to build for every commit and this means that sometimes it's not possible to completely keep the changes separated (for example, tests cannot be kept separated or need to be included after the fix). But we still try.

If it's not clear what I mean, I would suggest to have a look at the commit log. You can also check the other PR in the https://github.com/hibernate/hibernate-search or https://github.com/hibernate/hibernate-validator repositories. 

***No need to change this PR now*** but please keep this in mind for the next PR.

Thanks a lot again for the great work.

I will now continue the review ...

ntviet18:
@DavideD , sorry about that. I used a custom git-review tool that lazily squash all commits together. I will break commit to parts in the future.

DavideD:
Just a note to clarify that I haven't forgotten this PR. I'm having some problems upgrading Neo4j to version 3.3.5 and I'm working on it.

DavideD:
This PR requires some changes, I'm going to close it for now and send a new one soon build on top of this one.

Cheers

PR #1019, opened 2018-05-14
OGM-1473 Make our pom files more consistent with WildFly and JBoss parent pom files

gsmet:
* https://hibernate.atlassian.net/browse/OGM-1473

As you can imagine, it was a lot of fun :).

It's the exact same work I did on Search so Search and OGM will be consistent once merged. I still have to do this work on HV.

Let's wait for CI to run as I didn't test all the Neo4j combinations.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -140,7 +140,6 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <forkMode>once</forkMode>
-                    <skipExec>${skipUnitTests}</skipExec>
                     <!-- Apache Lucene uses assertions which currently fail on JDK9: -->
                     <!-- not sure yet how that is going to be resolved, but it's not an OGM problem. -->
                     <enableAssertions>false</enableAssertions>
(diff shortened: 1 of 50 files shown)


documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
```diff
@@ -2203,7 +2203,7 @@ in any NoSQL store supported by Infinispan:
 
 How to configure it? Here is a simple cheat sheet to get you started with this type of setup:
 
-* Add `org.infinispan:infinispan-directory-provider:{infinispanVersion}` to your dependencies
+* Add `org.infinispan:infinispan-directory-provider:{infinispan-version}` to your dependencies
```
gsmet: This one was a leftover of a previous cleanup.

core/pom.xml
```diff
@@ -140,7 +140,6 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <forkMode>once</forkMode>
-                    <skipExec>${skipUnitTests}</skipExec>
```
gsmet: We should use the traditional `-DskipTests` instead.

DavideD:
Thanks @gsmet 
Great work!

It's really cool to see a cleaned-up POM

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/1019

PR #1022, opened 2018-05-17
OGM-1426 InfinispanRemote native queries follow up

DavideD:
Follows https://github.com/hibernate/hibernate-ogm/pull/1012
 
https://hibernate.atlassian.net/browse/OGM-1426
https://hibernate.atlassian.net/browse/OGM-1463

This is a refactoring of the initial PR sent by @fax4ever:

* Reduce the amount of regular expressions needed for the parser
* Add new test cases to check queries with strange cases and use of spaces
* Give additional information when the query is invalid
* Test that expected exceptions are thrown

Using regular expressions for the parser scares me a bit, but it seems to work for now because we don't need to extract lot of info from the native query. We should consider switching to a proper parser in the future like Parboiled.

Please @fax4ever, have a look at this and if you find something that doesn't make sense let me know.
Otherwise, let's merge it and close this issue.

Thanks a lot

### DIFF
--- a/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemoteBasedQueryParserService.java
+++ b/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemoteBasedQueryParserService.java
@@ -34,7 +34,7 @@ public class InfinispanRemoteBasedQueryParserService extends BaseQueryParserServ
 
 	@Override
 	public QueryParsingResult parseQuery(SessionFactoryImplementor sessionFactory, String queryString, Map<String, Object> namedParameters) {
-		throw new UnsupportedOperationException( "The Infinispan Remote query parser supports parameterized queries" );
+		throw new UnsupportedOperationException( "The Infinispan Remote query parser doesn't support parameterized queries" );
 	}
 
 	@Override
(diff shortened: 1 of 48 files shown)


infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/query/parsing/impl/InfinispanRemoteBasedQueryParserService.java
```diff
@@ -34,7 +34,7 @@
 
 	@Override
 	public QueryParsingResult parseQuery(SessionFactoryImplementor sessionFactory, String queryString, Map<String, Object> namedParameters) {
-		throw new UnsupportedOperationException( "The Infinispan Remote query parser supports parameterized queries" );
+		throw new UnsupportedOperationException( "The Infinispan Remote query parser doesn't support parameterized queries" );
```
fax4ever: I think that supports was right. Like in **Neo4jBasedQueryParserService**.
In a nutshell: the client driver supports parameters natively, like in Neo4j, unlike MongoDB. Thus, from what I understood, other method must be called, because the passer supports parameterized queries.

DavideD: > Thank you @DavideD, reducing the amount of regular expressions needed for the parser was very good. I have just a minor doubt about the 5642f87.

Thanks @fax4ever,. Got it. Could you remove that commit and merge everything else, please? The message seems misleading but that's an issue for another time.

fax4ever: Sure!

fax4ever:
Rebased and merged (all except 5642f87)! Thanks ;)

PR #1024, opened 2018-05-21
Ogm 1222 mongodb object mutation test

ntviet18:
https://hibernate.atlassian.net/browse/OGM-1222

### DIFF
--- a/mongodb/pom.xml
+++ b/mongodb/pom.xml
@@ -112,6 +112,12 @@
             <artifactId>jboss-transaction-api_1.2_spec</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>de.flapdoodle.embed</groupId>
+            <artifactId>de.flapdoodle.embed.mongo</artifactId>
+            <version>2.0.3</version>
+            <scope>test</scope>
+        </dependency>
         <!-- Required by PackagingRule -->
         <dependency>
             <groupId>org.jboss.shrinkwrap</groupId>
--- a/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/InsertTest.java
+++ b/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/InsertTest.java
@@ -1,0 +1,71 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+
+package org.hibernate.ogm.datastore.mongodb.test;
+
+import javax.persistence.Entity;
+import javax.persistence.Id;
+
+import org.hibernate.ogm.datastore.mongodb.utils.MongoDbServerRunner;
+import org.hibernate.ogm.utils.OgmTestCase;
+import org.hibernate.ogm.utils.TestForIssue;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * @author The Viet Nguyen
+ */
+@TestForIssue(jiraKey = "OGM-1222")
+@RunWith(MongoDbServerRunner.class)
+public class InsertTest extends OgmTestCase {
+
+	@Test
+	public void testModifyObjectAfterPersisting() {
+		inTransaction( em -> {
+			Subject subject = new Subject( "1", "name" );
+			em.persist( subject );
+			subject.setName( "name2" );
+			em.flush();
+		} );
+	}
+
+	@Override
+	protected Class<?>[] getAnnotatedClasses() {
+		return new Class[] { Subject.class };
+	}
+
+	@Entity(name = "subject")
+	static class Subject {
+
+		@Id
+		private String id;
+
+		private String name;
+
+		public Subject(String id, String name) {
+			this.id = id;
+			this.name = name;
+		}
+
+		public String getId() {
+			return id;
+		}
+
+		public void setId(String id) {
+			this.id = id;
+		}
+
+		public String getName() {
+			return name;
+		}
+
+		public void setName(String name) {
+			this.name = name;
+		}
+	}
+}
--- a/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/InsertTest.java
+++ b/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/InsertTest.java
@@ -24,6 +24,10 @@ import org.junit.runner.RunWith;
 @RunWith(MongoDbServerRunner.class)
 public class InsertTest extends OgmTestCase {
 
+	/**
+	 * Verify flush does not duplicate Entity on changing field, even if we are in
+	 * the same transaction that creates the Entity itself.
+	 */
 	@Test
 	public void testModifyObjectAfterPersisting() {
 		inTransaction( em -> {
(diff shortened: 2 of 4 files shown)


mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/InsertTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * @author The Viet Nguyen
+ */
+@TestForIssue(jiraKey = "OGM-1222")
+@RunWith(MongoDbServerRunner.class)
+public class InsertTest extends OgmTestCase {
```
DavideD: What's the purpose of this test? We already test insertion and persisting I think

ntviet18: This test was submitted by an user. He thinks there is an error if we persist then mutate the entity. I tried to reproduce, but it seems already fixed.

DavideD: Ah OK, thanks.
OK, Thanks

mongodb/pom.xml
```diff
@@ -112,6 +112,12 @@
             <artifactId>jboss-transaction-api_1.2_spec</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>de.flapdoodle.embed</groupId>
+            <artifactId>de.flapdoodle.embed.mongo</artifactId>
+            <version>2.0.3</version>
+            <scope>test</scope>
+        </dependency>
```
fax4ever: Even if the dependency is used just in mongoDb module, we usually manage the version in parent pom. For instance take a look to:
```xml
<dependency>
  <groupId>org.skyscreamer</groupId>
  <artifactId>jsonassert</artifactId>
  <version>${version.org.skyscreamer.jsonassert}</version>
  <scope>test</scope>
</dependency>
```

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/InsertTest.java
fax4ever: Nice test. It seems the same case reported in the issue.
Could you add comment in javadoc like "Verify flush does not duplicate an entity on changing field, in the same transaction we create it". You could improve the text of the comment :P

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/utils/EmbeddedMongoDbServerRule.java
```diff
@@ -0,0 +1,95 @@
...
+				running.set( false );
+				mongodExecutable.stop();
+			}
+		}
+	}
+
+	private boolean portOpen() {
+		try {
+			Socket sock = new Socket( MONGODB_HOSTNAME, MONGODB_PORT );
+			sock.close();
```
fax4ever: Maybe we can move this line a finally clause. WDYT?

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/utils/EmbeddedMongoDbServerRule.java
```diff
@@ -0,0 +1,95 @@
...
+	private static final String MONGODB_HOSTNAME = env.getOrDefault( "MONGODB_HOSTNAME", "localhost" );
+	private static final int MONGODB_PORT = Integer.parseInt( env.getOrDefault( "MONGODB_PORT", "27018" ) );
+
+	private MongodExecutable mongodExecutable;
+
+	@Override
+	protected void before() throws Exception {
+		synchronized (running) {
+			if ( running.compareAndSet( false, true ) && portOpen() ) {
+				log.info( "Starting embedded MongoDb server..." );
```
fax4ever: What happens if the method portOpen() returns false?

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/InsertTest.java
fax4ever: Maybe "Verify flush does not duplicate Entity on changing field, even if we are in the same transaction that creates the Entity itself".

mongodb/pom.xml
gsmet: TBH, I'm a bit surprised we need this? Don't we already start a MongoDB server when we run the tests? Why would we need another one?

ntviet18: @fax4ever , thank you

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/InsertTest.java
fax4ever: It's a minor issue... but in this way the test will be more readable from outside.

mongodb/pom.xml
ntviet18: @gsmet , I think it's a little bit more convenient if we can run the tests directly from IDE. WDYT?

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/utils/EmbeddedMongoDbServerRule.java
ntviet18: AFAIR, the exception is thrown when we `close`. Let's me recheck.

mongodb/pom.xml
gsmet: Ah yes, usually, I start a MongoDB server for that.

If we want to do that, we should be consistent and do it everywhere. It's weird to have only one test working like that.

@DavideD I suppose it's you who should decide what you prefer for the test infrastructure.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/utils/EmbeddedMongoDbServerRule.java
ntviet18: In this case, nothing because we assume that MongoDb has already ran, e.g. by maven plugin. In a worse case, some process other than MongoDb acquired that port. Do you think we should log a warning?

fax4ever: Ah ok, if the port is busy it means that the server is up and running by maven. So nothing it is necessary. It's ok sorry.

mongodb/pom.xml
DavideD: I agree with @gsmet on this: if we use this approach for one test we should try to be consistent.
The problem that I see with this PR is that II wouldn't expect an issue about adding a specific test to also change the way we run the tests in the project. It seems a task big enough to deserve a separate issue. 

Could you update this PR for OGM-1222 (by removing the mongodb server runner) and create new JIRA and PR with the MongoDBserverRunner that also updates the existing tests?

@ntviet18, Would you find it surprising to be able to run this test without having a server running but not the others?

I also noticed that we have  a HotRodServerLifeCycle in the infinispan-remote project . It is bound to the surefire plugin, do we need something similar here as well?

Thanks

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/InsertTest.java
```diff
@@ -0,0 +1,75 @@
...
+	 * the same transaction that creates the Entity itself.
+	 */
+	@Test
+	public void testModifyObjectAfterPersisting() {
+		inTransaction( em -> {
+			Subject subject = new Subject( "1", "name" );
+			em.persist( subject );
+			subject.setName( "name2" );
+			em.flush();
+		} );
```
DavideD: Shouldn't you load the on object in a separate transaction and check that it's changed? How do you know that it's saved as "named2" and not just ignored?

DavideD:
@ntviet18 I would like to close this PR. I'm going to send a new one with only the test. Feel free to send a new PR for the MongoDB rule associated with a JIRA issue, thanks

PR #1025, opened 2018-05-21
OGM-1468 Support merge and refresh loaders in OgmEntityPersister

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1468

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
+++ b/core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
@@ -34,12 +34,14 @@ import org.hibernate.engine.OptimisticLockStyle;
 import org.hibernate.engine.internal.Versioning;
 import org.hibernate.engine.jdbc.env.spi.JdbcEnvironment;
 import org.hibernate.engine.jdbc.spi.JdbcServices;
+import org.hibernate.engine.spi.CascadingActions;
 import org.hibernate.engine.spi.EntityEntry;
 import org.hibernate.engine.spi.LoadQueryInfluencers;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.internal.DynamicFilterAliasGenerator;
 import org.hibernate.internal.FilterAliasGenerator;
+import org.hibernate.loader.entity.CascadeEntityLoader;
 import org.hibernate.loader.entity.UniqueEntityLoader;
 import org.hibernate.mapping.Column;
 import org.hibernate.mapping.PersistentClass;
@@ -909,19 +911,15 @@ public abstract class OgmEntityPersister extends AbstractEntityPersister impleme
 		loaders.put( LockMode.OPTIMISTIC, createEntityLoader( LockMode.OPTIMISTIC ) );
 		loaders.put( LockMode.OPTIMISTIC_FORCE_INCREMENT, createEntityLoader( LockMode.OPTIMISTIC_FORCE_INCREMENT ) );
 
-		//FIXME handle cascading merge and refresh
 		loaders.put(
 				"merge",
-				createEntityLoader( LockMode.READ )
-				//new CascadeEntityLoader( this, CascadingAction.MERGE, getFactory() )
+				new CascadeEntityLoader( this, CascadingActions.MERGE, getFactory() )
 			);
 		loaders.put(
 				"refresh",
-				createEntityLoader( LockMode.READ )
-				//new CascadeEntityLoader( this, CascadingAction.REFRESH, getFactory() )
+				new CascadeEntityLoader( this, CascadingActions.REFRESH, getFactory() )
 			);
 	}
-
 
 	// TODO copied from AbtractEntityPersister: change the visibility
 	@Override
--- a/core/src/test/java/org/hibernate/ogm/utils/OgmTestCase.java
+++ b/core/src/test/java/org/hibernate/ogm/utils/OgmTestCase.java
@@ -10,6 +10,7 @@ import static org.fest.assertions.Assertions.assertThat;
 import static org.hibernate.ogm.utils.TestHelper.getNumberOfAssociations;
 import static org.hibernate.ogm.utils.TestHelper.getNumberOfEntities;
 
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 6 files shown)


core/src/test/java/org/hibernate/ogm/test/cascade/CascadeEntityLoaderTest.java
```diff
@@ -0,0 +1,150 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.test.cascade;
```
DavideD: Shouldn't this test run for all the datastores?

core/src/test/java/org/hibernate/ogm/utils/OgmTestCase.java
```diff
@@ -123,4 +124,14 @@ public void inTransaction(Consumer<Session> consumer) {
 		}
 	}
 
+	protected <T> void deleteAll(Class<T> entity, Serializable... ids) {
```
DavideD: Can you just call this `delete`? I prefer to keep the suffix `All` only for cases where it's not possible to specify the ids.

core/src/test/java/org/hibernate/ogm/test/cascade/CascadeEntityLoaderTest.java
fax4ever: Yes, it should. I'm going to move the package.

core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
```diff
@@ -909,20 +911,16 @@ protected void createLoaders() {
...
 				"merge",
-				createEntityLoader( LockMode.READ )
-				//new CascadeEntityLoader( this, CascadingAction.MERGE, getFactory() )
+				new CascadeEntityLoader( this, CascadingActions.MERGE, getFactory() )
 			);
 		loaders.put(
 				"refresh",
-				createEntityLoader( LockMode.READ )
-				//new CascadeEntityLoader( this, CascadingAction.REFRESH, getFactory() )
+				new CascadeEntityLoader( this, CascadingActions.REFRESH, getFactory() )
```
gsmet: With these changes, I think we can remove the whole overriding `createLoaders()` method and use the one from ORM. That was my ultimate goal for this issue as I made significant changes to ORM in this area and I would like these changes to come for free to OGM.

Could you try and check?

core/src/test/java/org/hibernate/ogm/utils/OgmTestCase.java
fax4ever: delete was my first choice too. Unfortunately many implementations of *OGMBaseTest* already use that name :(

core/src/test/java/org/hibernate/ogm/test/cascade/CascadeEntityLoaderTest.java
```diff
@@ -0,0 +1,150 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+/**
+ * Testing merge and refresh {@link CascadeEntityLoader}.
+ * Those loader coming in action on subsequent Entity load from data store,
+ * when merging detached Entity or refreshing attached Entity,
+ * performing operation on the Entity itself or from Parent Entity enabling cascade.
+ *
```
gsmet: This comment is very unclear. I'm not even sure to understand what you mean. Could you maybe try to make shorter sentences?

core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
fax4ever: Very good idea. Remove code it is always better ;) I'm going to check if it really possible

core/src/test/java/org/hibernate/ogm/test/cascade/CascadeEntityLoaderTest.java
fax4ever: You're right. I'm going to cut and rephrase it.

fax4ever:
Thank you @DavideD and @gsmet. I'm going to apply the change you've proposed.

core/src/test/java/org/hibernate/ogm/utils/OgmTestCase.java
fax4ever: WDYT about deleteEntities?

DavideD: No need to change it, leave it `deleteAll`

fax4ever: ok

core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
fax4ever: @gsmet currently is it not possible to remove the whole method. If we wanted this, we should resize the scope of the issue, adding more changes on it, or alternatively we could open a follow up issue.

fax4ever:
I rebased it, with some changes.

core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
gsmet: What's the issue exactly? You're too vague to convince me :).

fax4ever: :) nice approach. In the current implementation of Hibernate ORM, *noneLockLoader* and r*eadLockLoader* are moved out from the loaders map. So in order to use the whole base method we have to change the getAppropriateLoader too, leading us out of scope of OGM-1376, from what I think.

gsmet: Hum, yes, you're right, `getLoaderByLockMode()` should have been accessible to the child classes. Bummer.

I opened https://hibernate.atlassian.net/browse/HHH-12615 to change that.

gsmet:
@fax4ever you can merge it.

We'll remove the method once https://hibernate.atlassian.net/browse/HHH-12615 is fixed.

core/src/main/java/org/hibernate/ogm/persister/impl/OgmEntityPersister.java
fax4ever: nice followup!

fax4ever:
Merged! Thank you guys for review

PR #1028, opened 2018-05-24
OGM-1474 Replace CREATE UNIQUE in Cypher queries with MERGE

DavideD:
Running CREATE UNIQUE queries will generate the following warning:

  ```
  The RULE planner is not available in the current CYPHER version,
  the query has been run by an older CYPHER version. CREATE UNIQUE is unsupported
  for current CYPHER version, the query has been execute by an older CYPHER version
  ```

https://hibernate.atlassian.net/browse/OGM-1474


fax4ever:
Merged #1028. Thanks

PR #1034, opened 2018-05-31
OGM-1491 adding custom Identity Genetator test

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1491

### DIFF
--- a/core/src/test/java/org/hibernate/ogm/backendtck/id/gen/CustomIdGeneratorTest.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/id/gen/CustomIdGeneratorTest.java
@@ -1,0 +1,58 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.backendtck.id.gen;
+
+import static org.hibernate.ogm.utils.OgmAssertions.assertThat;
+
+import java.util.List;
+
+import org.hibernate.id.IdentifierGenerator;
+import org.hibernate.ogm.utils.jpa.OgmJpaTestCase;
+
+import org.junit.Test;
+
+/**
+ * Test user defined {@link IdentifierGenerator}
+ */
+public class CustomIdGeneratorTest extends OgmJpaTestCase {
+
+	private CustomIdGeneratorEntity customIdGeneratorEntityNo1 = new CustomIdGeneratorEntity( CustomIdGenerator.ID_PREFIX + 1 );
+	private CustomIdGeneratorEntity customIdGeneratorEntityNo2 = new CustomIdGeneratorEntity( CustomIdGenerator.ID_PREFIX + 2 );
+	private CustomIdGeneratorEntity customIdGeneratorEntityNo3 = new CustomIdGeneratorEntity( CustomIdGenerator.ID_PREFIX + 3 );
+
+	@Test
+	public void testIdsFromCustomGenerator() {
+		// persisting one entity
+		inTransaction( em -> {
+			em.persist( new CustomIdGeneratorEntity() );
+		} );
+		inTransaction( em -> {
+			List<CustomIdGeneratorEntity> resultList = em.createQuery( "from CustomIdGeneratorEntity" ).getResultList();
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/hibernate/ogm/backendtck/id/gen/CustomIdGeneratorTest.java
```diff
@@ -0,0 +1,42 @@
...
+		inTransaction( em -> {
+			em.persist( new CustomIdGeneratorEntity() );
+		} );
+
+		inTransaction( em -> {
+			List<CustomIdGeneratorEntity> resultList = em.createQuery( "from CustomIdGeneratorEntity" ).getResultList();
+			assertThat( resultList ).hasSize( 1 );
+
+			CustomIdGeneratorEntity entity = resultList.get( 0 );
+			assertThat( entity.id ).isNotEmpty();
```
DavideD: How do you know that the id is generated by your custom generator and not by something else?

We are only inserting one element so I think we can return a constant and check that the id value is the one generated by the custom generator

core/src/test/java/org/hibernate/ogm/backendtck/id/gen/CustomIdGeneratorTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+	@Test
+	public void test() {
+		inTransaction( em -> {
+			em.persist( new CustomIdGeneratorEntity() );
+		} );
+
+		inTransaction( em -> {
+			List<CustomIdGeneratorEntity> resultList = em.createQuery( "from CustomIdGeneratorEntity" ).getResultList();
+			assertThat( resultList ).hasSize( 1 );
```
DavideD: I would prefer not having a check on the size because the library we use allows you to check the content of a collection using `assertThat( resultList ).containsExactly( ... )` or, if you are only interested to the  id (like in this case), `assertThat( resultList ).onProperty( "id" ).containsExactly( ...)`. This generates better messages if/when the test fails.

core/src/test/java/org/hibernate/ogm/backendtck/id/gen/CustomIdGeneratorTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.junit.Test;
+
+/**
+ * Test user defined {@link IdentifierGenerator}
+ */
+public class CustomIdGeneratorTest extends OgmJpaTestCase {
+
+	@Test
+	public void test() {
```
DavideD: not critical, but let's try to use a better name for this method. `testidFromCustomGenerator`?

core/src/test/java/org/hibernate/ogm/backendtck/id/gen/CustomIdGeneratorTest.java
DavideD: By the way, I know that you have copied the test from the issue, but I don't think that problem was caused by the Random generation of values. 

I also prefer not having random generation in tests, because if it causes a failure we might not be able to identify which value was used.

DavideD: Using a costant also works for the user scenario, because an error should occur when trying to insert two entities with the same id.

DavideD: > Using a costant also works for the user scenario, because an error should occur when trying to insert two entities with the same id.

Thinking again about this, it might not be a good idea to have a constant because some datastore might decide to simply override the existing value because they don't support unique constraints.

I think a better approach would be to create a sequence generator that generates id looking something like: "custom-id-001", "custom-id-002",... 
The test should check that when you insert the first value the id is equal to "custom-id-001"  (and only that value exists in the db).
We can then insert a second value and make sure that the sequence has been incremented as expected (and only those two value exist in the db).

core/src/test/java/org/hibernate/ogm/backendtck/id/gen/CustomIdGeneratorTest.java
DavideD: By the way, if you expect only one value from the query, it makes sense to also use `.getSingleResult()` (it simpler to test the result this way)

fax4ever:
Thank you @DavideD I removed the UUID generator, so we can now predict the generated values.

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/1034

Thanks

PR #1035, opened 2018-06-01
Ogm 1485 Infispan Remote support more JoinTable association type use cases

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1485

### DIFF
--- a/core/src/test/java/org/hibernate/ogm/backendtck/associations/collection/manytomany/ManyToManyExtraTest.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/associations/collection/manytomany/ManyToManyExtraTest.java
@@ -10,57 +10,206 @@ import static org.fest.assertions.Assertions.assertThat;
 import static org.hibernate.ogm.utils.TestHelper.getNumberOfAssociations;
 import static org.hibernate.ogm.utils.TestHelper.getNumberOfEntities;
 
-import org.hibernate.Session;
-import org.hibernate.Transaction;
 import org.hibernate.ogm.utils.OgmTestCase;
+
+import org.junit.After;
 import org.junit.Test;
 
 /**
+ * Verify cardinality of entities and associations
+ * in unidirectional {@link javax.persistence.OneToMany}.
+ *
+ * Rule of thumb:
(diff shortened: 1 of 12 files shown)


DavideD:
A small note about commit messages.

I'm not a huge fan of the commit messages starting with the same sentence. There is already the Jira issue to highlight that the change is part of the same issue, I would prefer to keep the character in the first line to contains a more detailed description of the change. Also, sometimes the change is not directly related to the JIRA, but has been applied anyway because it wouldn't make too much sense to create a JIRA for it.

For example, the verb "to fix" is generic, it doesn't tell me anything about the problem that you have fixed. It's much more helpful if you can add some details about it instead of ` Infinispan R. JoinTable - ` that doesn't really tell me anything useful about the commit and take several chars.

Here's another example:
`OGM-1485 Infinispan R. JoinTable - improve base test methods`

vs

`OGM-1485 Rollback transaction in OgmTestCase#deleteAll in case of exception`

Which one do you find more helpful?
You can also add some additional details in the message description after the first line if necessary

fax4ever:
Thank you @DavideD. It's very helpful, maybe we should add it to our commit recommended practiced document. I'm going to rebase commit messages.

DavideD:
One day I will write one :)

For now I would recommend  the "Clean Code: .." book by Robert C. Martin

DavideD:
> Thank you @DavideD. It's very helpful, maybe we should add it to our commit recommended practiced document. I'm going to rebase commit messages.

Maybe you can start by collecting the one you found helpful so far and we can then add them to the web site.

fax4ever:
I've taken note of them. Commit messages rebased

fax4ever:
@DavideD I'm going to split it into:
* Remove skip-tests from unidirectional many-to-many and one-to-many
* Adding more unidirectional many-to-many tests

DavideD:
Thanks

fax4ever:
I've split last commit, pushing output.

DavideD:
Rebased and Merged https://github.com/hibernate/hibernate-ogm/pull/1035

I just did a few minor changes to the comments in the first two commits.

fax4ever:
Thank you!

PR #1037, opened 2018-06-05
OGM-1222 Add MongoDB object mutation test

DavideD:
Follows https://github.com/hibernate/hibernate-ogm/pull/1024

https://hibernate.atlassian.net/browse/OGM-1222

This changes the previous PR in the following way:
* Assert that the value has been updated
* Move the test in core so that we can check the other dialects as well
* Update commit message
* Remove the MongoDB runner (it should be in a separate issue)


fax4ever:
@DavideD it is a very dialect-independent and general purpose test. I have no objection to merge it.

DavideD:
@fax4ever Did you forget to close the PR and the JIRA?

fax4ever:
Merged thanks

PR #1038, opened 2018-06-07
OGM-1431 Server Tasks for Infinispan Remote

DavideD:
A follow up for https://github.com/hibernate/hibernate-ogm/pull/1014

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureOutputsImpl.java
+++ b/core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureOutputsImpl.java
@@ -22,7 +22,9 @@ import javax.persistence.ParameterMode;
 
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.metamodel.spi.MetamodelImplementor;
+import org.hibernate.ogm.dialect.impl.GridDialects;
 import org.hibernate.ogm.dialect.query.spi.ClosableIterator;
+import org.hibernate.ogm.dialect.spi.GridDialect;
 import org.hibernate.ogm.dialect.spi.TupleContext;
 import org.hibernate.ogm.dialect.storedprocedure.spi.StoredProcedureAwareGridDialect;
 import org.hibernate.ogm.model.key.spi.EntityKeyMetadata;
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureOutputsImpl.java
```diff
@@ -55,13 +57,19 @@ public NoSQLProcedureOutputsImpl(NoSQLProcedureCallImpl procedureCall) {
 		this.sessionFactory = procedureCall.getSession().getFactory();
 		this.gridDialect = this.sessionFactory.getServiceRegistry().getService( StoredProcedureAwareGridDialect.class );
 		if ( gridDialect == null ) {
-			throw new UnsupportedOperationException( String.format(
-					"Your grid dialect do not supports stored procedures. You tried to call '%s' procedure",
-					procedureCall.getProcedureName()
-			) );
+			String dialectType = dialectType( sessionFactory );
+			throw new UnsupportedOperationException( String.format( "Grid dialect %s does not support server side procedures: procedure '%s' has been called",
```
fax4ever: Why not use the *org.hibernate.ogm.util.impl.Log*?

infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/utils/StoredProceduresDeployer.java
```diff
@@ -32,7 +32,7 @@
 /**
  * @author The Viet Nguyen
  */
-public class InfinispanRemoteStoredProceduresTestHelper {
+public class StoredProceduresDeployer {
```
fax4ever: I like new name +1

infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/utils/StoredProceduresDeployer.java
```diff
@@ -0,0 +1,126 @@
...
+				.addAsServiceProvider( ServerTask.class, SimpleValueProcedure.class );
+
+		JavaArchive resultSetJar = ShrinkWrap.create( JavaArchive.class, "result-set-procedure.jar" )
+				.addClass( Car.class )
+				.addClass( ResultSetProcedure.class )
+				.addAsServiceProvider( ServerTask.class, ResultSetProcedure.class );
+
+		JavaArchive exceptionalProcedureJar = ShrinkWrap.create( JavaArchive.class, "exceptional-procedure.jar" )
+				.addClass( ExceptionalProcedure.class )
+				.addAsServiceProvider( ServerTask.class, ExceptionalProcedure.class );
```
fax4ever: very nice!

core/src/main/java/org/hibernate/ogm/storedprocedure/impl/NoSQLProcedureOutputsImpl.java
DavideD: That's a good question,  can you create a separate JIRA and merge the rest?

I was in doubt about this because when something is not supported we usually don't use Log, but this should likely be an HIbernateException.

fax4ever: Yes I do

gsmet: @DavideD I think what you did is consistent with the other places in the code. We do throw an `UnsupportedOperationException` in other places.

fax4ever: Thank you @gsmet! @DavideD: So I do not have to create any issue ok?

fax4ever:
Merged thank you!

PR #1039, opened 2018-06-08
OGM-1478 Make EM#getCriteriaBuilder throw proper exception

DavideD:
https://hibernate.atlassian.net/browse/OGM-1478

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/hibernatecore/impl/OgmSessionFactoryImpl.java
+++ b/core/src/main/java/org/hibernate/ogm/hibernatecore/impl/OgmSessionFactoryImpl.java
@@ -18,6 +18,7 @@ import javax.persistence.EntityManager;
 import javax.persistence.EntityManagerFactory;
 import javax.persistence.PersistenceException;
 import javax.persistence.SynchronizationType;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/hibernate/ogm/hibernatecore/impl/OgmSessionFactoryImpl.java
```diff
@@ -91,6 +92,11 @@ public OgmSession getCurrentSession() throws HibernateException {
 		return new OgmSessionImpl( this, (EventSource) session );
 	}
 
+	@Override
+	public CriteriaBuilder getCriteriaBuilder() {
+		throw new NotSupportedException( "OGM-23", "Criteria queries are not supported yet" );
```
fax4ever: I don't know if we should add a test for it

DavideD: We usually don't have tests for feature we don't support. I think it's OK to not add a new test for this.

DavideD:
I think there is an error, I will close this one for now and re-open it later

PR #1041, opened 2018-06-12
OGM-1495 Generate @SequenceGenerator source id values with cluster counter

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1495

### DIFF
--- a/infinispan-remote/infinispan-server-testconfig/wildfly-trimmed-config.xml
+++ b/infinispan-remote/infinispan-server-testconfig/wildfly-trimmed-config.xml
@@ -99,6 +99,7 @@
         <subsystem xmlns="urn:infinispan:server:core:9.2">
             <cache-container name="clustered" default-cache="default" statistics="true">
                 <transport lock-timeout="60000"/>
+                <global-state/>
                 <distributed-cache-configuration name="transactional">
                     <transaction mode="NON_XA" locking="PESSIMISTIC"/>
                 </distributed-cache-configuration>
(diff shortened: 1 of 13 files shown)


infinispan-remote/infinispan-server-testconfig/wildfly-trimmed-config.xml
```diff
@@ -99,6 +99,7 @@
         <subsystem xmlns="urn:infinispan:server:core:9.2">
             <cache-container name="clustered" default-cache="default" statistics="true">
                 <transport lock-timeout="60000"/>
+                <global-state/>
```
DavideD: Remember that it's nice to have in the commit also a "why" you are doing something ( not necessarly in the first line of the commit)
In this case, why do you need to add it back?

fax4ever: File *wildfly-trimmed-config.xml* is a subset of cluster.xml configuration. The global state is present in all Infinispan server provided configurations (standalone.xml, clustered.xml, domain profiles) and it is necessary to allow to the server to store persistent counters data.

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/1041

I've added this to one of the commit before merging it:

> File wildfly-trimmed-config.xml is a subset of cluster.xml configuration. The global state is present in all Infinispan server provided configurations (standalone.xml, clustered.xml, domain profiles) and it is necessary to allow to the server to store persistent counters data.

Thanks

fax4ever:
Thank you!

PR #1044, opened 2018-06-18
OGM-1499 Update integration tests to use WildFly 13.0.0.Final

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1499

Unfortunately 5.2 Hibernate Search slot version is already taken in WildFly 13.0.0.Final, provisioned with Hibernate Search 5.10.1.Final, requiring Hibernate ORM 5.3.1.Final.
But the actual version of Hibernate OGM requires ORM 5.3.0.Final.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -46,7 +46,7 @@
             See http://search.maven.org/#search|gav|1|g%3A"org.wildfly"%20AND%20a%3A"wildfly-parent"
         -->
 
-        <version.wildfly>12.0.0.Final</version.wildfly>
+        <version.wildfly>13.0.0.Final</version.wildfly>
         <version.org.hibernate.commons.annotations>5.0.3.Final</version.org.hibernate.commons.annotations>
         <version.org.jboss.logging.jboss-logging>3.3.2.Final</version.org.jboss.logging.jboss-logging>
         <version.org.jboss.logging.jboss-logging-tools>2.1.0.Final</version.org.jboss.logging.jboss-logging-tools>
--- a/documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
+++ b/documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
@@ -574,36 +574,15 @@ link:$$https://repository.jboss.org/nexus/index.html#nexus-search;gav~org.hibern
 
 WildFly will by default attempt to guess which Persistence Provider you need by having a look at the `provider` section of the `persistence.xml`.
 
-==== Enable support for JEE 8
+==== Enable support for EE 8
 
-Hibernate OGM {hibernate-ogm-version} requires **CDI 2.0** and **JPA 2.2**, that belong to **JEE8** specification.
-WildFly 12 has limited support for Java EE8.
+Hibernate OGM {hibernate-ogm-version} requires **CDI 2.0** and **JPA 2.2**, that belong to **EE 8** specification.
+WildFly 13 has support for JavaEE 8.
 
-To enable required CDI version we need to start the server with __ee8.preview.mode__ Java system property set to **true** :
+But in order to enable required CDI and JPA versions we need to start the server with __ee8.preview.mode__ Java system property set to **true** :
 
 ----
 -Dee8.preview.mode=true
-----
-
-To enable required JPA version we need to apply __hibernate-jpa-api-2.2-wildflymodules__ patch to the server.
-Download the patch from here:
-
- * http://central.maven.org/maven2/org/hibernate/javax/persistence/hibernate-jpa-api-2.2-wildflymodules/1.0.0.Beta2/hibernate-jpa-api-2.2-wildflymodules-1.0.0.Beta2-wildfly-12.0.0.Final-patch.zip
-
-Or using these maven coordinates:
-[source, XML]
-[subs="verbatim,attributes"]
-----
-<groupId>org.hibernate.javax.persistence</groupId>
-<artifactId>hibernate-jpa-api-2.2-wildflymodules</artifactId>
-<classifier>wildfly-{wildfly-version}-patch</classifier>
-<version>1.0.0.Beta2</version>
-<type>zip</type>
-----
-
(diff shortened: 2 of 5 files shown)


documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
```diff
@@ -577,35 +577,14 @@ WildFly will by default attempt to guess which Persistence Provider you need by
 ==== Enable support for JEE 8
 
 Hibernate OGM {hibernate-ogm-version} requires **CDI 2.0** and **JPA 2.2**, that belong to **JEE8** specification.
-WildFly 12 has limited support for Java EE8.
+WildFly 13 has support for Java EE8. But in order to enable a JEE8 profile, we need to set a system property.
```
gsmet: This is not an improvement. The support is still limited AFAIK, at least, WF 13 is not certified so I don't think we can say if supports EE 8.

By the way I think it's written "EE 8".

The second sentence is redundant with the sentence below so I don't think you should add it.

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
```diff
@@ -577,35 +577,14 @@ WildFly will by default attempt to guess which Persistence Provider you need by
 ==== Enable support for JEE 8
 
 Hibernate OGM {hibernate-ogm-version} requires **CDI 2.0** and **JPA 2.2**, that belong to **JEE8** specification.
```
gsmet: to the **EE 8** specification

I don't think JEE should be used nowadays.

pom.xml
```diff
@@ -70,16 +70,15 @@
...
-
         <!-- Module slots -->
 
         <module-slot.org.hibernate.ogm.short-id>${parsed-version.majorVersion}.${parsed-version.minorVersion}</module-slot.org.hibernate.ogm.short-id>
         <module-slot.org.hibernate.ogm.jipijapa.full-id>${project.version}</module-slot.org.hibernate.ogm.jipijapa.full-id>
-        <module-slot.org.hibernate.search.short-id>${parsed-version.org.hibernate.search.majorVersion}.${parsed-version.org.hibernate.search.minorVersion}</module-slot.org.hibernate.search.short-id>
+
+        <!-- Temporary use of full version *5.10.0.Final* as module slot for Hibernate Search -->
+        <!-- Because *5.1* slot is reserved to *5.10.1.Final* in Wildfly 13.0.0.Final -->
+        <module-slot.org.hibernate.search.short-id>${version.org.hibernate.search}</module-slot.org.hibernate.search.short-id>
```
gsmet: I think, you mean "5.10", not "5.1".

By the way, is there a reason why you can't simply use the 5.10.1 version from WF? Or did we decide to override it anyway?

Wildfly should be written WildFly (with a capital F)

bom/pom.xml
```diff
@@ -717,16 +717,6 @@
...
-                    <execution>
-                        <id>parse-search-version</id>
-                        <goals>
-                            <goal>parse-version</goal>
-                        </goals>
-                        <configuration>
-                            <propertyPrefix>parsed-version.org.hibernate.search</propertyPrefix>
-                            <versionString>${version.org.hibernate.search}</versionString>
-                        </configuration>
-                    </execution>
```
gsmet: Please don't remove this. Especially since you state that your change is temporary.

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
fax4ever: Reading the release page: http://wildfly.org/news/2018/05/30/WildFly13-Final-Released/ I'm thinking that maybe we can remove the **limited** world. But as you prefer, is not critical in my opinion. Ok s/EE8/EE 8. Moreover I agree the latter sentence is redundant.

pom.xml
fax4ever: Thank you, yes of course it was 5.10. Good question! Unfortunately we cant use 5.10.1 because the **slot name** is defined by the feature pack project: org.hibernate:hibernate-search-jbossmodules-orm and the wildfly-server-provisioning-maven-plugin can't change the slot name. There is a debate on capability to change or define the final module instead at provisioning time.

bom/pom.xml
fax4ever: I'm going to put it pack.

documentation/manual/src/main/asciidoc/modules/configuration.asciidoc
gsmet: Yeah OK about `limited`.

pom.xml
gsmet: If we decide that we want to override the WF version even if we use the same major.minor, I think we should simply move this to be a `full-id` and always use the full version.

fax4ever:
@gsmet thank you!

pom.xml
fax4ever: I would prefer to use use the *full-id* as a temporary solution. But as you prefer. Using short slot it is in general more easier to the user.

gsmet: Well, is the user supposed to do something with it? I don't know what exactly @Sanne had in mind with the feature packs.

For the user, the easier would be that we just use the version of WF but I'm not sure it's something sustainable (considering it will happen to be in line sometimes but mostly not). The issue here is that, when the versions are aligned, we end up doing some weird gymnastic which is not very desirable either.

@Sanne any idea on how we want to do things here?

Sanne: We decided to use slot names in the format of major.minor when it comes to "depending on another project" as SPIs should be stable enough, so to make it easy for people to drop-in a micro fix without having to change other project modules.

Dependencies across modules of the same project can use the full version. For example hibernate-search-orm would depend on hibernate-search-engine using the full version.

Incidentally, that's why we release aliased modules in feature packs as well.

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/1044

Thanks!

fax4ever:
Thanks!

PR #1046, opened 2018-06-20
OGM-1496 Constraint and Index on same key set and collection

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1496

### DIFF
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBIndexesDefiner.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBIndexesDefiner.java
@@ -1,0 +1,245 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.datastore.mongodb.impl;
+
+import java.lang.invoke.MethodHandles;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.hibernate.boot.model.naming.Identifier;
+import org.hibernate.boot.model.naming.NamingHelper;
+import org.hibernate.boot.model.relational.Database;
+import org.hibernate.boot.model.relational.Namespace;
+import org.hibernate.cfg.Environment;
+import org.hibernate.mapping.Column;
+import org.hibernate.mapping.Index;
+import org.hibernate.mapping.Table;
+import org.hibernate.mapping.UniqueKey;
+import org.hibernate.ogm.datastore.mongodb.index.impl.MongoDBIndexSpec;
+import org.hibernate.ogm.datastore.mongodb.index.impl.MongoDBIndexType;
+import org.hibernate.ogm.datastore.mongodb.logging.impl.Log;
+import org.hibernate.ogm.datastore.mongodb.logging.impl.LoggerFactory;
+import org.hibernate.ogm.datastore.spi.SchemaDefiner;
+import org.hibernate.ogm.options.shared.impl.IndexOptionsOption;
+import org.hibernate.ogm.options.shared.spi.IndexOption;
+import org.hibernate.ogm.options.shared.spi.IndexOptions;
+import org.hibernate.ogm.options.spi.OptionsService;
+import org.hibernate.ogm.util.impl.StringHelper;
+import org.hibernate.tool.hbm2ddl.UniqueConstraintSchemaUpdateStrategy;
+
+import com.mongodb.MongoException;
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.MongoCursor;
+import com.mongodb.client.MongoDatabase;
+import org.bson.Document;
+
+/**
+ * Performs sanity checks of defined indexes in the mapped objects.
+ */
+public class MongoDBIndexesDefiner {
+
+	private static final Log log = LoggerFactory.make( MethodHandles.lookup() );
+	private static final int INDEX_CREATION_ERROR_CODE = 85;
+
+	private List<MongoDBIndexSpec> indexSpecs = new ArrayList<>();
+
+	public List<MongoDBIndexSpec> getIndexSpecs() {
+		return indexSpecs;
+	}
+
+	void validateIndexSpecs(SchemaDefiner.SchemaDefinitionContext context) {
+		OptionsService optionsService = context.getSessionFactory().getServiceRegistry().getService( OptionsService.class );
+		Map<String, Class<?>> tableEntityTypeMapping = context.getTableEntityTypeMapping();
+
+		Database database = context.getDatabase();
+		UniqueConstraintSchemaUpdateStrategy constraintMethod = UniqueConstraintSchemaUpdateStrategy.interpret(
+				context.getSessionFactory().getProperties().get( Environment.UNIQUE_CONSTRAINT_SCHEMA_UPDATE_STRATEGY ) );
+		if ( constraintMethod == UniqueConstraintSchemaUpdateStrategy.SKIP ) {
+			log.tracef( "Skipping generation of unique constraints" );
+		}
+
+		for ( Namespace namespace : database.getNamespaces() ) {
+			for ( Table table : namespace.getTables() ) {
+				if ( table.isPhysicalTable() ) {
+					Class<?> entityType = tableEntityTypeMapping.get( table.getName() );
+					if ( entityType == null ) {
+						continue;
+					}
+
+					IndexOptions indexOptions = getIndexOptions( optionsService, entityType );
+					Set<String> forIndexNotReferenced = new HashSet<>( indexOptions.getReferencedIndexes() );
+
+					validateIndexSpecsForUniqueColumns( table, indexOptions, forIndexNotReferenced, constraintMethod );
+
+					validateIndexSpecsForUniqueKeys( table, indexOptions, forIndexNotReferenced, constraintMethod );
+
+					validateIndexSpecsForIndexes( table, indexOptions, forIndexNotReferenced );
+
+					for ( String forIndex : forIndexNotReferenced ) {
+						log.indexOptionReferencingNonExistingIndex( table.getName(), forIndex );
+					}
+				}
+			}
+		}
+	}
+
+	void createIndexes(MongoDatabase database) {
+		for ( MongoDBIndexSpec indexSpec : indexSpecs ) {
+			createIndex( database, indexSpec );
+		}
+	}
+
+	@SuppressWarnings("unchecked")
+	private void validateIndexSpecsForUniqueColumns(Table table, IndexOptions indexOptions, Set<String> forIndexNotReferenced,
+			UniqueConstraintSchemaUpdateStrategy constraintMethod) {
+		Iterator<Column> columnIterator = table.getColumnIterator();
+		while ( columnIterator.hasNext() ) {
+			Column column = columnIterator.next();
+			if ( column.isUnique() ) {
+				String indexName = NamingHelper.INSTANCE.generateHashedConstraintName(
+						"UK_",
+						table.getNameIdentifier(),
+						Identifier.toIdentifier( column.getName() ) );
+				forIndexNotReferenced.remove( indexName );
+
+				if ( constraintMethod != UniqueConstraintSchemaUpdateStrategy.SKIP ) {
+					MongoDBIndexSpec indexSpec = new MongoDBIndexSpec( table.getName(), column.getName(), indexName,
+							getIndexOptionDocument( table, indexOptions.getOptionForIndex( indexName ) ) );
+					if ( validateIndexSpec( indexSpec ) ) {
+						indexSpecs.add( indexSpec );
+					}
+				}
+			}
+		}
+	}
+
+	private void validateIndexSpecsForUniqueKeys(Table table, IndexOptions indexOptions, Set<String> forIndexNotReferenced,
+			UniqueConstraintSchemaUpdateStrategy constraintMethod) {
+		Iterator<UniqueKey> keys = table.getUniqueKeyIterator();
+		while ( keys.hasNext() ) {
+			UniqueKey uniqueKey = keys.next();
+			forIndexNotReferenced.remove( uniqueKey.getName() );
+
+			if ( constraintMethod != UniqueConstraintSchemaUpdateStrategy.SKIP ) {
+				MongoDBIndexSpec indexSpec = new MongoDBIndexSpec( uniqueKey,
+						getIndexOptionDocument( table, indexOptions.getOptionForIndex( uniqueKey.getName() ) ) );
+				if ( validateIndexSpec( indexSpec ) ) {
+					indexSpecs.add( indexSpec );
+				}
+			}
+		}
+	}
+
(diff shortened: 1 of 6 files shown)


mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBIndexesDefiner.java
```diff
@@ -0,0 +1,270 @@
...
+		// we rely on hashcode and equals to verify conflicts
+		if ( !indexSpecs.contains( indexSpec ) ) {
+			indexSpecs.add( indexSpec );
+			return;
+		}
+
+		int existingIndexPosition = indexSpecs.indexOf( indexSpec );
+		MongoDBIndexSpec existingIndex = indexSpecs.get( existingIndexPosition );
+
+		// in case on conflict
```
DavideD: s/on/of/

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBIndexesDefiner.java
```diff
@@ -0,0 +1,270 @@
...
+		catch (Exception e) {
+			throw log.invalidOptionsFormatForIndex( table.getName(), indexOption.getTargetIndexName(), e );
+		}
+	}
+
+	private boolean validateIndexSpec(MongoDBIndexSpec indexSpec) {
+		boolean valid = true;
+		if ( StringHelper.isNullOrEmptyString( indexSpec.getIndexName() ) ) {
+			log.indexNameIsEmpty( indexSpec.getCollection() );
+			valid = false;
```
DavideD: can't you just return false here and avod the use of variable?

DavideD: Never mind, the code was already like this

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBIndexesDefiner.java
```diff
@@ -147,11 +147,36 @@ void validateIndexSpecsForIndexes(Table table, IndexOptions indexOptions, Set<St
...
+		// we rely on compareTo to choose the best candidate
+		if ( indexSpec.compareTo( existingIndex ) < 0 ) {
+			// replace the existing definition with the new one
+			indexSpecs.remove( existingIndexPosition );
+			indexSpecs.add( indexSpec );
+			log.tryToDefineMultipleIndexesOnTheSameKeySet( indexSpec.getIndexKeysDocument().toJson(), existingIndex.getIndexName() );
+		}
+		else {
+			// discard the current definition
+			log.tryToDefineMultipleIndexesOnTheSameKeySet( indexSpec.getIndexKeysDocument().toJson(), indexSpec.getIndexName() );
```
DavideD: I don't understand what's going on here. How the compare to helps you to choose the right candidate?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/index/impl/MongoDBIndexSpec.java
```diff
@@ -193,4 +194,38 @@ public String toString() {
...
+		return Objects.hash( collection, indexKeys );
+	}
+
+	/**
+	 * this method could be used to choose between different configuration
+	 */
+	@Override
+	public int compareTo(MongoDBIndexSpec o) {
+		// choose the one with unique over a non unique
+		if ( this.getOptions().isUnique() && !o.getOptions().isUnique() ) {
```
DavideD: I don't think it's a good idea to use the compareTo in this scenario. I think it is much easier to understand if you create a method in the class where you need this kind of comparison.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBIndexesDefiner.java
DavideD: Also, why you log the same message in both cases?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBIndexesDefiner.java
```diff
@@ -147,11 +147,36 @@ void validateIndexSpecsForIndexes(Table table, IndexOptions indexOptions, Set<St
...
+		if ( !indexSpecs.contains( indexSpec ) ) {
+			indexSpecs.add( indexSpec );
+			return;
+		}
+
+		int existingIndexPosition = indexSpecs.indexOf( indexSpec );
+		MongoDBIndexSpec existingIndex = indexSpecs.get( existingIndexPosition );
+
+		// in case on conflict
+		// we rely on compareTo to choose the best candidate
```
DavideD: This comment doesn't add anything useful to the follow up code. Also. what's the "best" candidate?
I think you can remove and add a  function with a sensible name instead.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/logging/impl/Log.java
```diff
@@ -144,4 +144,8 @@
 
 	@Message(id = 1240, value = "Procedures returning muliple documents are not supported. Procedure '%1$s' returned %2$d results")
 	HibernateException multipleDocumentReturnedByStoredProcedure(String storedProcedureName, int size);
+
+	@LogMessage(level = WARN)
+	@Message(id = 1241, value = "Try to define another index on the same key set %s: definition of index %s will be skipped!")
```
DavideD: I would change this to something like:

`Skipping creation of Index <name of the index>.  Index <name of the index> already defined for properties %s.`

By the way, does this only apply for unique constraints?
In that case I would be more precise, like, `Skipping creation of index %s. Unique constraint on properties %s already defined by index %s`

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/MongoDBUniqueContraintIndexTest.java
```diff
@@ -0,0 +1,75 @@
...
+			session.persist( policyHolder );
+		} );
+
+		inTransaction( session -> {
+			PolicyHolder stored = session.load( PolicyHolder.class, policyHolder.name );
+			assertThat( policyHolder.email ).isEqualTo( stored.getEmail() );
+			assertThat( policyHolder.address ).isEqualTo( stored.getAddress() );
+		} );
+
+		deleteAll( PolicyHolder.class, policyHolder.name );
```
DavideD: Can you move the delete and clean up in the after method. This makes sure that in case of failures the values are deleted from the db making it less likely to cause failures of the following tests.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/MongoDBIndexTest.java
```diff
@@ -39,8 +42,9 @@ public void testSuccessfulIndexCreation() throws Exception {
 				indexMap.get( "name_idx" ).toJson() );
 		assertJsonEquals( "{ 'v' : 2 , 'unique' : true , 'key' : { 'author' : 1 , 'name' : 1} , 'name' : 'author_name_idx' , 'ns' : 'ogm_test_database.T_POEM' , 'sparse' : true}",
 				indexMap.get( "author_name_idx" ).toJson() );
-		assertJsonEquals( "{ 'v' : 2 , 'key' : { 'name' : 1 , 'author' : 1} , 'name' : 'IDXjo3qu8pkq9vsofgrq58pacxfq' , 'ns' : 'ogm_test_database.T_POEM' }",
-				indexMap.get( "IDXjo3qu8pkq9vsofgrq58pacxfq" ).toJson() );
+
+		// Index IDXjo3qu8pkq9vsofgrq58pacxfq will be never defined
```
DavideD: This comment is not very useful if don't explain why.
In this scenario, I prefer to use the `.as("")` so that you can explained what made the assertion fails.

DavideD: Keep in mind that you are changing the code. so you know what that index is about. But when I  look at the  source code only I only see `IDXjo3qu8pkq9vsofgrq58pacxfq`  and I don't know what index is supposed to define and we we don't expect to exists. Also, what if it failes because the name is changed?

I know that you didn't write this test originally, but I think it needs to be updated. For example, one assertion state that the size is 5 but it only checks 3 of the indexes. Do you think it is possible to improve it a bit? I genuinely asking because I don't know what the other values in the map are

DavideD:
Finished the review, I've added some comments.
I think there are a couple of things to change before we can merge it.
Let me know.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBIndexesDefiner.java
fax4ever: Yes unfortunately the move of the method from another class it is not visible.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBIndexesDefiner.java
fax4ever: Here I reach the else branch if have 2 indexes, both defined on the same field set of the same collection. The solution proposed is to skip the definition of one of them.

fax4ever: The role of *compareTo* here is to choose one of them. The choose should be repeatable with the same result, so we need a fixed criteria. BTW we can avoid to use *compareTo* to this.

Sanne: the `log` statement will need guarding. Please check if the cathegory of the log level is high enough to not perform a `toJson()` operation for no reason.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/index/impl/MongoDBIndexSpec.java
fax4ever: I agree

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/logging/impl/Log.java
fax4ever: We can't know in this phase if the index is derived from a unique constraint or not. In fact a unique constraint it is equivalent to a normal unique index defined on same fields :(.

fax4ever: Anyway I like the first message you proposed.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBIndexesDefiner.java
fax4ever: Severity level is warning here, the row is executed every time a user tries to define another index on the same fields of a pre-existing one.

fax4ever: Then the operation is performed rarely. I would prefer leave `toJSon`. But as you prefer, it is not critical to me.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBIndexesDefiner.java
fax4ever: What do you think of:
`public boolean hasPreferableConfgiruationOver(MongoDBIndexSpec other)`?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBIndexesDefiner.java
DavideD: I think the point is to add a check like `if ( log.isEnabled( Logger.Level.DEBUG.WARN) ) { ... }`.
You don't have to remove the call to toJSon().

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBIndexesDefiner.java
DavideD: I think the problem for me is that I don't understand how one configuration is preferable to the other.
If they both add an index why do you update it? Couldn't you simply keep the existing one?

fax4ever: Suppose we have a unique constraint an a normal index non unique, like the test case. In this case I prefer to keep the unique property on the index, because probably the user want the unique.

fax4ever: Then the configuration with unique it should take the precedence over the non unique one.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBIndexesDefiner.java
fax4ever: I don't like anymore the comment. I'm going to change it.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBIndexesDefiner.java
fax4ever: Does it make sense guarding a WARN? It's just a curiosity. Anyway I'm going to to that.

DavideD: I guess in production somebody might be only interested to see ERRORs

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBIndexesDefiner.java
DavideD: I would expect the code to look something like:
```
if (isNotUnique(existingContraint)) {
    replaceWithUniqueConstratint(existingConstraint, newConstraint);
}
else {
   // keep existing (maybe)
}
```
But I'm not sure we should log a warning every time, like, if a user define a text index and a spatial index on the same field, shouldn't we throw an exception?
What do you think?

fax4ever: The idea was to define a method to compare and choose between two configurations, but you're right maybe it is overdesign. In fact the only criteria that we have at the moment is only the uniqueness.  I'm going to simplify it.

fax4ever: I have just remembered that there is another problem: the order of arrival of the configuration items is not guaranteed. Choosing different configurations over different creation can create a problem of consistency of the behaviour. For this reason I would like to keep the method to choose between configurations.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/MongoDBUniqueContraintIndexTest.java
fax4ever: Good idea. It is my miss!

fax4ever:
I'm going to close this PR, I'll reopen a new one.

PR #1048, opened 2018-06-22
OGM-1496 Handle indexes and constraints on same columns 

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1496

The first commit is only to add new in line test helper.

### DIFF
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBSchemaDefiner.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBSchemaDefiner.java
@@ -154,10 +154,10 @@ public class MongoDBSchemaDefiner extends BaseSchemaDefiner {
 					IndexOptions indexOptions = getIndexOptions( optionsService, entityType );
 					Set<String> forIndexNotReferenced = new HashSet<>( indexOptions.getReferencedIndexes() );
 
+					// Don't change the order of these validate methods:
+					// UniqueKeys are used to validate normal Indexes
 					validateIndexSpecsForUniqueColumns( table, indexOptions, forIndexNotReferenced, constraintMethod );
-
 					validateIndexSpecsForUniqueKeys( table, indexOptions, forIndexNotReferenced, constraintMethod );
-
 					validateIndexSpecsForIndexes( table, indexOptions, forIndexNotReferenced );
 
 					for ( String forIndex : forIndexNotReferenced ) {
@@ -216,10 +216,24 @@ public class MongoDBSchemaDefiner extends BaseSchemaDefiner {
 			forIndexNotReferenced.remove( index.getName() );
 			MongoDBIndexSpec indexSpec = new MongoDBIndexSpec( index,
 					getIndexOptionDocument( table, indexOptions.getOptionForIndex( index.getName() ) ) );
-			if ( validateIndexSpec( indexSpec ) ) {
+			if ( validateIndexSpec( indexSpec ) && thereIsNoUniqueConstraintOnSameColumns( indexSpec ) ) {
 				indexSpecs.add( indexSpec );
 			}
 		}
+	}
+
+	private boolean thereIsNoUniqueConstraintOnSameColumns(MongoDBIndexSpec indexSpec) {
+		for ( MongoDBIndexSpec otherIndex : indexSpecs ) {
+			boolean hasSameColumns = otherIndex.getIndexKeysDocument().equals( indexSpec.getIndexKeysDocument() );
+			boolean isUniqueConstraint = otherIndex.getOptions().isUnique();
+
+			// Skip index creation if a unique constraint is already present
+			if ( hasSameColumns && isUniqueConstraint ) {
+				return false;
+			}
+		}
+
+		return true;
 	}
 
 	private Document getIndexOptionDocument(Table table, IndexOption indexOption) {
--- a/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/MongoDBIndexOptionsConflictsTest.java
+++ b/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/MongoDBIndexOptionsConflictsTest.java
@@ -1,0 +1,285 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.datastore.mongodb.test.index;
+
+import static org.fest.assertions.Assertions.assertThat;
+import static org.hibernate.ogm.utils.TestHelper.inTransaction;
+import static org.junit.Assert.fail;
+
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Index;
+import javax.persistence.Table;
+import javax.persistence.UniqueConstraint;
+
+import org.hibernate.HibernateException;
+import org.hibernate.SessionFactory;
+import org.hibernate.ogm.options.shared.IndexOption;
+import org.hibernate.ogm.options.shared.IndexOptions;
+import org.hibernate.ogm.sessionfactory.SessionFactoryBuilder;
+import org.hibernate.ogm.utils.TestForIssue;
+
+import org.junit.After;
+import org.junit.Test;
+
+import com.mongodb.MongoCommandException;
+
+/**
+ * Tests what happens when an index and a constraint or different indexes are defined on the same column set.
+ *
+ * Some configurations are legal, just because MongoDB could tolerate them, others are not.
+ * It depends on the presence of index option conflicts.
+ * An index option conflict for MongoDB is when two indexes are defined on the same columns with different option configurations.
+ *
+ * Moreover a unique constraint is equivalent to a unique index. Then unique constraint can generate index option conflicts too.
+ * Even different types generate option conflicts.
+ *
+ * But we decide to support also the cases:
+ * <li>Having a unique constraint and not unique index defined on the same columns</li>
+ * <li>Having a unique index and not unique index defined on the same columns</li>
+ *
+ * the expected behaviour here is that all work, simply skipping the overlapped *not unique* index.
+ *
+ * @author Fabio Massimo Ercoli
+ */
+@TestForIssue(jiraKey = "OGM-1496")
+public class MongoDBIndexOptionsConflictsTest {
+
+	private SessionFactoryBuilder config;
+	private Object entity;
+
+	@After
+	public void cleanUp() {
+		if ( entity == null ) {
+			return;
+		}
+
+		try ( SessionFactory sessionFactory = config.build() ) {
+			inTransaction( sessionFactory, session -> {
+				session.delete( entity );
+			} );
+		}
+	}
+
+	@Test
+	public void testUniqueConstraintAndNOTUniqueIndex() {
+		config = SessionFactoryBuilder.entities( UniqueConstraintAndNotUniqueIndexEntity.class );
+
+		UniqueConstraintAndNotUniqueIndexEntity entity = new UniqueConstraintAndNotUniqueIndexEntity();
+		entity.name = "Fabio Massimo";
+		entity.email = "fabiomassimo@myemailprovider.eu";
+		entity.address = "11bis Rue Roquepine Paris";
+		this.entity = entity;
+
+		try ( SessionFactory sessionFactory = config.build() ) {
+			inTransaction( sessionFactory, session -> {
+				session.persist( entity );
+			} );
+
+			inTransaction( sessionFactory, session -> {
+				UniqueConstraintAndNotUniqueIndexEntity stored = session.load( UniqueConstraintAndNotUniqueIndexEntity.class, entity.name );
+				assertThat( entity.email ).isEqualTo( stored.getEmail() );
+				assertThat( entity.address ).isEqualTo( stored.getAddress() );
+			} );
+		}
+	}
+
+	@Test
+	public void testUniqueConstraintAndUniqueIndex() {
+		config = SessionFactoryBuilder.entities( UniqueConstraintAndUniqueIndexEntity.class );
+
+		UniqueConstraintAndUniqueIndexEntity entity = new UniqueConstraintAndUniqueIndexEntity();
+		entity.name = "Fabio Massimo";
+		entity.email = "fabiomassimo@myemailprovider.eu";
+		entity.address = "11bis Rue Roquepine Paris";
+		this.entity = entity;
+
+		try ( SessionFactory sessionFactory = config.build() ) {
+			inTransaction( sessionFactory, session -> {
+				session.persist( entity );
+			} );
+
+			inTransaction( sessionFactory, session -> {
+				UniqueConstraintAndUniqueIndexEntity stored = session.load( UniqueConstraintAndUniqueIndexEntity.class, entity.name );
+				assertThat( entity.email ).isEqualTo( stored.getEmail() );
+				assertThat( entity.address ).isEqualTo( stored.getAddress() );
+			} );
+		}
+	}
+
+	@Test
+	public void testUniqueIndexAndNOTUniqueIndex() {
+		config = SessionFactoryBuilder.entities( UniqueIndexAndNotUniqueIndexEntity.class );
+
+		UniqueIndexAndNotUniqueIndexEntity entity = new UniqueIndexAndNotUniqueIndexEntity();
+		entity.name = "Fabio Massimo";
+		entity.email = "fabiomassimo@myemailprovider.eu";
+		entity.address = "11bis Rue Roquepine Paris";
+		this.entity = entity;
+
+		try ( SessionFactory sessionFactory = config.build() ) {
+			inTransaction( sessionFactory, session -> {
+				session.persist( entity );
+			} );
+
+			inTransaction( sessionFactory, session -> {
+				UniqueIndexAndNotUniqueIndexEntity stored = session.load( UniqueIndexAndNotUniqueIndexEntity.class, entity.name );
+				assertThat( entity.email ).isEqualTo( stored.getEmail() );
+				assertThat( entity.address ).isEqualTo( stored.getAddress() );
+			} );
+		}
+	}
+
+	@Test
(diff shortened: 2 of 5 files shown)


core/src/test/java/org/hibernate/ogm/sessionfactory/SessionFactoryConfig.java
```diff
@@ -0,0 +1,71 @@
...
+import org.hibernate.ogm.cfg.OgmConfiguration;
+import org.hibernate.ogm.utils.TestHelper;
+
+/**
+ * Builder class to create a test configuration
+ * to use within {@link SessionFactoryInlineTest} inline helper.
+ *
+ * @author Fabio Massimo Ercoli
+ */
+public class SessionFactoryConfig {
```
gsmet: I think I would have called it `SessionFactoryBuilder`.

core/src/test/java/org/hibernate/ogm/sessionfactory/SessionFactoryInlineTest.java
```diff
@@ -0,0 +1,68 @@
...
+	 * @param builder given test configuration
+	 * @param consumer function to apply to the created session factory
+	 */
+	public static void withInSessionFactory(SessionFactoryConfig builder, Consumer<SessionFactory> consumer) {
+		try ( SessionFactory sessionFactory = builder.getConfiguration().buildSessionFactory( builder.getServiceRegistry() ) ) {
+			consumer.accept( sessionFactory );
+		}
+	}
+
+	public static void inTransaction(SessionFactory sessionFactory, Consumer<Session> consumer) {
```
gsmet: Doesn't ORM already have this sort of helper? I recall @DavideD using them.

core/src/test/java/org/hibernate/ogm/sessionfactory/SessionFactoryConfig.java
```diff
@@ -0,0 +1,71 @@
...
+
+	public Configuration getConfiguration() {
+		return configuration;
+	}
+
+	public StandardServiceRegistry getServiceRegistry() {
+		return serviceRegistry;
+	}
+
+	public SessionFactoryConfig build() {
```
gsmet: And the `build()` method would simply have returned a `SessionFactory`.

You would avoid having 2 getters and a state that is not very clear.

core/src/test/java/org/hibernate/ogm/sessionfactory/SessionFactoryInlineTest.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+public class SessionFactoryInlineTest {
+
+	/**
+	 * Expect {@link Exception} during the build of {@link SessionFactory}
+	 *
+	 * @param builder given test configuration
+	 * @param consumer function to apply to the expected exception
+	 */
+	public static void insideCatchException(SessionFactoryConfig builder, Consumer<Exception> consumer) {
```
gsmet: Maybe `assertInitializationFailure()`?

core/src/test/java/org/hibernate/ogm/sessionfactory/SessionFactoryInlineTest.java
```diff
@@ -0,0 +1,68 @@
...
+		}
+	}
+
+	/**
+	 * Expect creation of {@link SessionFactory}
+	 *
+	 * @param builder given test configuration
+	 * @param consumer function to apply to the created session factory
+	 */
+	public static void withInSessionFactory(SessionFactoryConfig builder, Consumer<SessionFactory> consumer) {
```
gsmet: `within`

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/MongoDBIndexOptionsConflictsTest.java
```diff
@@ -0,0 +1,246 @@
...
+import org.hibernate.ogm.utils.TestForIssue;
+
+import org.junit.Test;
+
+import com.mongodb.MongoCommandException;
+
+/**
+ * Tests what happens when an index and a constraint or different indexes are defined on the same column set.
+ *
+ * Some configurations are legal, because MongoDB could tolerate them, other not.
```
gsmet: `other not` -> `others are not`

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/MongoDBIndexOptionsConflictsTest.java
```diff
@@ -0,0 +1,246 @@
...
+ * It depends on the presence of index option conflicts.
+ * An index option conflict for MongoDB is when two indexes are defined on the same columns with different option configurations.
+ *
+ * Moreover a unique constraint is equivalent to a unique index. Then unique constraint can generate index option conflicts too.
+ * Even different types generate option conflicts.
```
gsmet: Is this sentence necessary considering the one above? If it is, it's unclear.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/MongoDBIndexOptionsConflictsTest.java
```diff
@@ -0,0 +1,246 @@
...
+ *
+ * @author Fabio Massimo Ercoli
+ */
+@TestForIssue(jiraKey = "OGM-1496")
+public class MongoDBIndexOptionsConflictsTest {
+
+	@Test
+	public void testUniqueConstraintAndNOTUniqueIndex() {
+		SessionFactoryConfig config = SessionFactoryConfig.entities( UniqueConstraintAndNotUniqueIndexEntity.class ).build();
```
gsmet: With my proposal, you wouldn't call the `build()` method here and keep the builder.

core/src/test/java/org/hibernate/ogm/sessionfactory/SessionFactoryInlineTest.java
fax4ever: Good question. Hibernate ORM already has, but in its private test module. We already use it a non static within `OgmTestCase` and `OgmJpaTestCase`. In the future we will have all this classes in a dedicated test module, see the issue: https://hibernate.atlassian.net/browse/OGM-1487.

core/src/test/java/org/hibernate/ogm/sessionfactory/SessionFactoryConfig.java
fax4ever: Good question again. But the creation of SesssionFactory is a critical moment and it should be performed by the `SessionFactoryInlineTest`. In particular an exception could be cached or not, it depends by the test expectations.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/MongoDBIndexOptionsConflictsTest.java
fax4ever: The latter refers to conflicts generated using different types, whereas the former refers to conflicts generated using different unique attribute value.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/MongoDBIndexOptionsConflictsTest.java
fax4ever: In general we could have many entities and many options, I would prefer to perform build action just once, if I understood well what you mean here.

core/src/test/java/org/hibernate/ogm/sessionfactory/SessionFactoryConfig.java
gsmet: Sure, you can just call the `build()` method at the right time in the test.

The fact is that you don't really need the config. And your building pattern for the config is not very satisfactory.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/MongoDBIndexOptionsConflictsTest.java
gsmet: I don't see how what I propose would mean performing the build action several times?

core/src/test/java/org/hibernate/ogm/sessionfactory/SessionFactoryConfig.java
gsmet: Well, with my proposal, it would be a `SessionFactoryBuilder` because you would build the `SessionFactory` with the `build()` method. Currently, you `build()` method builds 2 objects that are then accessible via getters, which could have been called earlier but you would have null results. I don't think it's a good pattern.

`SessionFactoryTestBuilder` would be an acceptable name if we go with my pattern. I agree if you keep yours, your name is OK.

Try what I suggest with building the `SessionFactory` in the `build()` method. I think you'll see it simplifies things and make everything cleaner. (Unless I'm missing something which is possible too :))

core/src/test/java/org/hibernate/ogm/sessionfactory/SessionFactoryConfig.java
fax4ever: Sorry @gsmet I got it now what you mean. It is definitely a good idea.

core/src/test/java/org/hibernate/ogm/sessionfactory/SessionFactoryConfig.java
fax4ever: You're right

core/src/test/java/org/hibernate/ogm/sessionfactory/SessionFactoryInlineTest.java
```diff
@@ -0,0 +1,68 @@
...
+		}
+	}
+
+	/**
+	 * Expect creation of {@link SessionFactory}
+	 *
+	 * @param builder given test configuration
+	 * @param consumer function to apply to the created session factory
+	 */
+	public static void withinSessionFactory(SessionFactoryBuilder builder, Consumer<SessionFactory> consumer) {
```
DavideD: Can we remove this method? I don't find it very useful.
I find 
```
try ( SessionFactory sessionFactory = builder.build() ) {
```

much more expressive and easier to understand than

```
withinSessionFactory( config, sessionFactory -> {
```

where I don't know exactly what the method is going to do.
It also doesn't save me much in terms of lines.

core/src/test/java/org/hibernate/ogm/sessionfactory/SessionFactoryInlineTest.java
```diff
@@ -0,0 +1,68 @@
...
+	 * @param consumer function to apply to the expected exception
+	 */
+	public static void assertInitializationFailure(SessionFactoryBuilder builder, Consumer<Exception> consumer) {
+		try ( SessionFactory sessionFactory = builder.build() ) {
+			fail( "Expected exception at Hibernate factory creation time was not raised!" );
+		}
+		catch (Exception rootException) {
+			consumer.accept( rootException );
+		}
+	}
```
DavideD: This method doesn't seem all that useful either, why not call builder.build() in the test and then test the exception using the rule form JUnit?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBSchemaDefiner.java
```diff
@@ -156,6 +156,8 @@ private void validateIndexSpecs(SchemaDefinitionContext context) {
 
 					validateIndexSpecsForUniqueColumns( table, indexOptions, forIndexNotReferenced, constraintMethod );
 
+					// keep this order
```
DavideD: What order? The 3 validate methods? Better to add the order at the beginning then instead than in the middle

core/src/test/java/org/hibernate/ogm/sessionfactory/SessionFactoryInlineTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.hibernate.Transaction;
+
+/**
+ * Helps to test {@link SessionFactory} creation
+ * providing an inline test API
+ *
+ * @author Fabio Massimo Ercoli
+ */
+public class SessionFactoryInlineTest {
+
```
DavideD: I don't like the name of this class, it's misleading because it's not a Test.
Also, I don't find the methods in it all that useful.

The problem for me is that some methods simply replace existing approaches without adding much, meaning that becomes something more to learn for new users. Feel free to challenge this

What I would do is to rename this class to something like `JPATestHelper` or `HibernateTestHelper` and keep only the `inTransaction` method . We could also move the existing method in `OgmJpaTestCase` and `OgmTestCase` here so that we have everything in one place and we can use it even when the test class doesn't extends those classes.

Let me know what you think

DavideD:
The SessionFactoryBuilder looks handy, I'm not convinced about some of the other utility methods. I would prefer not to add them unless they add real benefits in writing the tests. Feel free to challenge this

fax4ever:
@DavideD about the utility of `SessionFactoryBuilder` just think about how many tests we have that are scoped on what happened on session factory creation.

DavideD:
> @DavideD about the utility of SessionFactoryBuilder just think about how many tests we have that are scoped on what happened on session factory creation.

See my previous example with the `try - catch `, how is the utility method better?

fax4ever:
For instance let's imagine `MongoDBIndexOptionsConflictsTest` without the helper.

DavideD:
Sure, let's imagine it:
```
	@Test
	public void testNormalIndexAndFullTextSearchIndex() {
		SessionFactoryBuilder config = SessionFactoryBuilder.entities( NormalIndexAndFullTextSearchIndexEntity.class );

		try ( SessionFactory facotry = config.build() ) {
			fail( "A failure should have occured because ..." );
		}
		catch (Exception e) {
			assertThat( e )
					.isExactlyInstanceOf( HibernateException.class )
					.hasMessage( "OGM001231: Unable to create index normal on collection MyDoc" );

			assertThat( e.getCause() ).isExactlyInstanceOf( MongoCommandException.class );
			assertThat( e.getCause().getMessage() ).contains( "already exists with different options" );
		}
	}
```

What's so bad about it?

I assume you are referring to this case, because in the other ones nothing really change.
But this is a special case where we want to dig a bit in the exception thrown.
Most of the time we are just interested t the first exception and `thrown.expectMessage` is more than enough. So you wouldn't even need the try cactch for those scenarios

fax4ever:
@DavideD I like your example. I think you're right. So let's remove the helper, definitely. In contrast we should keep the builder. Does it right?

DavideD:
I think extracting the `inTransaction` would be a nice touch if you have the time to do it. I don't think it is going to be too hard. Note that you can jsut import the class in OgmJpaTestCase and OgmTestCase so that you don't have to override all the places where inTransaction is used for now

fax4ever:
Thank you @DavideD for the review. I'm going to apply the suggested changes here.

fax4ever:
Now tests looks very pretty. Thank you @DavideD and @gsmet for suggestions here. I made the changes and rebased the branch.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBSchemaDefiner.java
```diff
@@ -216,12 +218,27 @@ private void validateIndexSpecsForIndexes(Table table, IndexOptions indexOptions
...
 	}
 
+	private boolean thereIsNoUniqueConstraintOnSameColumns(MongoDBIndexSpec indexSpec) {
+		for ( MongoDBIndexSpec otherIndex : indexSpecs ) {
+			boolean hasSameColumns = otherIndex.getIndexKeysDocument().equals( indexSpec.getIndexKeysDocument() );
+			boolean isUniqueConstraint = otherIndex.getOptions().isUnique();
+
+			// In case of presence of a unique constraint on the same columns
+			// we skip the index creation
+			if ( hasSameColumns && isUniqueConstraint ) {
```
DavideD: Change the comment to:
```
Skip index creation if a unique constraint is already present

File: mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/MongoDBIndexOptionsConflictsTest.java [hunk 14]
```diff
@@ -0,0 +1,279 @@
...
+			} );
+		}
+	}
+
+	@Test
+	public void testNormalIndexAndFullTextSearchIndex() {
+		SessionFactoryBuilder config = SessionFactoryBuilder.entities( NormalIndexAndFullTextSearchIndexEntity.class );
+
+		try ( SessionFactory ignored = config.build() ) {
+			fail( "Expected exception at Hibernate factory creation time was not raised!" );
```
DavideD: Let's change the message to:
```
Should fail  because we defined two index with conflicting configurations on field `UniqueConstraintAndNotUniqueIndexEntity#email`
```
Or whatever the reason you expect a failure

DavideD:
By the way, no  need to send a new PR, just merge it once you have applied those changes

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/MongoDBIndexOptionsConflictsTest.java
gsmet: s/index/indexes/

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/index/MongoDBIndexOptionsConflictsTest.java
```diff
@@ -0,0 +1,279 @@
...
+
+			inTransaction( sessionFactory, session -> {
+				TwoUniqueIndexesEntity stored = session.load( TwoUniqueIndexesEntity.class, entity.name );
+				assertThat( entity.email ).isEqualTo( stored.getEmail() );
+				assertThat( entity.address ).isEqualTo( stored.getAddress() );
+			} );
+
+			inTransaction( sessionFactory, session -> {
+				session.delete( entity );
+			} );
```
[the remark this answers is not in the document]
fax4ever: Sure I can

DavideD:
@fax4ever Is anything missing in this PR to be included?

fax4ever:
No there isn't @DavideD. I think that we can merge it. I'm going to do that.

DavideD:
Thanks Fabio

fax4ever:
Merged #1048

PR #1054, opened 2018-06-28
OGM-1506 Introduce a Maven profile to test on JDK10

Sanne:
https://hibernate.atlassian.net/browse/OGM-1506


fax4ever:
Thank you! Merged #1054

PR #1055, opened 2018-07-01
OGM-1511 Remove thread local serializzation context

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1511

### DIFF
--- a/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
+++ b/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
@@ -80,7 +80,6 @@ import org.infinispan.commons.util.CloseableIterator;
 import org.infinispan.query.dsl.FilterConditionContext;
 import org.infinispan.query.dsl.Query;
 import org.infinispan.query.dsl.QueryBuilder;
-import org.infinispan.query.dsl.QueryFactory;
 
 /**
  * Some implementation notes for evolution:
@@ -358,9 +357,9 @@ public class InfinispanRemoteDialect<EK, AK, ISK> extends AbstractGroupingByEnti
 		final String cacheName = cacheName( key );
 		ProtostreamAssociationMappingAdapter mapper = provider.getCollectionsDataMapper( cacheName );
 		return mapper.withinCacheEncodingContext( c -> {
-			QueryFactory queryFactory = Search.getQueryFactory( c );
+			QueryBuilder qb = Search.getQueryFactory( c ).from( provider.getEntityType( c ) );
+
 			final String[] columnNames = key.getColumnNames();
-			QueryBuilder qb = queryFactory.from( ProtostreamPayload.class );
 			FilterConditionContext bqEnd = null;
 			boolean firstIteration = true;
 			for ( int i = 0; i < columnNames.length; i++ ) {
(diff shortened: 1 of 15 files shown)


fax4ever:
The issue is a just a preview... I shall do the refactoring of old (not needed anymore) code!

anistor:
Lots of ```import org.infinispan.protostream.impl``` imports here. That will surely give you a lot of work when migrating to protostream 4.3 which starts to make such stuff package local.

anistor:
The situation on master is now like this: 
SchemaValidator imports org.infinispan.protostream.impl.parser.SquareProtoParser;
and ProtostreamSerializerSetup imports org.infinispan.protostream.impl.SerializationContextImpl.

I'll make a quick PR for master to fix those 2.

fax4ever:
Hi @anistor. Thank you for support! Yes we know about it. Fortunately all access to internal packages are confined within `org.hibernate.ogm.datastore.infinispanremote.impl.protostream.multimessage.MultiMessageExtension`. So we expect that we will have to change this class when we will update the library version.

anistor:
Yes, but is seems to me you are treating protostream like a Lego set and re-assemble the (private) pieces  :) Unfortunately 4.3 removes some of that internal stuff because it was refactored quite heavily. So even if confined to a single place it's still not too nice as it will most likely require full reimplementation on OMG's side.  I did not make a full analysis of your changes to see what you need to be exposed cleanly. Seems I'll have to release a 4.3.Alpha1 before you can proceed with this PR.

Or if it's really urgent for you, go on, but be warned it will require rework.

fax4ever:
I agree with you, in general I never use an internal API. This one would be a very very temporary solution. :( Whenever you want we can talk about the feature that Hibernate OGM would require from ProtoStream, maybe also with @Sanne. WDYT?

anistor:
https://github.com/hibernate/hibernate-ogm/pull/1056 solves the impl references from master :)

DavideD:
@fax4ever there are some conflicts in this PR, can you rebase it to the latest changes, please?

fax4ever:
Thanks @DavideD I'm going to rebase it.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/InfinispanRemoteDialect.java
```diff
@@ -358,9 +357,10 @@ public Association getAssociation(AssociationKey key, AssociationContext associa
 		final String cacheName = cacheName( key );
 		ProtostreamAssociationMappingAdapter mapper = provider.getCollectionsDataMapper( cacheName );
 		return mapper.withinCacheEncodingContext( c -> {
-			QueryFactory queryFactory = Search.getQueryFactory( c );
+			QueryBuilder qb = Search.getQueryFactory( c )
+					.from( provider.getProtobufPackageName() + "." + c.getName() );
```
DavideD: This `provider.getProtobufPackageName() + "." + c.getName()` should probably be in a utility method because you are using it also in the TestHelper

fax4ever: Davide, good idea. I've just added a commit with this refactor-step.

fax4ever:
In my opinion we can merge it. We have just to take in consideration the fact the with the new version of ProtoBuf library we **could and should** remove the internal access.

fax4ever:
BTW I'm going to create also an proper issue for it.

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/1055

THanks

fax4ever:
Thank you @DavideD!

PR #1056, opened 2018-07-02
OGM-1510 Remove some references to protostream internals

anistor:
https://hibernate.atlassian.net/browse/OGM-1510

### DIFF
--- a/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protobuf/schema/SchemaValidator.java
+++ b/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protobuf/schema/SchemaValidator.java
@@ -9,7 +9,6 @@ package org.hibernate.ogm.datastore.infinispanremote.impl.protobuf.schema;
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.net.URL;
-import java.util.Map;
 
 import org.hibernate.HibernateException;
 import org.hibernate.ogm.datastore.infinispanremote.impl.schema.SequenceTableDefinition;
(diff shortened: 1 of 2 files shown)


infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protobuf/schema/SchemaValidator.java
```diff
@@ -92,12 +92,11 @@ private void validate() {
 		}
 	}
 
-	private FileDescriptor parseSchema() {
+	private FileDescriptor parseSchema() throws DescriptorParserException, IOException {
 		FileDescriptorSource fileDescriptorSource = FileDescriptorSource.fromString( schemaName, protoSchema );
 		Configuration config = Configuration.builder().build();
-
-		SquareProtoParser protoParser = new SquareProtoParser( config );
-		Map<String, FileDescriptor> fileDescriptorMap = protoParser.parse( fileDescriptorSource );
```
anistor: Please note (from parse method javadoc) that semantic checking (it type name resolution) was not previously performed here, but the new code does it. That might lead to strange failures when validating each schema individually as they should be validated together so type resolution can be peformed properly.
I did not look deeply in the how OGM uses schemas to be able to tell if that's going to be an issue.
The test work, but that might not be the whole truth :)

anistor:
Updated commit message to include jira id

fax4ever:
@anistor thank you again! Yes here we need just a way to obtain the FileDescriptor. It's ok to get it in other way! Nice to see that we don't depend anymore on the internals of ProtoStream. I'm going to merge it after the CI feedback.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protobuf/schema/SchemaValidator.java
fax4ever: The semantic checking is performed by Hibernate OGM, we check that FileDescriptor instance contains all the required caches, messages and fields. Anyway I'm going to add a test on other PR to check the behavior in case of semantic error of provided schema itself.

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/1056

Thanks

PR #1058, opened 2018-07-03
OGM-1420 Test ElementCollection is loaded from JPQL query

DavideD:
https://hibernate.atlassian.net/browse/OGM-1420
 

A couple of tests that I created some time ago.

### DIFF
--- a/core/src/test/java/org/hibernate/ogm/backendtck/embeddable/ElementCollectionFromJPQLQueryTest.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/embeddable/ElementCollectionFromJPQLQueryTest.java
@@ -1,0 +1,79 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.backendtck.embeddable;
+
+import static org.fest.assertions.Assertions.assertThat;
+
+import java.util.List;
+
+import org.hibernate.ogm.utils.OgmTestCase;
+import org.hibernate.ogm.utils.TestForIssue;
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/hibernate/ogm/backendtck/embeddable/ElementCollectionFromJPQLQueryTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.util.List;
+
+import org.hibernate.ogm.utils.OgmTestCase;
+import org.hibernate.ogm.utils.TestForIssue;
+import org.junit.After;
+import org.junit.Test;
+
+/**
+ * Test that the collection of elements is loaded with the entity when it's the result of a JPQL query using JPA.
```
fax4ever: What you think of: `Tests that {@link ElementCollection} is loaded within the entity when it's the result of a JPQL query using Hibernate.`

core/src/test/java/org/hibernate/ogm/backendtck/embeddable/ElementCollectionFromJPQLQueryWithJPATest.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.util.List;
+
+import org.hibernate.ogm.utils.TestForIssue;
+import org.hibernate.ogm.utils.jpa.OgmJpaTestCase;
+import org.junit.After;
+import org.junit.Test;
+
+/**
+ * Test that the collection of elements is loaded with the entity when it's the result of a JPQL query using Hibernat.
```
fax4ever: Even here, what do you think of `Tests that {@link ElementCollection} is loaded within the entity when it's the result of a JPQL query using JPA.'

fax4ever:
Thank you @DavideD. Nice to have a test for it! I've just a very very minimal suggestion on test Javadoc.

DavideD:
We can probably merge the two superclasses together

fax4ever:
Merged #1058 thanks!

PR #1060, opened 2018-07-05
OGM-1501 Add support for ReadConcern

Krerg:


mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -862,8 +881,11 @@ public void forEachTuple(ModelConsumer consumer, TupleTypeContext tupleTypeConte
 		}
 
 		String collectionName = getCollectionName( backendQuery, queryDescriptor, entityKeyMetadata );
+		//WriteConcern writeConcern = getWriteConcern( tupleContext );
+		//ReadPreference readPreference = getReadPreference( tupleContext );
 		MongoCollection<Document> collection = provider.getDatabase().getCollection( collectionName );
-
```
DavideD: How come you area not applying the withOption method in this case? You are not applying anymore the write or read options, I'm surprised that no test fails in because of this

Krerg: Applying options when executing backend query is OGM-1448 bug. In tests only getObject, getObjects and getAssociation are called

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -415,6 +418,21 @@ private static Object prepareIdObjectValue(String[] columnNames, Object[] column
 		return getCollection( key.getTable() );
 	}
 
+	private MongoCollection<Document> withOptions(MongoCollection<Document> collection,  ReadPreference readPreference, ReadConcern readConcern) {
+		if ( readPreference != null && readConcern != null ) {
```
DavideD: I think this method should also apply the writeConcern and should be used everytime with need a collection.
This way the options are applied always by the same function.

I also find this code harder to read than needed, what do you think about refactoring it to look something like:

```
	private MongoCollection<Document> withOptions(MongoCollection<Document> collection, AssociationContext context) {
		return withOptions( collection, context.getAssociationTypeContext().getOptionsContext() );
	}

	private MongoCollection<Document> withOptions(MongoCollection<Document> collection, TupleTypeContext context) {
		return withOptions( collection, context.getOptionsContext() );
	}

	private MongoCollection<Document> withOptions(MongoCollection<Document> collection, OptionsContext context) {
		MongoCollection<Document> newCollection = collection;
		newCollection = withReadPreference( context, newCollection );
		newCollection = withReadConcern( context, newCollection );
		return newCollection;
	}

	private MongoCollection<Document> withReadConcern(OptionsContext context, MongoCollection<Document> newCollection) {
		ReadConcern readConcern = context.getUnique( ReadConcernOption.class );
		if ( readConcern != null ) {
			newCollection = newCollection.withReadConcern( readConcern );
		}
		return newCollection;
	}

	private MongoCollection<Document> withReadPreference(OptionsContext context, MongoCollection<Document> newCollection) {
		ReadPreference readPreference = context.getUnique( ReadPreferenceOption.class );
		if ( readPreference != null ) {
			return newCollection.withReadPreference( readPreference );
		}
		return newCollection;
	}
```

Now that think about it it might make sense to call withOptions everytime we call the getCollection( ),
something like:

```
	private MongoCollection<Document> getCollection(String table, ...) {
		return withOptions( currentDB.getCollection( table ), ... );
	}
```

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
DavideD: Mmmh... OK

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
DavideD: > for example the classes and the tests could be in separate commits.

What I meant is the *new* classes and the new tests could be in their own commits

Krerg:
@DavideD, Changes are ready to review

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1004,6 +1015,15 @@ public DuplicateInsertPreventionStrategy getDuplicateInsertPreventionStrategy(En
 		return new MongoDBAggregationOutput( output, entityKeyMetadata );
 	}
 
+	private static final OptionsContext getTypeContext(TupleContext tupleContext) {
+		try {
+			return tupleContext.getTupleTypeContext().getOptionsContext();
+		}
+		catch (HibernateException exception) {
+			return null;
```
DavideD: What kind of exception occurs here that you can ignore?

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -482,17 +512,17 @@ public boolean updateTupleWithOptimisticLock(EntityKey entityKey, Tuple oldLockS
...
 	}
 
 	/*
 	 * InsertOne the tuple and return an object containing the id in the field ID_FIELDNAME
 	 */
-	private Document insertDocument(EntityKeyMetadata entityKeyMetadata, Tuple tuple, WriteConcern writeConcern) {
+	private Document insertDocument(EntityKeyMetadata entityKeyMetadata, Tuple tuple, WriteConcern writeConcern, TupleContext tupleContext ) {
 		Document dbObject = objectForInsert( tuple, ( (MongoDBTupleSnapshot) tuple.getSnapshot() ).getDbObject() );
-		getCollection( entityKeyMetadata ).withWriteConcern( writeConcern ).insertOne( dbObject );
+		getCollection( entityKeyMetadata.getTable() , tupleContext.getTupleTypeContext().getOptionsContext() ).withWriteConcern( writeConcern ).insertOne( dbObject );
```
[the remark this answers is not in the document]
DavideD: By the way, there is a space after `getTable`: should be `.getTable(), ` and not `.getTable() , `

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
Krerg: There could be OnlyWithTransactionContext which throws exception with message "The tuple context is not available, probably because we are dealing with more than a single entity type".

DavideD:
Merged with some changes https://github.com/hibernate/hibernate-ogm/pull/1060

PR #1061, opened 2018-07-13
OGM-1494 Infinispan HotRod transaction support

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1494
This is a preview of supporting HotRod transactions,
Many tests didn't work yet.

The major issues are: 
* Thread local context (solved by another PR)
* size and query apis are not transactional :(

### DIFF
--- a/core/src/test/java/org/hibernate/ogm/backendtck/queries/SimpleQueriesTest.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/queries/SimpleQueriesTest.java
@@ -531,6 +531,7 @@ public class SimpleQueriesTest extends OgmTestCase {
 
 	@Test
 	@TestForIssue(jiraKey = "OGM-424")
+	@SkipByGridDialect(INFINISPAN_REMOTE)
 	public void testAutoFlushIsAppliedDuringQueryExecution() throws Exception {
 		Query<Hypothesis> query = session.createQuery( "from Hypothesis" );
 		assertQuery( session, 8, query );
@@ -553,6 +554,7 @@ public class SimpleQueriesTest extends OgmTestCase {
 
 	@Test
 	@TestForIssue(jiraKey = "OGM-424")
+	@SkipByGridDialect(INFINISPAN_REMOTE)
 	public void testEntitiesInsertedInCurrentSessionAreFoundByQueriesNotBasedOnHibernateSearch() throws Exception {
 		Query<Hypothesis> query = session.createQuery( "from Hypothesis h where h.position = 30" );
 		assertQuery( session, 0, query );
@@ -575,6 +577,7 @@ public class SimpleQueriesTest extends OgmTestCase {
 
 	@Test
 	@TestForIssue(jiraKey = "OGM-424")
+	@SkipByGridDialect(INFINISPAN_REMOTE)
 	public void testSetFlushModeIsApplied() throws Exception {
 		Query<Hypothesis> query = session.createQuery( "from Hypothesis h where h.position = 31" );
 		assertQuery( session, 0, query );
--- a/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/InfinispanRemoteDatastoreProvider.java
+++ b/infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/InfinispanRemoteDatastoreProvider.java
@@ -14,6 +14,7 @@ import java.util.Map;
 import java.util.Set;
 
 import org.hibernate.boot.model.relational.Sequence;
+import org.hibernate.engine.transaction.jta.platform.spi.JtaPlatform;
 import org.hibernate.ogm.datastore.infinispanremote.InfinispanRemoteDialect;
 import org.hibernate.ogm.datastore.infinispanremote.configuration.impl.InfinispanRemoteConfiguration;
 import org.hibernate.ogm.datastore.infinispanremote.impl.cachehandler.HotRodCacheCreationHandler;
@@ -66,6 +67,8 @@ public class InfinispanRemoteDatastoreProvider extends BaseDatastoreProvider
 	 * by our Hot Rod client.
 	 */
 	private final OgmProtoStreamMarshaller marshaller = new OgmProtoStreamMarshaller();
+
+	private JtaPlatform jtaPlatform;
 
 	// Only available during configuration
 	private InfinispanRemoteConfiguration config;
@@ -125,7 +128,10 @@ public class InfinispanRemoteDatastoreProvider extends BaseDatastoreProvider
 
 	@Override
 	public void start() {
-		hotrodClient = HotRodClientBuilder.builder().withConfiguration( config, marshaller ).build();
+		hotrodClient = HotRodClientBuilder.builder()
+				.withConfiguration( config, marshaller )
+				.withTransactionMode( transactionMode, jtaPlatform ).build();
+
 		// TODO temporary create another cache manager to handle remote script registration and invocation due to incompatibility of ProtoStreamMarshaller.
 		// When https://issues.jboss.org/browse/ISPN-8020 is closed, we could remove it and reuse the common hotrodClient.
 		scriptManager = HotRodClientBuilder.builder().withConfiguration( config, new GenericJBossMarshaller() ).build();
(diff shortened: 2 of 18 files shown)


infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/utils/InfinispanRemoteTestHelper.java
```diff
@@ -68,6 +69,30 @@ public long getNumberOfAssociations(SessionFactory sessionFactory) {
...
+		for ( CollectionPersister ep : persisters ) {
+			OgmCollectionPersister persister = (OgmCollectionPersister) ep;
+			String tableName = persister.getTableName();
+			SingleTableOgmEntityPersister owningPersister = (SingleTableOgmEntityPersister) persister.getOwnerEntityPersister().getEntityPersister();
+			String ownerTableName = owningPersister.getTableName();
+
+			RemoteCache<Object, Object> cache = datastoreProvider.getCache( tableName );
+			if ( !( cache instanceof TransactionalRemoteCacheImpl ) ) {
+				totalCount += countAssociations( tableName, ownerTableName, datastoreProvider );
+				continue;
```
DavideD: A small note, could you use a if-then-else instead of the `continue`, please? I find the code easier to follow without the `continue`.

infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/utils/InfinispanRemoteTestHelper.java
```diff
@@ -68,6 +69,30 @@ public long getNumberOfAssociations(SessionFactory sessionFactory) {
 		return totalCount;
 	}
 
+	@Override
+	public long getNumberOfAssociations(Session session) {
+		final InfinispanRemoteDatastoreProvider datastoreProvider = getProvider( session.getSessionFactory() );
```
DavideD: shouldn't this method be simply `return getNumberOfAssociation( session.getSessionFactory );`?

infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/utils/InfinispanRemoteTestHelper.java
```diff
@@ -175,6 +200,28 @@ public long getNumberOfEntities(SessionFactory sessionFactory) {
...
+			OgmEntityPersister persister = (OgmEntityPersister) ep;
+			String tableName = persister.getTableName();
+			RemoteCache<Object, Object> cache = datastoreProvider.getCache( tableName );
+			if ( !( cache instanceof TransactionalRemoteCacheImpl ) ) {
+				counter.addAndGet( cache.size() );
+				continue;
+			}
+
+			int size = TransactionalRemoteCacheCounter.count( persister, (TransactionalRemoteCacheImpl) cache, datastoreProvider.getDataMapperForCache( tableName ), session );
+			counter.addAndGet( size );
```
DavideD: maybe a bit of explanation here might be helpful for the future. Like, why the `if`?

DavideD: Wouldn't be better to have a separate method to get the size and then add the value to the counter?

infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/utils/TransactionalRemoteCacheCounter.java
```diff
@@ -0,0 +1,133 @@
...
+		cache.keySet().forEach( committedKey -> {
+			Boolean stillPresentInLocalTransaction = mapper.withinCacheEncodingContext( c -> c.containsKey( committedKey ) );
+			if ( stillPresentInLocalTransaction ) {
+				// skip count of removed entries
+				committedKeys.add( committedKey );
+			}
+		} );
+
+		// count all committed keys not removed by the local transaction
+		Integer counter = new Integer( committedKeys.size() );
```
DavideD: Do you need a new Integer? If that's the case `Integer.valueOf` is better

core/src/test/java/org/hibernate/ogm/backendtck/queries/SimpleQueriesTest.java
```diff
@@ -553,6 +554,7 @@ public void testAutoFlushIsAppliedDuringQueryExecution() throws Exception {
 
 	@Test
 	@TestForIssue(jiraKey = "OGM-424")
+	@SkipByGridDialect(INFINISPAN_REMOTE)
```
DavideD: How come these tests are failing and the other aren't?

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/InfinispanRemoteDatastoreProvider.java
```diff
@@ -252,7 +257,7 @@ public HotRodSequenceHandler getSequenceHandler() {
 	}
 
 	public <K, V> RemoteCache<K, V> getCache(String cacheName) {
-		RemoteCache<K,V> cache = hotrodClient.getCache( cacheName );
+		RemoteCache<K,V> cache = hotrodClient.getCache( cacheName, TransactionMode.NONE );
```
DavideD: Should this become a property?

DavideD:
The approach seems alright. Do you need me to check something specific?

infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/utils/InfinispanRemoteTestHelper.java
fax4ever: Sure

infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/utils/InfinispanRemoteTestHelper.java
fax4ever: The main idea here is to handle the fact that size and query APIs are not transactional. This is a big issue for all test that try to count within transactions. The workaround that I used here is use the pesistent context data of a specific session to *enrich* the result with local transaction transient data. The algoritmh is very complex, for this reason I'm not sure to merge it. Unfortunately this one is the only way that I found to count entities and associations. :(

infinispan-remote/src/test/java/org/hibernate/ogm/datastore/infinispanremote/utils/TransactionalRemoteCacheCounter.java
fax4ever: Sure! I've repeated this mistake. Sorry :P

fax4ever: Even simpler, i'm going to use a just an int.

core/src/test/java/org/hibernate/ogm/backendtck/queries/SimpleQueriesTest.java
fax4ever: Because the others verify results on a different transaction.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/InfinispanRemoteDatastoreProvider.java
fax4ever: The param TransactionMode.NONE should be removed at all. I didn't because some tests not pass yet without it.

fax4ever:
Thank you @DavideD. I gathered your advice. You can find it in the last commit. By the way, the issue is still under working. I'm fixing right now a strange behavior on cascade checks. Furthermore, other tests do not work properly yet. I still have many doubts about  the fact that this integration would be possible at the moment.

fax4ever:
The PR is still a preview, but now just few tests fails. This solution is dependent by [OGM-1512 issue](https://hibernate.atlassian.net/browse/OGM-1512), thus is rebased on [pull/1063](https://github.com/hibernate/hibernate-ogm/pull/1063).

fax4ever:
The issue is no more a preview, because now all tests pass. Moreover, we handle the problem of manage non transactional caches, adding the property `hibernate.ogm.cache.transaction.mode`. I'm going to update our docs in a subsequent commit.

DavideD:
@fax4ever can you have a look at this? It seems it's not building anymore after a rebase

fax4ever:
Sure @DavideD. I'll do

PR #1063, opened 2018-07-22
OGM-1512 Flush operation after a cascade delete could prejudice commit

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1512


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/1063

Thanks

PR #1068, opened 2018-08-13
OGM-1517 Add test persistence.xml without provider

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1517

### DIFF
--- a/core/src/test/java/org/hibernate/ogm/backendtck/jpa/JPAStartWithoutProvider.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/jpa/JPAStartWithoutProvider.java
@@ -1,0 +1,81 @@
+/*
+ * Hibernate OGM, Domain model persistence for NoSQL datastores
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.ogm.backendtck.jpa;
+
+import static org.fest.assertions.Assertions.assertThat;
+
+import java.util.HashMap;
+import java.util.Map;
+import javax.persistence.EntityManager;
+import javax.persistence.EntityManagerFactory;
+import javax.persistence.EntityTransaction;
+import javax.persistence.Persistence;
+
+import org.hibernate.ogm.utils.PackagingRule;
+import org.hibernate.ogm.utils.TestForIssue;
+import org.hibernate.ogm.utils.TestHelper;
+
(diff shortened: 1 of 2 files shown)


fax4ever:
I've put `Preview` label, because we need to check better the test here.

core/src/test/java/org/hibernate/ogm/backendtck/jpa/JPAStartWithoutProvider.java
```diff
@@ -0,0 +1,81 @@
...
+import org.junit.rules.ExpectedException;
+
+/**
+ * Tests a store and load scenario, using the JPA interface,
+ * without defining a persistence provider to enable Hibernate OGM engine.
+ *
+ * @author Fabio Massimo Ercoli
+ */
+@TestForIssue(jiraKey = "OGM-1517")
+public class JPAStartWithoutProvider {
```
DavideD: Class name should end with `-Test`

core/src/test/java/org/hibernate/ogm/backendtck/jpa/JPAStartWithoutProvider.java
```diff
@@ -0,0 +1,81 @@
...
+
+	@Rule
+	public PackagingRule packaging = new PackagingRule( "persistencexml/ogm-noprovider.xml", Poem.class );
+
+	@Rule
+	public ExpectedException thrown = ExpectedException.none();
+
+	@Test
+	public void testJPAStartWithoutProvider() {
+		Map<String, Object> settings = new HashMap<>( TestHelper.getDefaultTestSettings() );
```
DavideD: Do you need this? Can't you simply pass the `Map<String, String>`?
```
factory = Persistence.createEntityManagerFactory( "ogm", TestHelper.getDefaultTestSettings() );
```

core/src/test/java/org/hibernate/ogm/backendtck/jpa/JPAStartWithoutProvider.java
```diff
@@ -0,0 +1,81 @@
...
+			finally {
+				if ( manager != null ) {
+					manager.close();
+				}
+			}
+		}
+		finally {
+			if ( factory != null ) {
+				factory.close();
+			}
```
DavideD: I think that for this test I would prefer to close everything in a `@After` method. This way the test becomes much smaller and easier to follow. This is just a suggestion.

fax4ever:
@DavideD thank you for the hints here. I'm going to close this PR.
In order to apply the strategy that we decided in the last meeting, described in the last comment of the issue itself.

PR #1070, opened 2018-08-16
Enable tests for Infinispan Remote that require tx

DavideD:
https://hibernate.atlassian.net/browse/OGM-1520


core/src/test/java/org/hibernate/ogm/backendtck/jpa/JPAJTATest.java
```diff
@@ -31,8 +31,8 @@
 
 	@Test
 	@SkipByGridDialect(
-			value = { GridDialectType.MONGODB, GridDialectType.INFINISPAN_REMOTE },
```
fax4ever: I think there is miss space here before the `}`

DavideD:
Yes, please, if you can fix it and merge it that would be great!

fax4ever:
Merged #1070 thank you!

PR #1072, opened 2018-08-28
Fixed bad looking code example

Anastasko:
Hi guys, just removed redundant double quote which was preventing next lines to be colored in right way.
Here is a screen from my PC.
screenshot from 2018-08-28 13-33-18


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/1072

I've just added the issue code to the commit.

DavideD:
Thanks a lot @Anastasko!

PR #1075, opened 2018-08-29
OGM-1518 Test empty hotrodclient.properties

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1518

I've just added a test case.
It seems that everything works even with an empty `hotrodclient.properties`-

Does the issue make sense? Or shall we solve it as `Out Of Date`?
Additional question: Does it make sense to override  the Infinispan defaults?

### DIFF
--- a/documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
+++ b/documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
@@ -728,6 +728,24 @@ minIdle = 1
 ----
 ====
 
+The `hotrodclient.properties` is optional, in this case Hibernate OGM will use the following:
+
+|===
+| HotRod client property | Hibernate OGM mandatory value
+| infinispan.client.hotrod.marshaller
+| org.hibernate.ogm.datastore.infinispanremote.impl.protostream.OgmProtoStreamMarshaller
+| infinispan.client.hotrod.force_return_values
+| true
+|===
+
+Every other property will use the default values defined by http://infinispan.org/docs/stable/user_guide/user_guide.html#hotrod:java-client[Infinispan - Java Hot Rod client].
+
+[IMPORTANT]
+====
+These values for `infinispan.client.hotrod.marshaller` and `infinispan.client.hotrod.force_return_values` are mandatory in order to keep the expected behaviour of the dialect.
+If you change them, an exception will be thrown.
+====
+
 Alternatively you can embed the Hot Rod properties in your Hibernate (or JPA) configuration
 file, but you'll have to replace the `infinispan.client.hotrod.` prefix with the custom
 prefix `hibernate.ogm.infinispan_remote.client.`.
--- a/infinispan-remote/src/test/resources/hotrodclient-minimal.properties
+++ b/infinispan-remote/src/test/resources/hotrodclient-minimal.properties
@@ -1,0 +1,7 @@
+# empty HotRod configuration file
+# we use default for all properties here
+# it should work if there is an Infinispan server listening on host:port 127.0.0.1:11222
+
+# infinispan.client.hotrod.server_list = 127.0.0.1:11222
+# infinispan.client.hotrod.marshaller = org.hibernate.ogm.datastore.infinispanremote.impl.protostream.OgmProtoStreamMarshaller
+# infinispan.client.hotrod.async_executor_factory = org.infinispan.client.hotrod.impl.async.DefaultAsyncExecutorFactory
--- a/core/src/main/java/org/hibernate/ogm/util/configurationreader/impl/SimplePropertyReaderContext.java
+++ b/core/src/main/java/org/hibernate/ogm/util/configurationreader/impl/SimplePropertyReaderContext.java
@@ -164,17 +164,18 @@ public class SimplePropertyReaderContext<T> extends PropertyReaderContextBase<T>
 
 	private URL getAsUrl() {
 		Object value = getConfiguredValue();
-
 		if ( StringHelper.isNullOrEmptyString( value ) && getDefaultValue() != null ) {
 			return (URL) getDefaultValue();
 		}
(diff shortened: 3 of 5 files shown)


DavideD:
> Additional question: Does it make sense to override the Infinispan defaults?

I don't know, does it? :)

I mean, what's the default in Infinispan? Unless it causes problems with OGM  I wouldn't do anything (we can still add the test).

I think the important aspect is to update the documentation mentioning which defaults are applied

DavideD:
> Does the issue make sense? Or shall we solve it as Out Of Date?

I think we can change the issue to "Make sure that OGM works with HotRod default values"

DavideD:
> I think the important aspect is to update the documentation mentioning which defaults are applied

Maybe we can just add that the HotRod defaults are applied without entering into too much details.
@fax4ever Do you know which defaults get used? Could have a look at it.

Everything else seems alright, I'll wait for you to have quick look at that and then we can merge it

documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
```diff
@@ -728,6 +728,28 @@ minIdle = 1
 ----
 ====
 
+You can pass even an empty `hotrodclient.properties`.
+In this case the default HotRod properties of Infinispan will be applyied.
```
DavideD: /applyied/applied/

documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
```diff
@@ -728,6 +728,28 @@ minIdle = 1
...
+See http://infinispan.org/docs/stable/user_guide/user_guide.html#hotrod:java-client[Infinispan - Java Hot Rod client].
+
+The only exceptions are these properties:
```
DavideD: /The only exceptions are/Except for/

documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
```diff
@@ -728,6 +728,28 @@ minIdle = 1
...
+|===
+
+That Hibernate OGM will provide for you if you don't declare them, overriding the Infinispan default.
+
+[IMPORTANT]
+====
+These values for these properties are mandatory in order to keep the expected behaviour of the dialect.
+Then if you try to change these, an exception will be thrown.
+====
+
```
DavideD: It's not bad but can we use a different structure?
Something like:

The  `hotrodclient.properties` is optional, in this case Hibernate OGM will use the following:
...
Every other property will use the default values defined by Infinispan [link]

Let me know what you think about it. If you don't agree, feel free to merge it as it is (with the syntax errors fixed, please).

documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
```diff
@@ -728,6 +728,28 @@ minIdle = 1
...
+| infinispan.client.hotrod.force_return_values
+| true
+|===
+
+That Hibernate OGM will provide for you if you don't declare them, overriding the Infinispan default.
+
+[IMPORTANT]
+====
+These values for these properties are mandatory in order to keep the expected behaviour of the dialect.
+Then if you try to change these, an exception will be thrown.
```
DavideD: /Then if you try to change these/If you change them/

documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
fax4ever: I will apply the change you suggested.

documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
```diff
@@ -728,6 +728,24 @@ minIdle = 1
...
+| org.hibernate.ogm.datastore.infinispanremote.impl.protostream.OgmProtoStreamMarshaller
+| infinispan.client.hotrod.force_return_values
+| true
+|===
+
+Every other property will use the default values defined by http://infinispan.org/docs/stable/user_guide/user_guide.html#hotrod:java-client[Infinispan - Java Hot Rod client].
+
+[IMPORTANT]
+====
+These values for these properties are mandatory in order to keep the expected behaviour of the dialect.
```
Sanne: This warning could be clearer if you could state explicitly that you are referring to the above two properties, or maybe even repeat them one last time?

infinispan-remote/src/test/resources/hotrodclient-minimal.properties
```diff
@@ -0,0 +1,7 @@
+# empty HotRod configuration file
```
Sanne: Very nice!

Could we also test that everything works when there is no hotrod properties file at all ? (if we don't already do?)

fax4ever: Currently it wouldn't work without a hotrod properties file. Are we sure to allow that? I think that at least a --host:port-- combination is required, in a real, non localhost, cases. Isn't it?

documentation/manual/src/main/asciidoc/modules/infinispan.asciidoc
fax4ever: Good idea. I will change it in that way

infinispan-remote/src/test/resources/hotrodclient-minimal.properties
Sanne: But itsn't it silly to require to have an empty file?

I would expect that when the user does not specify a configuration resource name / filename, we will apply the defaults.

fax4ever: It is, I agree. I will change it.

Sanne:
merged! thanks @fax4ever

fax4ever:
Thanks @Sanne!

PR #1076, opened 2018-08-30
OGM-1349 Support stored procedures for Neo4j

grin7728:
https://hibernate.atlassian.net/browse/OGM-1349

### DIFF
--- a/core/src/test/java/org/hibernate/ogm/backendtck/storedprocedures/PositionalParametersStoredProcedureCallTest.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/storedprocedures/PositionalParametersStoredProcedureCallTest.java
@@ -67,156 +67,190 @@ public class PositionalParametersStoredProcedureCallTest extends OgmJpaTestCase 
 
 	@Test
 	public void testSingleResultDynamicCall() throws Exception {
-		StoredProcedureQuery storedProcedureQuery = em.createStoredProcedureQuery( SIMPLE_VALUE_PROC );
-		storedProcedureQuery.registerStoredProcedureParameter( 0, Integer.class, ParameterMode.IN );
-		storedProcedureQuery.setParameter( 0, 1 );
+		inTransaction( entityManager -> {
+			StoredProcedureQuery storedProcedureQuery = entityManager.createStoredProcedureQuery( SIMPLE_VALUE_PROC );
+			storedProcedureQuery.registerStoredProcedureParameter( 0, Integer.class, ParameterMode.IN );
+			storedProcedureQuery.setParameter( 0, 1 );
 
-		Number singleResult = (Number) storedProcedureQuery.getSingleResult();
-		assertThat( singleResult.intValue() ).isEqualTo( 1 );
+			Number singleResult = (Number) storedProcedureQuery.getSingleResult();
+			assertThat( singleResult.intValue() ).isEqualTo( 1 );
+		} );
 	}
 
 	@Test
 	public void testResultSetDynamicCallWithResultClass() throws Exception {
-		StoredProcedureQuery storedProcedureQuery = em.createStoredProcedureQuery( RESULT_SET_PROC, Car.class );
-		storedProcedureQuery.registerStoredProcedureParameter( 0, Void.class, ParameterMode.REF_CURSOR );
-		storedProcedureQuery.registerStoredProcedureParameter( 1, Integer.class, ParameterMode.IN );
-		storedProcedureQuery.registerStoredProcedureParameter( 2, String.class, ParameterMode.IN );
-		storedProcedureQuery.setParameter( 1, 1 );
-		storedProcedureQuery.setParameter( 2, "title" );
+		inTransaction( entityManager -> {
+			StoredProcedureQuery storedProcedureQuery = entityManager.createStoredProcedureQuery( RESULT_SET_PROC, Car.class );
+			storedProcedureQuery.registerStoredProcedureParameter( 0, Void.class, ParameterMode.REF_CURSOR );
+			storedProcedureQuery.registerStoredProcedureParameter( 1, Integer.class, ParameterMode.IN );
+			storedProcedureQuery.registerStoredProcedureParameter( 2, String.class, ParameterMode.IN );
+			storedProcedureQuery.setParameter( 1, 1 );
+			storedProcedureQuery.setParameter( 2, "title" );
 
-		@SuppressWarnings("unchecked")
-		List<Car> listResult = storedProcedureQuery.getResultList();
-		assertThat( listResult ).containsOnly( new Car( 1, "title" ) );
+			@SuppressWarnings("unchecked")
+			List<Car> listResult = storedProcedureQuery.getResultList();
+			assertThat( listResult ).containsOnly( new Car( 1, "title" ) );
+		} );
 	}
 
 	@Test
 	public void testResultSetDynamicCallWithResultMapping() throws Exception {
-		StoredProcedureQuery storedProcedureQuery = em.createStoredProcedureQuery( RESULT_SET_PROC, "carMapping" );
-		storedProcedureQuery.registerStoredProcedureParameter( 0, Void.class, ParameterMode.REF_CURSOR );
-		storedProcedureQuery.registerStoredProcedureParameter( 1, Integer.class, ParameterMode.IN );
-		storedProcedureQuery.registerStoredProcedureParameter( 2, String.class, ParameterMode.IN );
-		storedProcedureQuery.setParameter( 1, 1 );
-		Parameter<String> p2 = storedProcedureQuery.getParameter( 2,String.class );
-		storedProcedureQuery.setParameter( p2, "title'1" );
+		inTransaction( entityManager -> {
+			StoredProcedureQuery storedProcedureQuery = entityManager.createStoredProcedureQuery( RESULT_SET_PROC, "carMapping" );
+			storedProcedureQuery.registerStoredProcedureParameter( 0, Void.class, ParameterMode.REF_CURSOR );
+			storedProcedureQuery.registerStoredProcedureParameter( 1, Integer.class, ParameterMode.IN );
+			storedProcedureQuery.registerStoredProcedureParameter( 2, String.class, ParameterMode.IN );
+			storedProcedureQuery.setParameter( 1, 1 );
+			Parameter<String> p2 = storedProcedureQuery.getParameter( 2, String.class );
+			storedProcedureQuery.setParameter( p2, "title'1" );
 
-		@SuppressWarnings("unchecked")
-		List<Car> listResult = storedProcedureQuery.getResultList();
-		assertThat( listResult ).containsOnly( new Car( 1, "title'1" ) );
+			@SuppressWarnings("unchecked")
+			List<Car> listResult = storedProcedureQuery.getResultList();
+			assertThat( listResult ).containsOnly( new Car( 1, "title'1" ) );
+		} );
 	}
 
 	@Test
 	public void testResultSetStaticCallWithResultClass() throws Exception {
-		StoredProcedureQuery storedProcedureQuery = em.createNamedStoredProcedureQuery( "returnPositionalParametersWithEntity" );
-		// First parameter is void
-		storedProcedureQuery.setParameter( 2, 1 );
-		storedProcedureQuery.setParameter( 3, "title" );
+		inTransaction( entityManager -> {
+			StoredProcedureQuery storedProcedureQuery = entityManager.createNamedStoredProcedureQuery(
+					"returnPositionalParametersWithEntity" );
+			// First parameter is void
+			storedProcedureQuery.setParameter( 2, 1 );
+			storedProcedureQuery.setParameter( 3, "title" );
 
-		@SuppressWarnings("unchecked")
-		List<Car> listResult = storedProcedureQuery.getResultList();
-		assertThat( listResult ).containsOnly( new Car( 1, "title" ) );
+			@SuppressWarnings("unchecked")
+			List<Car> listResult = storedProcedureQuery.getResultList();
+			assertThat( listResult ).containsOnly( new Car( 1, "title" ) );
+		} );
 	}
 
 	@Test
 	public void testResultSetStaticCallWithResultMapping() throws Exception {
-		StoredProcedureQuery storedProcedureQuery = em.createNamedStoredProcedureQuery( "returnPositionalParametersWithMapping" );
-		// First parameter is void
-		storedProcedureQuery.setParameter( 2, 2 );
-		storedProcedureQuery.setParameter( 3, "title'2" );
+		inTransaction( entityManager -> {
+			StoredProcedureQuery storedProcedureQuery = entityManager.createNamedStoredProcedureQuery(
+					"returnPositionalParametersWithMapping" );
+			// First parameter is void
+			storedProcedureQuery.setParameter( 2, 2 );
+			storedProcedureQuery.setParameter( 3, "title'2" );
 
-		@SuppressWarnings("unchecked")
-		List<Car> listResult = storedProcedureQuery.getResultList();
-		assertThat( listResult ).containsExactly( new Car( 2, "title'2" ) );
+			@SuppressWarnings("unchecked")
+			List<Car> listResult = storedProcedureQuery.getResultList();
+			assertThat( listResult ).containsExactly( new Car( 2, "title'2" ) );
+		} );
 	}
 
 	@Test
 	public void testResultSetStaticCallRaw() throws Exception {
-		StoredProcedureQuery storedProcedureQuery = em.createNamedStoredProcedureQuery( "returnPositionalParametersRaw" );
-		// First parameter is void
-		storedProcedureQuery.setParameter( 2, 2 );
-		storedProcedureQuery.setParameter( 3, "title'2" );
+		inTransaction( entityManager -> {
+			StoredProcedureQuery storedProcedureQuery = entityManager.createNamedStoredProcedureQuery(
+					"returnPositionalParametersRaw" );
+			// First parameter is void
+			storedProcedureQuery.setParameter( 2, 2 );
+			storedProcedureQuery.setParameter( 3, "title'2" );
 
-		@SuppressWarnings("unchecked")
-		List<Object[]> listResult = storedProcedureQuery.getResultList();
-		assertThat( listResult ).containsExactly( new Object[]{ 2, "title'2" } );
+			@SuppressWarnings("unchecked")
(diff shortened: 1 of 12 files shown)


fax4ever:
ok to test

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BaseNeo4jDialect.java
```diff
@@ -297,4 +300,18 @@ protected EntityKeyMetadata entityKeyMetadata(EntityKeyMetadata keyMetadata, Tup
...
 	protected abstract A createNeo4jAssociationQueries(EntityKeyMetadata ownerEntityKeyMetadata, AssociationKeyMetadata associationKeyMetadata);
+
+	protected Map.Entry<String, Map> buildProcedureQueryWithParams(String storedProcedureName,
+			ProcedureQueryParameters queryParameters) {
+		StringBuilder queryBuilder = new StringBuilder();
+		queryBuilder.append( "CALL " ).append( storedProcedureName ).append( "(" );
+		Map namedParams = new HashMap<>();
+		List parameters = queryParameters.getPositionalParameters();
+		for ( int i = 0; i < parameters.size(); i++ ) {
+			queryBuilder.append( "{a" + i + "}," );
```
fax4ever: I think it's fine, I've just a curiosity: why add an 'a' in front of parameter name?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BoltNeo4jDialect.java
```diff
@@ -589,6 +592,31 @@ public void forEachTuple(ModelConsumer consumer, TupleTypeContext tupleTypeConte
...
+	@Override
+	public ClosableIterator<Tuple> callStoredProcedure(
+			String storedProcedureName, ProcedureQueryParameters queryParameters, TupleContext tupleContext) {
+		Map.Entry<String, Map> queryAndParams = buildProcedureQueryWithParams( storedProcedureName, queryParameters );
+		Transaction transaction = transaction( tupleContext );
+		if ( transaction == null ) {
+			DatastoreProvider datastoreProvider = getServiceRegistry().getService( DatastoreProvider.class );
+			BoltNeo4jDatastoreProvider neo4jProvider = (BoltNeo4jDatastoreProvider) datastoreProvider;
+			BoltNeo4jClient client = neo4jProvider.getClient();
+			transaction = client.getDriver().session().beginTransaction();
```
fax4ever: Actually I don't know very well the neo4j transaction interface. Can we sure that the transaction will be closed in all cases?

fax4ever: Shouldn't the transaction be opened by the client?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/HttpNeo4jDialect.java
```diff
@@ -486,6 +490,46 @@ public void forEachTuple(ModelConsumer consumer, TupleTypeContext tupleTypeConte
...
+		statement.setResultDataContents( Collections.singletonList( Statement.AS_ROW ) );
+		Statements statements = new Statements();
+		statements.addStatement( statement );
+		Long txId = transactionId( tupleContext.getTransactionContext() );
+		StatementsResponse response;
+		if ( txId != null ) {
+			response = client.executeQueriesInOpenTransaction( txId, statements );
+		}
+		else {
+			response = client.executeQueriesInNewTransaction( statements );
```
fax4ever: I'm just thinking about: if there isn't an ongoing transaction, should be allow to execute a procedure? It seems that neo4j doesn't allow that. Should we allow it or not?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BaseNeo4jDialect.java
grin7728: Just not to cast value of parameter identifier to string manually on next line

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BoltNeo4jDialect.java
grin7728: Thank you for comment, I'll investigate the question for bolt dialect more carefully

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BoltNeo4jDialect.java
```diff
@@ -597,13 +597,18 @@ public void forEachTuple(ModelConsumer consumer, TupleTypeContext tupleTypeConte
...
 		Transaction transaction = transaction( tupleContext );
+		StatementResult result;
 		if ( transaction == null ) {
 			DatastoreProvider datastoreProvider = getServiceRegistry().getService( DatastoreProvider.class );
 			BoltNeo4jDatastoreProvider neo4jProvider = (BoltNeo4jDatastoreProvider) datastoreProvider;
 			BoltNeo4jClient client = neo4jProvider.getClient();
-			transaction = client.getDriver().session().beginTransaction();
+			try ( Session session = client.getDriver().session() ) {
+				result = session.run( queryAndParams.getKey(), queryAndParams.getValue() );
+			}
```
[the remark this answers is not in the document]
fax4ever: I like this solution. Thanks!

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BoltNeo4jDialect.java
```diff
@@ -589,6 +592,36 @@ public void forEachTuple(ModelConsumer consumer, TupleTypeContext tupleTypeConte
...
+
+	private List<Tuple> extractTuples(StatementResult result) {
+		List<Tuple> tuples = new ArrayList<>();
+		while ( result.hasNext() ) {
+			Tuple tuple = new Tuple();
+			result.next().asMap().forEach( (key, value) -> tuple.put( key, value ) );
+			tuples.add( tuple );
+		}
+		return tuples;
+	}
```
DavideD: I don't think you need this method, can't you use the following?
```
return new BoltNeo4jMapsTupleIterator( statementResult );
```
Check what we do for `executeBackendQuery`.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/EmbeddedNeo4jDialect.java
```diff
@@ -533,6 +537,24 @@ public void forEachTuple(ModelConsumer consumer, TupleTypeContext tupleTypeConte
...
+	}
+
+	private List<Tuple> extractTuples(Result result) {
+		List<Tuple> tuples = new ArrayList<>();
+		while ( result.hasNext() ) {
+			Tuple tuple = new Tuple();
+			result.next().forEach( (key, value) -> tuple.put( key, value ) );
+			tuples.add( tuple );
+		}
+		return tuples;
```
DavideD: Similar to the previous comment, check what we do for `executeBackendQuery`. You should be able to reuse the code we have to convert the result.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/HttpNeo4jDialect.java
DavideD: Operations with Neo4j should always occur inside a transactions, we wrote this in the [documentation](https://docs.jboss.org/hibernate/ogm/5.4/reference/en-US/html_single/#ogm-neo4j-transactions):

```
In Neo4j, operations must be executed inside a transaction.

Make sure your interactions with Hibernate OGM are within a transaction when you target Neo4J.
```

Make sure to check what we do in `executeBackendQuery` and reuse the same code we have when the returned entity is not specified.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/remote/http/transaction/impl/HttpNeo4jResourceLocalTransactionCoordinator.java
```diff
@@ -330,7 +330,7 @@ public void markRollbackOnly() {
 
 		@Override
 		public Object getTransactionId() {
-			return tx.getId();
+			return tx != null ? tx.getId() : null;
```
DavideD: Transaction is mandatory.

What kind of error we show in case a transaction is not present? If it's not clear enough we can improve it. Anyway, this would be a separate issue

core/src/test/java/org/hibernate/ogm/backendtck/storedprocedures/PositionalParametersStoredProcedureCallTest.java
```diff
@@ -46,7 +44,7 @@
  * @author Sergey Chernolyas &amp;sergey_chernolyas@gmail.com&amp;
  */
 @SkipByGridDialect(
-		value = { HASHMAP, INFINISPAN, INFINISPAN_REMOTE, NEO4J_EMBEDDED, NEO4J_REMOTE },
+		value = { HASHMAP, INFINISPAN, INFINISPAN_REMOTE },
```
DavideD: This won't work on CI because the build doesn't load the stored procedures for the remote dialects

core/src/test/java/org/hibernate/ogm/backendtck/storedprocedures/PositionalParametersStoredProcedureCallTest.java
```diff
@@ -46,7 +45,7 @@
  * @author Sergey Chernolyas &amp;sergey_chernolyas@gmail.com&amp;
  */
 @SkipByGridDialect(
-		value = { HASHMAP, INFINISPAN, INFINISPAN_REMOTE, NEO4J_EMBEDDED, NEO4J_REMOTE },
+		value = {HASHMAP, INFINISPAN, INFINISPAN_REMOTE, NEO4J_REMOTE},
 		comment = "These dialects not support stored procedures with positional parameters")
 @TestForIssue(jiraKey = { "OGM-359" })
 public class PositionalParametersStoredProcedureCallTest extends OgmJpaTestCase {
```
DavideD: What about `NamedParametersStoredProcedureCallTest`?

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/BaseNeo4jDialect.java
DavideD: > Just not to cast value of parameter identifier to string manually on next line

Please, use the casting. I'm not a big fun of this tricks. They just make the code weird.
You will also be able to reuse (most of) the code for named parameters if you avoid it.

DavideD:
About the tests for the remote dialects, we should at least create an integration test in the other module if it's not possible to create one in the neo4j module.

@grin7728 How do you upload a stored procedure on Neo4j? Can't you create it with a query?

Thanks

grin7728:
Hello @DavideD  I uploaded JAR file with stored procedure to neo4j instance manually. I didn't find a way to create procedure with driver or http from client side. 
I have one more question. 
As I understood, these tests from \core\src\test\java\org\hibernate\ogm\backendtck\storedprocedures don't use transactions.  So do I need to create similar tests but with transactions?

DavideD:
> Hello @DavideD I uploaded JAR file with stored procedure to neo4j instance manually. I didn't find a way to create procedure with driver or http from client side.

OK, but how the part of the stored procedure looks like. Is it a java class? What do you need to add to the JAR to upload the stored procedures? Can you publish the jar somewhere so that I can have a look?

> As I understood, these tests from \core\src\test\java\org\hibernate\ogm\backendtck\storedprocedures don't use transactions. So do I need to create similar tests but with transactions?

I think that's an error, because MongoDB doesn't have transactions we forgot to open them.
Please, update the existing tests so that a transaction is used. You can use the method from the superclass: `inTransaction( consumer );`.

There are several examples in the source code using it.

Thanks a lot

grin7728:
@DavideD  here you can find an example jar https://drive.google.com/file/d/1X3HQsojySOLz-bZX03JVxxDFPGBJfUek/view 
The procedure itself can be found in this PR CarStoredProcedures.java  (the same class that used for embedded neo4j testing). And here is a template for procedure creation https://github.com/neo4j-examples/neo4j-procedure-template

core/src/test/java/org/hibernate/ogm/backendtck/storedprocedures/PositionalParametersStoredProcedureCallTest.java
grin7728: I've added tests for named params

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/EmbeddedNeo4jDialect.java
grin7728: Hello @DavideD  is it about EmbeddedNeo4jBackendQueryResultIterator? Because it processes result slightly different, for example it change the order of parameters somehow in tests

DavideD: I'm afraid I'm not following you. Can you give me an example?
I don't see any difference between what you are doing and EmbeddedNeo4jBackendQueryResultIterator#convert (or better, mapSnapshot(...))

grin7728: @DavideD  I think, the problem is related to random order of columns in Result class, org.neo4j.graphdb.result.next which is called in EmbeddedNeo4jBackendQueryResultIterator return Map<String, Object> where order of keys are not the same as order of columns returned embedded neo4j. Result class provide a method org.neo4j.graphdb.result.columns which return the columns in correct order, but EmbeddedNeo4jBackendQueryResultIterator doesn't take it into account.

grin7728: @DavideD  I think, the problem is related to random order of columns in Result class, org.neo4j.graphdb.result.next which is called in EmbeddedNeo4jBackendQueryResultIterator return Map<String, Object> where order of keys are not the same as order of columns returned embedded neo4j. Result class provide a method org.neo4j.graphdb.result.columns which return the columns in correct order, but EmbeddedNeo4jBackendQueryResultIterator doesn't take it into account.

grin7728: @DavideD  I think, the problem is related to random order of columns in Result class, org.neo4j.graphdb.result.next which is called in EmbeddedNeo4jBackendQueryResultIterator return Map<String, Object> where order of keys are not the same as order of columns returned embedded neo4j.

grin7728: Hello, the problem here is result.next() returns Map<String, Object> where keys are not in the same order as returned columns returned by neo4j embedded.

neo4j/src/main/java/org/hibernate/ogm/datastore/neo4j/embedded/dialect/impl/EmbeddedNeo4jBackendQueryResultIterator.java
```diff
@@ -64,6 +68,13 @@ private TupleSnapshot createSnapshot(Map<String, Object> next) {
 
 	private TupleSnapshot mapSnapshot(Map<String, Object> next) {
 		TupleSnapshot snapshot;
+		if ( this.columns != null ) {
+			Map<String, Object> sortedColumns = new LinkedHashMap<>();
+			for ( String column : this.columns ) {
+				sortedColumns.put( column, next.get( column ) );
+			}
+			next = sortedColumns;
+		}
```
DavideD: Could you create a test for this? Just in case somebody in the future decide to change it? ALso, are there any test failures in the build if this is not committed? Thanks

grin7728: Yes, I've created a test which will fail if this code is removed. Also PositionalParametersStoredProcedureCallTest.testResultSetStaticCallRaw will fail, because in this test we have no EntityMetadataInformation or no other source of information about proper column order

DavideD: Thanks

DavideD:
It seems that this branch has some conflicts with the current master. Could you check and rebase it, please? Thanks

grin7728:
> It seems that this branch has some conflicts with the current master. Could you check and rebase it, please? Thanks

Done, sorry for that, I misunderstood the reason of the git conflict message

neo4j/src/test/java/org/hibernate/ogm/datastore/neo4j/test/embedded/dialect/EmbeddedNeo4jBackendQueryResultIteratorTest.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.fest.assertions.Assertions.assertThat;
+
+
+public class EmbeddedNeo4jBackendQueryResultIteratorTest {
```
fax4ever: Maybe we should add a JavaDoc here, saying that we're testing the order consistency?

grin7728: Thanks a lot

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/1076

DavideD:
Thanks @grin7728,
Great job!

I just changed the test for the columns order so that it uses Mockito instead of the Dummy class. This way less code is needed.

I will create a new issue for testing the  procedures on a remote Neo4j instance, I tested manually and it works but I think it's going to be a bit complicated with the current build process.

Cheers

fax4ever:
Thanks @grin7728 !

DavideD:
FYI: https://hibernate.atlassian.net/browse/OGM-1540

grin7728:
@DavideD @fax4ever  thanks a lot, I'm glad to make my contribution to your beautiful project

PR #1077, opened 2018-09-21
OGM-1458 Change mapping of byte type field to ProtoBuf int32

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1458

See also: https://github.com/infinispan/infinispan/pull/6191

### DIFF
--- a/core/src/test/java/org/hibernate/ogm/backendtck/queries/parameters/QueryWithParametersTest.java
+++ b/core/src/test/java/org/hibernate/ogm/backendtck/queries/parameters/QueryWithParametersTest.java
@@ -35,7 +35,7 @@ public class QueryWithParametersTest extends OgmJpaTestCase {
 	public PackagingRule packaging = new PackagingRule( "persistencexml/ogm.xml", Movie.class );
 
 	@Test
-	@SkipByGridDialect(value = { INFINISPAN_REMOTE }, comment = "Query on Byte field are currently not supported")
+	@SkipByGridDialect(value = { INFINISPAN_REMOTE }, comment = "Due to ISPN-7863, fields starting with 'v' are currently not supported.")
 	public void canUseByteForSimpleComparison() {
 		EntityManager entityManager = getFactory().createEntityManager();
 		entityManager.getTransaction().begin();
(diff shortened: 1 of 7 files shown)


core/src/test/java/org/hibernate/ogm/backendtck/queries/parameters/QueryWithParametersTest.java
```diff
@@ -35,7 +35,7 @@
 	public PackagingRule packaging = new PackagingRule( "persistencexml/ogm.xml", Movie.class );
 
 	@Test
-	@SkipByGridDialect(value = { INFINISPAN_REMOTE }, comment = "Query on Byte field are currently not supported")
+	@SkipByGridDialect(value = { INFINISPAN_REMOTE }, comment = "Query on --starting with 'v'-- field are currently not supported")
```
gsmet: This looks weird. What does it mean?

anistor: A strange reference to ISPN-7863 , maybe :))))

gsmet: Hum, OK. AFAICS, it's fixed in 9.3.4.Final so I don't think it's worth it to reference this ISPN bug in our code.

fax4ever: Thank you @gsmet and @anistor. I will change the comment here.

infinispan-remote/src/main/java/org/hibernate/ogm/datastore/infinispanremote/impl/protobuf/ByteProtofieldAccessor.java
```diff
@@ -11,29 +11,24 @@
 
 /**
  * There is no explicit "byte" support in Protostream.
- * Decided to encode it as a byte array type; on reading a longer sequence it will be truncated as we only read the first byte.
+ * Decided to encode it as a int type; on reading an int will be truncated as a byte,
+ * according to the endian behavior adopted by the {@link Integer#byteValue()} method
```
Sanne: What _endian behaviour_ ?

We need to make sure the encoding is the same independently from the specific platform this is running on.

I assumed the byteValue() method provided a stable specification, but either I'm wrong or this comment is overly concerning :)

fax4ever: this comment is definitely overly concerning :) I think we can rely on a consistency of Integer#byteValue(). I will change it. Thank you :)

fax4ever:
@Sanne I changed the comment here :)

Sanne:
merged!

fax4ever:
Thank you @Sanne @anistor @gsmet for the advice here

PR #1071, opened 2018-08-23
OGM-786. Support GridFS in MongoDB

schernolyas:
https://hibernate.atlassian.net/browse/OGM-786

PR #1080, opened 2018-09-28
[PREVIEW] OGM-786 GridFS support

DavideD:
https://hibernate.atlassian.net/browse/OGM-1527

This is not ready to be merged.
It follows https://github.com/hibernate/hibernate-ogm/pull/1071

@schernolyas Can you have a look at this PR? I removed the proxies but I had to update core.

Something is not clear to me though, how come did you use Blob instead of byte[]?

### DIFF
--- a/core/src/main/java/org/hibernate/ogm/type/descriptor/impl/GridValueBinder.java
+++ b/core/src/main/java/org/hibernate/ogm/type/descriptor/impl/GridValueBinder.java
@@ -7,6 +7,7 @@
 package org.hibernate.ogm.type.descriptor.impl;
 
 import org.hibernate.ogm.model.spi.Tuple;
+import org.hibernate.type.descriptor.WrapperOptions;
 
 /**
  * Contract to bind a value to the resultset
@@ -14,6 +15,10 @@ import org.hibernate.ogm.model.spi.Tuple;
  * @author Emmanuel Bernard
  */
 public interface GridValueBinder<X> {
-	//WrappedOptions for streams?
+
 	void bind(Tuple resultset, X value, String[] names);
+
+	default void bind(Tuple rs, X value, String[] names, WrapperOptions options) {
+		bind( rs, value, names );
+	}
 }
--- a/core/src/main/java/org/hibernate/ogm/type/descriptor/impl/GridValueExtractor.java
+++ b/core/src/main/java/org/hibernate/ogm/type/descriptor/impl/GridValueExtractor.java
@@ -7,6 +7,7 @@
 package org.hibernate.ogm.type.descriptor.impl;
 
 import org.hibernate.ogm.model.spi.Tuple;
+import org.hibernate.type.descriptor.WrapperOptions;
 
 /**
  * Extract value from the result set
@@ -14,6 +15,10 @@ import org.hibernate.ogm.model.spi.Tuple;
  * @author Emmanuel Bernard
  */
 public interface GridValueExtractor<X> {
-	//WrappedOptions for streams?
+
 	X extract(Tuple resultset, String name);
+
+	default X extract(Tuple resultset, String name, WrapperOptions options) {
+		return extract( resultset, name );
+	}
 }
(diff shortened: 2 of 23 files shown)


schernolyas:
Hi @DavideD !

Main reason for using Blob insteed byte[] is size of binary content. GridFS recomended for using if you will store binary content with size greater than 16M. You can read about it [here](https://docs.mongodb.com/manual/reference/limits/#limit-bson-document-size)

DavideD:
I'm not following you. Are you saying that you cannot create a byte[] of 16 MB or more?

schernolyas:
Hi @DavideD !

Of course, I can. But ... Is it optimally ? BLOB allow read binary content chunk by chunk. It save a memory.

schernolyas:
If you see to class BlobMappedGridTypeDescriptor and org.hibernate.engine.jdbc.BlobProxy ... 
in theory, I can do it. I need a check reaction if I set lenght of stream to 0.  Because  we can say something about length if we read all bytes.

DavideD:
The problem is that `java.sql.Blob` class is really used for JDBC drivers to map the value returned from the DB. I don't think it should be used for mapping the field of the entities.

DavideD:
In particular, for Photo class it doesn't make sense to use the `@Lob` annotation and the Blob class.
The @Lob annotation is supposed to be used specifically for the use cases where big files need to be stored.

DavideD:
I think we should focus on the use cases where the user wants to use String or byte[]. Specifically these:

```
	@Lob
	@GridFSBucket(BUCKET_NAME)
	@BinaryStorage( GRID_FS )
	private String blobAsString;

        @Lob
	@GridFSBucket(BUCKET_NAME)
	@BinaryStorage( GRID_FS )
	private byte[] blobAsBytes;
```

I also updated the test so that it creates  30MB bytes array to make sure that it works with those sizes.

schernolyas:
Okey Davide. Do as you think.

DavideD:
By the way, why do you also need to store the whole  `TableEntityTypeMapping` in the `MongoDBDatastoreProvider`? I think a better a approach would be to scan, during the schema initialization, the classes and the fields and only create a map inside BinaryStorageManager with only the fields requiring the GridFS approach. The BinaryStorageManager should be stored in the MongoDBDatastoreProvider and not vice-versa. 
Now that I think about it, the `@Lob` annotation is probably not required.

The dalect right now is also storing the files using GridFS a bit to soon. I guess this requires to update the `BatchInsertionTask` so that during the `flushAll` we can get wich files needs to be stored in the GridFS bucket.

DavideD:
> Okey Davide. Do as you think.

I'm not planning to work on this right now, if you want, you can take this branch I just created and continue to work on it. I think now we have a better Idea about what it's needed.

About `Blob`, if Hibernate ORM supports it as a mapping type (can you check this? Maybe you can ask on HipChat), I guess we can keep it. But we also need to support the other use cases.

schernolyas:
Yes. ORM supported.

With best regards, Sergey Chernolyas

вс, 30 сент. 2018 г., 0:04 DavideD <notifications@github.com>:

DavideD:
> Yes. ORM supported.

Well, let's keep it then

DavideD:
> Am I right that I have to support store byte[] and String as LOB to GridFS?

Not exactly, they should be stored as GridFS only if they are annotated with `@BinaryStorage( GRID_FS )`
I don't think the `@Lob` annotation is necessary for this to work.

DavideD:
I think this should be true for Blob as well, the reason is that even if the type is a Blob or a byte[] it might not be big enough to require the GridFS approach and storing it as a byte[] might be enough for the user. GridFS should be an additional option for users to save this information when sensible.

DavideD:
By the way, @schernolyas, apart form String and byte[], I think there are also another couple of things requiring some fixing:

1. The fact that we copy and keep track of the whole `TableEntityMapping`

> By the way, why do you also need to store the whole TableEntityTypeMapping in the MongoDBDatastoreProvider? I think a better a approach would be to scan, during the schema initialization, the classes and the fields and only create a map inside BinaryStorageManager with only the fields requiring the GridFS approach. The BinaryStorageManager should be stored in the MongoDBDatastoreProvider and not vice-versa.
Now that I think about it, the @Lob annotation is probably not required.

2. The fact that we create the GridFS file too soon

> The dalect right now is also storing the files using GridFS a bit to soon. I guess this requires to update the BatchInsertionTask so that during the flushAll we can get wich files needs to be stored in the GridFS bucket.

PR #1081, opened 2018-10-02
1425 mongodb tableperclass

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1425


fax4ever:
This is a `preview`: If the base class is abstract, the new test fails, because it tries to instantiate an abstract class. I would like to solve the issue and update the test in subsequent commits.

fax4ever:
Thank you!

PR #1084, opened 2018-10-09
OGM-786 GridFS support

schernolyas:

### DIFF
--- a/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/gridfs/GridFSTest.java
+++ b/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/gridfs/GridFSTest.java
@@ -9,6 +9,8 @@ package org.hibernate.ogm.datastore.mongodb.test.gridfs;
 import static org.fest.assertions.Assertions.assertThat;
 
 import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.sql.Blob;
 import java.sql.SQLException;
@@ -32,6 +34,7 @@ import com.mongodb.client.MongoDatabase;
 import com.mongodb.client.gridfs.GridFSBucket;
 import com.mongodb.client.gridfs.GridFSBuckets;
 import com.mongodb.client.gridfs.model.GridFSFile;
+import org.apache.commons.io.IOUtils;
 
 /**
  * @author Davide D'Alto
@@ -41,10 +44,10 @@ import com.mongodb.client.gridfs.model.GridFSFile;
 @TestForIssue(jiraKey = "OGM-786")
 public class GridFSTest extends OgmJpaTestCase {
 
-	// GridFS is usually for file of size bigger than 16 MB
-	private final static int BLOB_SIZE = 30 * 1024 * 1024; // 30 MB
-
 	static final String BUCKET_NAME = "photos";
+
+	// GridFS is usually for file of size bigger than 16 MB
+	private static final int BLOB_SIZE = 30 * 1024 * 1024; // 30 MB
 
 	private static final byte[] BLOB_CONTENT_1 = createByteArray( 'a', BLOB_SIZE );
 	private static final byte[] BLOB_CONTENT_2 = createByteArray( 'x', BLOB_SIZE );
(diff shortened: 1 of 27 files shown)


mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/gridfs/GridFSTest.java
```diff
@@ -153,6 +164,8 @@ public void canRemoveEntityWithBlob() {
 			assertBlobAreEqual( photo.getContentAsBlob(), BLOB_CONTENT_1 );
 			assertThat( photo.getContentAsByteArray() ).isNotNull();
 			assertThat( photo.getContentAsByteArray() ).isEqualTo( BLOB_CONTENT_1 );
+			assertThat( photo.getContentAsString() ).isNotNull();
+			assertThat( photo.getContentAsString() ).isEqualTo( BLOB_STRING_CONTENT_1 );
```
DavideD: I don't think these tests are right for this issue. They would pass even if you don't store the field using GridFS. Is there a way to check if the field has been store effectively using GridFS?

Also there is no need to asser

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/gridfs/GridFSTest.java
```diff
@@ -153,6 +164,8 @@ public void canRemoveEntityWithBlob() {
 			assertBlobAreEqual( photo.getContentAsBlob(), BLOB_CONTENT_1 );
 			assertThat( photo.getContentAsByteArray() ).isNotNull();
 			assertThat( photo.getContentAsByteArray() ).isEqualTo( BLOB_CONTENT_1 );
+			assertThat( photo.getContentAsString() ).isNotNull();
```
DavideD: This line is not needed, the  isEqualTo already take care of null values

schernolyas: Ok. I will remove excess line.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/gridfs/GridFSTest.java
schernolyas: you right.  About your question. Yes. We can check GridFS. We can write additional information to meta.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/gridfs/GridFSTest.java
```diff
@@ -0,0 +1,197 @@
...
+			photo.setContentAsByteArray( BLOB_CONTENT_1 );
+			photo.setContentAsString( BLOB_STRING_CONTENT_1 );
+			em.persist( photo );
+		} );
+
+		inTransaction( em -> {
+			MongoDatabase mongoDatabase = getCurrentDB( em );
+			GridFSBucket gridFSFilesBucket = GridFSBuckets.create( mongoDatabase, BUCKET_NAME );
+			MongoCursor<GridFSFile> cursor = gridFSFilesBucket.find().iterator();
+			for ( int i = 0; i < 3; i++ ) {
```
DavideD: I don't understand why you are using a `for` loop in this scenario when you have a cursor.
What happens if the cursor returns 10 values but you only check the first 3?
Also, where the number 3 is coming from?

schernolyas: I understood what you mean. Thank for 
 criticism and idea!

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/gridfs/GridFSTest.java
DavideD: Yes, I guess... you need to make sure that it contains the values you expect. Does that asnwer your question?

schernolyas: Hi @DavideD !

Yes,of course.

DavideD:
I'm going to close this one.
After trying to refactor this PR I realized that this approach of using the annotation doesn't work well with our existing code and make it harder to have custom types.

I also decided that it's not helpful to assume that there are going to be other binary storages for now. This will simplify the code a lot,

Even if we go with this approach there are some cases there are not covered by tests and make the PR fails.

I will send a new PR when I have time. Currently, there is PREVIEW on my repository: https://github.com/DavideD/hibernate-ogm/tree/OGM-786-Preview

Thanks

PR #1092, opened 2018-10-24
OGM-786 Support for GridFS in MongoDB

DavideD:

This PR follows: https://github.com/hibernate/hibernate-ogm/pull/1084

This solution is not as nice as the one showed in the issue but it's simpler to implement and more flexible for the user.
I've also fixed some issues that weren't covered in the original test case and added some documentation that should explain what this issue is about.

We might also want to add some native mapping tests, but I will kepp it for a separate PR.

I' not sure we should do include a new feature after a version in CR, but I'm inclined to include it anyway since it doesn't affect existing code.

### DIFF
--- a/documentation/manual/src/main/asciidoc/modules/mongodb.asciidoc
+++ b/documentation/manual/src/main/asciidoc/modules/mongodb.asciidoc
@@ -475,6 +475,95 @@ in the corresponding object in the db.
 This can have consequences when it comes to queries on null value.
 ====
 
+[[mongodb-gridfs-support]]
+==== GridFS Support
+
+This feature is experimental.
+
+GridFS is a specification for storing and retrieving files that
+exceed the BSON-document size limit of 16 MB.
+
+You can find more details about it in
+the https://docs.mongodb.com/manual/core/gridfs/[official MongoDB documentation].
+
+It's possible to store a value as GridFS using the type
+`org.hibernate.ogm.datastore.mongodb.type.GridFS`.
+
+.Field mapped using GridFS
+====
+[source, JAVA]
+----
+import org.hibernate.ogm.datastore.mongodb.type.GridFS;
+
+@Entity
+public class Photo {
+
+    @Id
+    String name;
+
+    GridFS image;
+
+    // ...
+}
+----
+
+----
+> db.Photo.find()
+{ "_id" : "photo.jpg", "image" : ObjectId("5bce7202826ce81d7fe7ebe5") }
+
+> db.Photo_bucket.files.find().pretty()
+{
+        "_id" : ObjectId("5bce7202826ce81d7fe7ebe5"),
+        "filename" : "image_photo.jpg",
+        "length" : NumberLong(30000000),
+        "chunkSize" : 261120,
+        "uploadDate" : ISODate("2018-10-22T20:57:37.047Z"),
+        "md5" : "a08988fc1e3da0e80d2ba7b55599d1f7"
+}
+----
+====
+
+The default bucket name is the name of the class with the suffix `_bucket` while the name
+of the filename is the combination of the field name with the id of the document.
+
+It's possible to select a different bucket name using the annotation `@GridFSBucket`.
+
+.Field mapped using GridFS with custom bucket name
+====
+[source, JAVA]
+----
+import org.hibernate.ogm.datastore.mongodb.type.GridFS;
+import org.hibernate.ogm.datastore.mongodb.options.GridFSBucket;
+
+@Entity
+public class Photo {
+
+    @Id
+    String name;
+
+    @GridFSBucket("PhotoGallery")
+    GridFS image;
+
+    // ...
+}
+----
+
+----
+> db.Photo.find()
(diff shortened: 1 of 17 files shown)


DavideD:
The checkstyle error should be fixed now

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1567,7 +1580,7 @@ else if ( operation instanceof RemoveTupleOperation ) {
 				operation = queue.poll();
 			}
 
-			flushInserts( inserts );
+			flushInserts( provider, inserts );
```
fax4ever: Why do we pass the provider to the `flushInsert` method? Maybe we could just use the field in the method.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/gridfs/GridFSTest.java
```diff
@@ -0,0 +1,293 @@
...
+import com.mongodb.client.gridfs.GridFSBuckets;
+import com.mongodb.client.gridfs.model.GridFSFile;
+
+/**
+ * @author Davide D'Alto
+ * @author Sergey Chernolyas &amp;sergey_chernolyas@gmail.com&amp;
+ * @see <a href="http://mongodb.github.io/mongo-java-driver/3.4/driver/tutorials/gridfs/">GridFSBucket</a>
+ */
+@TestForIssue(jiraKey = "OGM-786")
+public class GridFSTest extends OgmJpaTestCase {
```
fax4ever: The test suite is really complete, maybe we could just add a query load case on it.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/binarystorage/GridFSStorageManager.java
```diff
@@ -0,0 +1,168 @@
...
+ * <p>
+ * GridFS is a specification for storing and retrieving files that exceed the BSON-document size limit of 16 MB.
+ * <p>
+ * The default bucket name is the name of the class followed by the suffix "_bucket". The file name is created using the
+ * field name and the id of the document.
+ *
+ * @see <a href="https://docs.mongodb.com/manual/core/gridfs/">MongoDB GridFS documentation</a>
+ * @author Sergey Chernolyas &amp;sergey_chernolyas@gmail.com&amp;
+ */
+public class GridFSStorageManager {
```
fax4ever: Good idea to use a new class for this

documentation/manual/src/main/asciidoc/modules/mongodb.asciidoc
```diff
@@ -475,6 +475,95 @@ in the corresponding object in the db.
 This can have consequences when it comes to queries on null value.
 ====
 
+[[mongodb-gridfs-support]]
```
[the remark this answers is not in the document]
DavideD: I prefer it her because it's where we just after we described the supported types

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/gridfs/GridFSTest.java
DavideD: What do you mean?

fax4ever: I don't know whether it would have sense. I mean adding a test which uses a query to retrieve the entity, instead of a load by id.

DavideD: Ah ... good point! I will add one

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
fax4ever: Sorry, it was my mistake. I hadn't seen that the method `flushInserts` was static

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/gridfs/GridFSTest.java
fax4ever: Thanks!

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
DavideD: I don't think it has to be static, but in general I prefer to pass the parameters to the methods. It makes refactoring easier.  It also makes it simpler to figure out what a method needs. Sometimes you have methods with no parameters and then you discover that they use 30 different fields.
Obviously, this is not a strict rule, more a loose guide line.

fax4ever:
I opened a new one, with the test case #1094

PR #1094, opened 2018-10-25
786 mongodb gridfs

fax4ever:
That follows #1092.

I've only added a test case on top of the latter.


DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/1094
Thanks

PR #1095, opened 2018-10-31
OGM-1541 Upgrade Jackson to 2.9.5 (WF 13.0.0)

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1541

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -52,7 +52,7 @@
         <version.org.jboss.logging.jboss-logging-tools>2.1.0.Final</version.org.jboss.logging.jboss-logging-tools>
         <version.com.fasterxml.classmate>1.3.3</version.com.fasterxml.classmate>
         <version.org.jboss.resteasy>3.0.24.Final</version.org.jboss.resteasy> <!-- Used by CouchDB and Neo4j remote -->
-        <version.com.fasterxml.jackson>2.8.9</version.com.fasterxml.jackson>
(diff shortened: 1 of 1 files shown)


bom/pom.xml
```diff
@@ -52,7 +52,7 @@
         <version.org.jboss.logging.jboss-logging-tools>2.1.0.Final</version.org.jboss.logging.jboss-logging-tools>
         <version.com.fasterxml.classmate>1.3.3</version.com.fasterxml.classmate>
         <version.org.jboss.resteasy>3.0.24.Final</version.org.jboss.resteasy> <!-- Used by CouchDB and Neo4j remote -->
-        <version.com.fasterxml.jackson>2.8.9</version.com.fasterxml.jackson>
+        <version.com.fasterxml.jackson>2.9.7</version.com.fasterxml.jackson>
```
gsmet: The idea here is to be in line with the WF dependencies.

So not sure how we should deal with this.

fax4ever: Let me check the one provided by WF.

fax4ever: WildFly 13.0.0 uses jackson-core-2.9.5.jar

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/1095

Thanks

fax4ever:
Thanks @DavideD !

PR #1096, opened 2018-11-02
1538 association order by

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1538


fax4ever:
That's a preview. I'm going to add in subsequent commits the relative dialect implementations.

core/src/main/java/org/hibernate/ogm/model/spi/AssociationOrderBy.java
```diff
@@ -0,0 +1,36 @@
...
+	public enum Order {
+		ASC, DESC
+	}
+
+	private final String columnName;
+	private final Order kind;
+
+	public AssociationOrderBy(String expression, String order) {
+		columnName = expression.substring( 1, expression.length() - 1 );
+		kind = ( "desc".equalsIgnoreCase( order ) ) ? Order.DESC : Order.ASC;
```
DavideD: Nit picking, but we don't need the extra brackets. Let's just keep: 
```
kind =  "desc".equalsIgnoreCase( order ) ? Order.DESC : Order.ASC;
```

core/src/main/java/org/hibernate/ogm/model/spi/AssociationOrderBy.java
```diff
@@ -0,0 +1,36 @@
...
+
+	public enum Order {
+		ASC, DESC
+	}
+
+	private final String columnName;
+	private final Order kind;
+
+	public AssociationOrderBy(String expression, String order) {
+		columnName = expression.substring( 1, expression.length() - 1 );
```
DavideD: What's going on here? This is the kind of place where a comment with an example might be helpful

core/src/main/java/org/hibernate/ogm/query/impl/OGMOrderByParser.java
```diff
@@ -0,0 +1,78 @@
...
+		// translated fragment.
+		OrderByFragmentParser parser = new OrderByFragmentParser( lexer, context );
+		try {
+			parser.orderByFragment();
+		}
+		catch ( HibernateException e ) {
+			throw e;
+		}
+		catch ( Throwable t ) {
+			throw new HibernateException( "Unable to parse order-by fragment", t );
```
DavideD: Shouldn't this be in the log interface?

core/src/main/java/org/hibernate/ogm/query/impl/OGMOrderByParser.java
```diff
@@ -0,0 +1,78 @@
...
+		}
+
+		OGMOrderByRendered rendered = new OGMOrderByRendered();
+		try {
+			rendered.orderByFragment( parser.getAST() );
+			return rendered.getOrderByItems();
+		}
+		catch (RecognitionException e) {
+			throw new HibernateException( "Unable to render order-by fragment", e );
```
DavideD: Shouldn't this be in the log interface?

DavideD:
At first glance it looks OK, I'm not sure how you would use this new information with the dialects.
Just added a couple of minor remarks (I forgot it was a preview :P)

core/src/test/java/org/hibernate/ogm/backendtck/associations/order/OpenSourceProject.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.persistence.OrderBy;
+
+@Entity
+public class OpenSourceProject {
+
+	@Id
+	private String name;
+
+	@OneToMany
+	@JoinTable(name = "OpenProject_contributors")
```
DavideD: Is the JoinTable Needed? I prefer to keep the test classes as simple as poissible

core/src/test/java/org/hibernate/ogm/backendtck/associations/order/OrderByQualifierTest.java
```diff
@@ -0,0 +1,66 @@
...
+		Programmer telegraph4eternity = new Programmer( CONTRIBUTOR3 );
+		Programmer fax4ever = new Programmer( CONTRIBUTOR4 );
+		Programmer phone4induction = new Programmer( CONTRIBUTOR5 );
+
+		// that's the original order:
+		project.addContributor( unclebob );
+		project.addContributor( ddalto );
+		project.addContributor( telegraph4eternity );
+		project.addContributor( fax4ever );
+		project.addContributor( phone4induction );
```
DavideD: Instead of having the Id as constants, why don't you create all the entities as constants? This makes the unit test shorter and easier to read

core/src/test/java/org/hibernate/ogm/backendtck/associations/order/OrderByQualifierTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+
+	public static final String PROJECT_NAME = "Hibernate OGM";
+
+	@Test
+	public void testOrderByQualifier() {
+		OpenSourceProject project = new OpenSourceProject();
+		project.setName( PROJECT_NAME );
+
+		Programmer unclebob = new Programmer( CONTRIBUTOR1 );
```
DavideD: I think a bit of consistency here ogoes a long way, it's hard for me to remember that the `CONTRIBUTOR1` is `unclebob`, `CONTRIBUTOR2` is `ddalto` and so on. Wouldn't be easier to call them `UNCLEBOB_ID`,` DDALTO_ID` and so on?
 
I mentioned this before but in this case I would make the whole entities as constant.

core/src/test/java/org/hibernate/ogm/backendtck/associations/order/OpenSourceProject.java
DavideD: I think, in general, it's better to create new models when we introduce new tests. This way we can change the model without the risk of affecting existing tests.

core/src/main/java/org/hibernate/ogm/query/impl/OGMOrderByParser.java
fax4ever: Actually I don't know, because I think that it is not possible to raise the exception on Hibernate OGM side. Since the expression has been already parsed by Hibernate ORM before this code is executed. In other words, it wouldn't be possible create a API level test to verify the case.

core/src/main/java/org/hibernate/ogm/query/impl/OGMOrderByParser.java
fax4ever: Same as above. But if you want we can use the log interface.

core/src/main/java/org/hibernate/ogm/query/impl/OGMOrderByParser.java
DavideD: Not sure if we are talking about the same thing.
I'm just saying to replace this line with
```
throw log.unableToParseOrderBy( e );
```

DavideD: You don't need to add a test case for it's not possible

core/src/main/java/org/hibernate/ogm/model/spi/AssociationOrderBy.java
DavideD: Ah :)
It's obvious now that there is a comment

fax4ever:
@DavideD now tests pass, so not more preview.
I'm updating the doc, explaining that now @OrderBy is now supported.

core/src/test/java/org/hibernate/ogm/backendtck/associations/order/OrderByQualifierTest.java
```diff
@@ -0,0 +1,80 @@
...
+		project.addContributor( FAX_4_EVER );
+		project.addContributor( PHONE_4_INDUCTION );
+
+		inTransaction( session -> persistAll( session, UNCLEBOB, DDALTO, TELEGRAPH_4_ETERNITY, FAX_4_EVER, PHONE_4_INDUCTION, project ) );
+
+		inTransaction( session -> {
+			OpenSourceProject load = session.load( OpenSourceProject.class, PROJECT_NAME );
+
+			// TODO: Switch to AssertJ, so that we will able to do something like that:
+			// assertThat( load.getContributors() ).containsExactly( ... );
```
DavideD: This comment is misleading, you can already use containsExactly

fax4ever:
Thanks for merging it, @DavideD!

PR #1097, opened 2018-11-09
OGM-1530 Aggregation operations in HQL

Krerg:

PR #1098, opened 2018-11-15
OGM-1530 Aggregate operations for MongoDB

DavideD:
https://hibernate.atlassian.net/browse/OGM-1530

A follow up for https://github.com/hibernate/hibernate-ogm/pull/1097

This is a refactoring of the PR from @Krerg, @fax4ever Can you have a quick look, please?


mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/MongoDBDialect.java
```diff
@@ -1433,7 +1442,7 @@ private static int doReplaceOne(final MongoDBQueryDescriptor queryDescriptor, fi
 			collation = ( col != null ) ? getCollation( col ) : null;
 		}
 
-		final UpdateOptions updateOptions = new UpdateOptions().upsert( upsert ).collation( collation );
+		final ReplaceOptions updateOptions = new ReplaceOptions().upsert( upsert ).collation( collation );
```
fax4ever: maybe we can change variable name in `replaceOptions` to be consistent with its type

fax4ever:
@Krerg and @DavideD great work guys! I merged with a little change (a remove a tab for the checkstyle and I've changed the name of a local variable). Great idea to use a dedicated object to parse and render the aggregation query part. Very good!

PR #1099, opened 2018-11-19
OGM-1542 Fix distinct count.

Krerg:
- Add 'group' pipeline in case distinct count


core/src/test/java/org/hibernate/ogm/backendtck/queries/AggregateOperationQueryTest.java
```diff
@@ -52,6 +52,30 @@ public void shouldCountEntities() {
 		} );
 	}
 
+	@Test
+	public void shouldCountDistinct() {
+		inTransaction( em -> {
+			Number result = (Number) em.createQuery( "SELECT COUNT(author.name) FROM Author author" ).getSingleResult();
```
DavideD: Just a few quick questions, I haven't checked the code in detail:

- Shouldn't the tests look like ```SELECT COUNT( DISTINCT author.name ) ...```?

- I would also expect to see a couple of tests running the count on the same column with `COUNT` and `COUNT(DISTINCT  ...)` showing different results.

Am I missing anything?

Krerg: Oh sorry, I forget about 'distinct' word. There are some changes should be done

DavideD:
A summary of what needs to be change in this PR.

This test doesn't seem right:
```
Number result = (Number) em.createQuery( "SELECT COUNT(author.name) FROM Author author" ).getSingleResult();
assertThat( result.intValue() ).isEqualTo( 2 );
```

It's missing the DISTINCT keyword and should return 4 results (there are 4 authors in the db).
This should be similar to what happens in the other test that uses `COUNT(*)`.

Krerg:
@DavideD, I didn't forget) I was on the vacation,  I'll push changes in few days

DavideD:
No problem, I was checking this PR because I'm planning a small release and I think it would be nice to include this. If you can fix this in a couple of days that would be great.

DavideD:
Thanks by the way

Krerg:
Done. @DavideD, changes are ready to review

PR #1100, opened 2018-11-20
OGM-1523 Use WildFly 14 default Hibernate Search and ORM

fax4ever:
https://hibernate.atlassian.net/browse/OGM-1523

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -46,7 +46,7 @@
             See http://search.maven.org/#search|gav|1|g%3A"org.wildfly"%20AND%20a%3A"wildfly-parent"
         -->
 
-        <version.wildfly>13.0.0.Final</version.wildfly>
+        <version.wildfly>14.0.1.Final</version.wildfly>
         <version.org.hibernate.commons.annotations>5.0.3.Final</version.org.hibernate.commons.annotations>
         <version.org.jboss.logging.jboss-logging>3.3.2.Final</version.org.jboss.logging.jboss-logging>
         <version.org.jboss.logging.jboss-logging-tools>2.1.0.Final</version.org.jboss.logging.jboss-logging-tools>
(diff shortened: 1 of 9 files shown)


fax4ever:
I have to fix the WildFly NoSql version

fax4ever:
Hi @Sanne,
it's a very nice idea to parameterize the path of the module. So that we can restore the property. I'll do that.
It would be very nice having multiple WFs integration test, but I guess that in order to have this feature, we should provide different OGM feature pack module distributions. Shouldn't we?
Thanks for the insights!

Sanne:
> It would be very nice having multiple WFs integration test, but I guess that in order to have this feature, we should provide different OGM feature pack module distributions. Shouldn't we?

I think you're right, to be truly able to test with ANY server version we would need more flexibility as possibly our modules structure would need to change internally.

That might be too much work, so not asking for that.  I just hope that merely having these properties in place we will be able to test some servers, specifically the ones which are similar enough to the current one (so with some luck the next WF15 betas could "just work" - no changes needed).

So let's just have the properties so that we can _test_ on different WildFly versions; we're not going to actually publish multiple module versions.

That will help to also know if - for example the modules pack we released for 14.0.1. is compatible with a slightly different server version (or not - it's just important to know about it).

Thanks

fax4ever:
I put back the property. On the other hand, deepening better the issue, it would seem that the problem was the module slot and not the module name.

DavideD:
This solution seems to work @fax4ever, is there anything else to update?

By the way, if these are properties, the creation of new job on Jenkins for the different WildFLy is very easy. We don't need to test each version of WildFly each time we run the build

fax4ever:
@DavideD It's OK for me.

fax4ever:
Thanks @DavideD!

PR #1109, opened 2019-05-27
[PREVIEW] Lazy start mongodb datastore provider

ggaborg:
Relates to https://discourse.hibernate.org/t/lazy-start-mongodb-datastore-provider/2740

### DIFF
--- a/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
+++ b/mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
@@ -21,6 +21,7 @@ import com.mongodb.client.MongoDatabase;
 import org.hibernate.boot.registry.classloading.spi.ClassLoaderService;
 import org.hibernate.ogm.cfg.spi.Hosts;
 import org.hibernate.ogm.datastore.mongodb.MongoDBDialect;
+import org.hibernate.ogm.datastore.mongodb.MongoDBProperties;
 import org.hibernate.ogm.datastore.mongodb.configuration.impl.MongoDBConfiguration;
 import org.hibernate.ogm.datastore.mongodb.logging.impl.Log;
 import org.hibernate.ogm.datastore.mongodb.logging.impl.LoggerFactory;
@@ -51,8 +52,11 @@ public class MongoDBDatastoreProvider extends BaseDatastoreProvider implements S
 	private ServiceRegistryImplementor serviceRegistry;
 
 	private MongoClient mongo;
-	private MongoDatabase mongoDb;
+	private volatile MongoDatabase mongoDb;
 	private MongoDBConfiguration config;
+
+	private boolean lazyGetDatabase;
+	private List<Runnable> startupOperations;
 
 	public MongoDBDatastoreProvider() {
 	}
@@ -71,7 +75,8 @@ public class MongoDBDatastoreProvider extends BaseDatastoreProvider implements S
 		OptionsService optionsService = serviceRegistry.getService( OptionsService.class );
 		ClassLoaderService classLoaderService = serviceRegistry.getService( ClassLoaderService.class );
 		ConfigurationPropertyReader propertyReader = new ConfigurationPropertyReader( configurationValues, classLoaderService );
-
+		lazyGetDatabase = propertyReader.property( MongoDBProperties.LAZY_GET_DATABASE, Boolean.class )
+				.withDefault( Boolean.FALSE ).getValue().booleanValue();
 		try {
 			this.config = new MongoDBConfiguration( propertyReader, optionsService.context().getGlobalOptions() );
 		}
@@ -113,7 +118,9 @@ public class MongoDBDatastoreProvider extends BaseDatastoreProvider implements S
 			if ( mongo == null ) {
 				mongo = createMongoClient( config );
 			}
-			mongoDb = extractDatabase( mongo, config );
+			if ( !lazyGetDatabase ) {
+				getDatabase();
+			}
 		}
 		catch (Exception e) {
 			// Wrap Exception in a ServiceException to make the stack trace more friendly
@@ -147,6 +154,19 @@ public class MongoDBDatastoreProvider extends BaseDatastoreProvider implements S
 	}
 
 	public MongoDatabase getDatabase() {
+		if ( null == mongoDb ) {
+			synchronized ( this ) {
+				if ( null == mongoDb ) {
+					mongoDb = extractDatabase( mongo, config );
+					if ( null != startupOperations ) {
+						for ( Runnable operation : startupOperations ) {
+							operation.run();
+						}
+						startupOperations = null;
+					}
+				}
+			}
+		}
 		return mongoDb;
 	}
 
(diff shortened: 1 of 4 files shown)


mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
```diff
@@ -147,6 +154,19 @@ public void stop() {
 	}
 
 	public MongoDatabase getDatabase() {
+		if ( null == mongoDb ) {
+			synchronized ( this ) {
+				if ( null == mongoDb ) {
+					mongoDb = extractDatabase( mongo, config );
```
DavideD: Is this double-check necessary? Can't we use the SIngletone approach for lazy initialization? I'm referring to this article: http://blog.crazybob.org/2007/01/lazy-loading-singletons.html

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
```diff
@@ -51,9 +52,12 @@
 	private ServiceRegistryImplementor serviceRegistry;
 
 	private MongoClient mongo;
-	private MongoDatabase mongoDb;
+	private volatile MongoDatabase mongoDb;
 	private MongoDBConfiguration config;
 
+	private boolean lazyGetDatabase;
```
DavideD: I don't think this boolean is necessary? It seems to only add complexity. In which case would I not want to use lazy initialization? It there are benefits in using it, let's always use it.

DavideD:
Thanks a lot, I left a couple of comments but I think we can add this after the changes.
Let me know what you think.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
ggaborg: don't know, maybe I want to successfully start my dependent service only if there is db connection

DavideD: Testing in this kind of situation is quite easy, one can just run a query. I don't think setting an option on/off is necessary. Let's keep it simple and remove the property for now. We can add it later if somebody ask or there is a compelling case.

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
ggaborg: We can use the IODH if you're not worried about class loading through reflection as the same thread points out. Just tell me if you want and I'll do the change.

DavideD: I don't think it's going to be a problem in this context. We can revert it back later if I'm wrong

mongodb/src/main/java/org/hibernate/ogm/datastore/mongodb/impl/MongoDBDatastoreProvider.java
ggaborg: Ok, I will do the changes and open a JIRA. For the test I'll have to do some more digging.

DavideD:
Thanks a lot, I'll close this PR for now. Feel free to send a new one when you are ready.

ggaborg:
Hi, just realized that the IODH idiom is not fit for us here because the construction of the singleton instance is not guaranteed to not fail. If there is no database connection for example the constructor will fail and we'll end up into NoClassDefFoundError.

DavideD:
I'm not following you. Why can't you catch the exception and throw a more meaningful error?

ggaborg:
> I'm not following you. Why can't you catch the exception and throw a more meaningful error?

The class loader will try to load the instance holder class but will fail and will throw an ExceptionInInitializerError and it marks that class as "bad". And will not take any attempts to load it in future but will throw a NoClassDefFoundError instead.

DavideD:
OK.

I have to admit that I'm doubtful about this feature usefulness. What does it change if you have the error before or after other services are initialised? You won't still be able to connect to the datastore if the service is not ready when you need it and even if it's loaded lazily you still have to wait for it to be created.

I think what you are looking for is an async/reactive approach so that operations are executed when the service (in this case the db) is ready.

That said maybe I'm missing something, we don't add this kind of changes without tests anyway. So, make sure to cover the cases you mentioned. In particular, it would be nice to have an integration test that showcase the example (simplified) you had in mind when you decided to propose this issue.

ggaborg:
Let me respond with a couple of questions:
- If the db fails after the service started, should the service shutdown because it does not have db connection anymore?
- If the service exposes two http calls for example, one which needs the db and the other not, isn't it ok if my service still starts and can fulfill part of its job? Maybe one needs only the non db dependent http call to use and he shiould be able to use it.
- In the end it all comes down to: what is it better to have, runtime dependent services or startup dependent services?

DavideD:
I'm not sure this should be solved by OGM. If you define a service, shouldn't you also define what happens to the service in case of error? Specifically, you could restart the SessionFactory (in this case) or wait for the datastore to start (you could do it with a timeout). 

If your other services don't need a db, wouldn't be better to make sure  they don't depend on the services requiring the DB? If something is wrong with the db I want to know that immediately, not when the application is already started and running (usually). 

By  the way, this solution is only for MongoDB, but we should be consistent with the other datastores as well.

ggaborg:
Then why if I'm switching to mysql the application is starting?

DavideD:
That's a good question. Would you be able to check what's the difference between the two? I suspect that with MySQL there is no validation at startup and therefore doesn't need to connect to the datastore

DavideD:
I haven't checked though

ggaborg:
It depends on the hibernate.hbm2ddl.auto value. Pf course that any value other than 'none' will make also mysql to fail, but with 'none' the behavior should be consistent across all providers, no?

DavideD:
Yes, sadly Hibernate OGM doesn't check that property. We wanted to do it but we hadn't the time. There is open issue about it. Maybe you could fix this issue so that OGM execute the validations only if the property is set to validate.

ggaborg:
I suppose this is the issue https://hibernate.atlassian.net/browse/OGM-571. It seems a rather complex one.

DavideD:
Yes, that's the one/ I don't expect you to solve the full issue, but maybe you can create one issue where we skip the validation if the value is set to 'none'.

ggaborg:
Hi, after having a more thorough look, it seems that even if we implement the skip of validation/schema creation on 'none', the MongoDBDatastoreProvider will still 'extract' the database on start:
	@Override
	public void start() {
		try {
			if ( mongo == null ) {
				mongo = createMongoClient( config );
			}
			**mongoDb = extractDatabase( mongo, config );**
		}
		catch (Exception e) {
			// Wrap Exception in a ServiceException to make the stack trace more friendly
			// Otherwise a generic unable to request service is thrown
			throw log.unableToStartDatastoreProvider( e );
		}
	}
So the need for skipping the 'extract' at start and lazy 'extract' on getDatabase() would still be needed. Correct me if I'm wrong.

DavideD:
Does the getDabase() method connect to MongoDB?

ggaborg:
No, the getDatabase() method only returns the database which was previously instantiated on start()

ggaborg:
Hi, I have created a new pull request related to https://hibernate.atlassian.net/browse/OGM-571.
This is it https://github.com/hibernate/hibernate-ogm/pull/1111.
It does not solve the initial issue of lazy start of the mongodb datastore provider but maybe we can reach there after.

DavideD:
Let's continue the conversation on the new PR: https://github.com/hibernate/hibernate-ogm/pull/1111

PR #1117, opened 2020-04-21
Add build profile to support JDK >= 11. In JDK 9, some APIs got decla…

gurkerl83:
…red as deprecated, including javax.xml.bind. From version 11 onwards, these APIs got removed from the JDK. Utilization requires an explicit dependency declaration.

https://hibernate.atlassian.net/browse/OGM-1567

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -887,6 +887,32 @@
             </properties>
         </profile>
         <profile>
+            <id>java11</id>
+            <activation>
+                <jdk>[11,)</jdk>
+            </activation>
+            <properties>
+                <!-- Standard APIs >= JDK 11 -->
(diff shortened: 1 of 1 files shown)


Sanne:
hi @gurkerl83 , ok we can add this profile. Normally I expect the end user to want to choose the dependency implementation (especially the version) but since you made it an opt-in profile I see no harm in merging this.

Could you create a JIRA ticket? We prefer having a jira code for each change.

gurkerl83:
I have created the ticket https://hibernate.atlassian.net/browse/OGM-1567

Sanne:
thanks! merged

PR #1134, opened 2024-08-31
Upgrade MongoDB driver dependency to mongodb-driver-legacy 4.11.3

DavideD:
Fix https://hibernate.atlassian.net/jira/software/c/projects/OGM/issues/OGM-1588

Based on https://github.com/hibernate/hibernate-ogm/pull/1127

Supersedes https://github.com/hibernate/hibernate-ogm/pull/1127

### DIFF
--- a/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBSessionCLIQueryTest.java
+++ b/mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBSessionCLIQueryTest.java
@@ -25,6 +25,7 @@ import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
 import com.mongodb.BasicDBList;
+import org.bson.Document;
 import org.fest.assertions.Fail;
 import org.fest.assertions.MapAssert;
 
@@ -719,6 +720,8 @@ public class MongoDBSessionCLIQueryTest extends OgmTestCase {
 			queryWithModifiers.append( "'$query': { } " );
 			queryWithModifiers.append( ", '$max': { 'year' : 1881 } " );
 			queryWithModifiers.append( ", '$hint': { 'year' : 1 } " );
+			// It's important to have at least one test with the `$explain` modifier set to false because
+			// if something goes wrong it will return a different result type
 			queryWithModifiers.append( ", '$explain': false " );
 			queryWithModifiers.append( ", '$snapshot': false " );
 			queryWithModifiers.append( ", '$comment': 'Testing comment' " );
@@ -726,10 +729,9 @@ public class MongoDBSessionCLIQueryTest extends OgmTestCase {
 
 			String nativeQuery = "db." + OscarWildePoem.TABLE_NAME + ".find({" + queryWithModifiers + "})";
 
-			NativeQuery query = session.createNativeQuery( nativeQuery ).addEntity( OscarWildePoem.class );
-			@SuppressWarnings("unchecked")
+			NativeQuery<OscarWildePoem> query = session.createNativeQuery( nativeQuery ).addEntity( OscarWildePoem.class );
 			List<OscarWildePoem> result = query.list();
-			assertThat( result ).onProperty( "id" ).containsExactly( athanasia.getId() );
+			assertThat( result ).containsExactly( athanasia );
 		} );
 	}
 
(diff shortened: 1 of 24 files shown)


mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBSessionCLIQueryTest.java
```diff
@@ -680,6 +681,9 @@ public void testAggregateWithMatchSortAndRegex() {
 	}
 
 	@Test
+	// As of MongoDB 4.2, $max requires an index hint.  Given how little $min is used in practice, doesn't seem worth
+	// it to fix the test
+	@Ignore
```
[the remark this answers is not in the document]
jyemin: I did mean `$max`.

The query with`$max` could always just be written with `$lt` in the filter (and `$gt` for `$min`).  See [this documentation](https://www.mongodb.com/docs/manual/reference/method/cursor.max/#behavior) for more info.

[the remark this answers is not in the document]
DavideD: It's possible to specify hints in the native query, I will just update the tests

DavideD:
I think we are ready to merge this PR:

* It tests with MongoDB 7.0 (remote) and MongoDB 3.6 (embedded)
* The `$explain` operator seems to work with native queries and I've changed the test to actually check the result
* I've fixed the tests for the indexes. I've disabled one that we could probably remove in the future or adapt for MongoDB 7.0
* The feature pack works now, but WilldFly cannot create a MongoClient using the subsystem
* The tests for calls to stored procedure is are disabled when using the remote server (we still test them with the embedded one)
* Everything is rebased on top of the latest `main`

A few questions from [the commit in the previous PR](https://github.com/hibernate/hibernate-ogm/pull/1127/commits/e729b52682a3e516386425abf3d5b03d9db425d2):

> * Remove support for $explain.  The only way to explain with the 4.x driver is by calling `FindIterable#explain`.  We could do that, and then use `SingleTupleIterator`, but is it worth it?

I'm prone to say yes, but only because we already support this and it wasn't too hard to implement.
It's probably not the most critical feature, but I guess it can be handy for testing, and if Hibernate OGM supports this, it will also take care of compatibility issues when upgrading the client.

Because we have a native query language, people expect it to support the same features as the MongoDB shell.
We don't make any promises and I don't think it's really possible at the moment, but I think it's nice if we add what's doable.

> Is explain only supported by OGM as a native query?

I think so. I don't think there's anything similar in ORM to map it too. Maybe it could be part of the statistics? Not familiar with that area.

@jyemin Could you have a look at this and let me know if these changes make sense, please?

/cc @yrodiere and @Sanne

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBSessionCLIQueryTest.java
```diff
@@ -743,12 +743,17 @@ public void testFindWithExplain() {
...
 
-			NativeQuery query = session.createNativeQuery( nativeQuery );
-			@SuppressWarnings("unchecked")
-			List<Object[]> result = query.list();
-			// I'm not sure if we can test the content because this is the result of the explain command
-			// and I believe it might change among versions
-			assertThat( result.get( 0 ) ).isNotEmpty();
+			Object[] result = (Object[]) session.createNativeQuery( nativeQuery ).uniqueResult();
+			assertThat( result ).isNotEmpty();
+			// In MongoDB 7.0, the first entry of the result is "1" (I don't know why).
```
jyemin, line 748: I know why you're seeing this: some time after 3.6 the explain format totally changed, and the "first: (but see below) field is the explainVersion, which can either be 1 or 2 depending on configuration.  In my testing against a fresh 7.0 standalone, the value is actually 2 so the test fails.  I tried changing the logic to `result[0] instance of Number ? 1 : 0` to handle different explain versions, but then ran into a different problem: the order of the fields returned in `result` array does not appear to be stable. On some runs the value of `result[index]` is `Document`, as expected by the test, but in other runs it's `String`.  I'm not sure if this is due to to the server returning the fields in different order depending on some factor that we can't control, or whether it's Hibernate, but either way it doesn't seem safe to depend on the order.  

I suggest changing the test to something more flexible, perhaps like this:

```
			boolean foundExplainEvidence = false;
			for ( Object value : result ) {
				// Look for a document value containing the "namespace key". 
                                 // It's expected in the queryPlanner field of the explain output
				if ( value instanceof Document ) {
					Document possibleQueryPlannerDocument = (Document) value;
					if ( possibleQueryPlannerDocument.containsKey( "namespace" ) ) {
						foundExplainEvidence = true;
						break;
					}
				}
			}
			assertThat( foundExplainEvidence ).isEqualTo( true );
```

What do you think?

jyemin (changes requested):
Looks good.  Just one suggestion about the changes to the explain test.

mongodb/src/test/java/org/hibernate/ogm/datastore/mongodb/test/query/nativequery/MongoDBSessionCLIQueryTest.java
DavideD: Ok, thanks. Yes, I will update it.

DavideD: Should be fixed now: 21113763c0e007322d4a8d2ef0cfcf92b6028c79

If you could confirm that it works for you, I will merge this

jyemin: Confirmed.

DavideD:
Merged https://github.com/hibernate/hibernate-ogm/pull/1134

Thanks!

